PR #1, opened 2024-05-15
feat(eigenda-plasma): Commitment verification on read

epociask:


verify/verifier.go
```diff
@@ -0,0 +1,67 @@
+package verify
+
+import (
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda/encoding/rs"
```
teddyknox, line 6: TODO @teddyknox move this logic to the API package since it's being imported.

teddyknox (approved):
LGTM

PR #11, opened 2024-05-17
Update commitment encoding

epociask:
**Changes**
* Converted `DaCommitment` to `GenericCommitment` to align with OP implementation
* Reduced server side encoding to only prepend `DA_TYPE, DA_VERSION` onto the commitment when writing to response body in put method since OP stack will handle the commitment type encoding


teddyknox (approved):
LGTM

PR #15, opened 2024-05-22
feat: mem store

epociask:
**Changes:**
- Add an ephemeral memory store DA backend which is usable via new feature flags
- Added mem store tests for key set/get and expiration flows
- Removed plasma reference in favor of proxy

**Fixes:** https://github.com/Layr-Labs/op-plasma-eigenda/issues/14

### DIFF
--- a/store/memory.go
+++ b/store/memory.go
@@ -8,15 +8,14 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
-	"github.com/urfave/cli"
+	"github.com/urfave/cli/v2"
 )
 
 const (
-	MemStoreName       = "memstore"
-	MemStoreFlagName   = "enable"
-	ExpirationFlagName = "expiration"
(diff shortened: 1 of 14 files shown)


store/memory.go
```diff
@@ -0,0 +1,145 @@
...
+	if _, exists := e.store[key]; !exists {
+		return nil, fmt.Errorf("commitment key not found")
+	}
+
+	return e.store[key], nil
+}
+
+// Put inserts a value into the store.
+func (e *MemStore) Put(ctx context.Context, value []byte) ([]byte, error) {
+	e.Lock()
```
teddyknox, line 95: Is a map in golang not threadsafe? Is there a need to lock here?

teddyknox (approved):
LGTM

store/memory.go
teddyknox: After some googling, apparently it is not threadsafe for writes, unless you use sync.Map. But this seems fine for now.

PR #22, opened 2024-05-25
Add data domain option and max blob size config setting

epociask:
Also updated to produce a realistic DA cert with a KZG commitment.


common/common.go
```diff
@@ -13,3 +46,50 @@ func EqualSlices[P comparable](s1, s2 []P) bool {
 
 	return true
 }
+
+func ConvertToBytes(s string) (uint64, error) {
```
epociask, line 50: Let's rename this to be better reflective of the fact that we're converting to digital units

eigenda/commitment.go
```diff
@@ -35,6 +36,11 @@ func (c Commitment) Encode() []byte {
 	return append([]byte{byte(EigenDA), byte(EigenV0)}, c...)
 }
 
+func StringToCommit(key string) (Commitment, error) {
+	comm, _ := hexutil.Decode(key)
```
epociask, line 40: any reason why we're implicitly ignoring the err?

server.go
```diff
@@ -98,69 +134,58 @@ func (d *DAServer) Start() error {
...
+	svr.WriteResponse(w, input)
+}
 
-	route := path.Dir(r.URL.Path)
-	if route != "/put" {
-		w.WriteHeader(http.StatusBadRequest)
-		return
-	}
+func (svr *Server) HandlePut(w http.ResponseWriter, r *http.Request) {
+	svr.log.Info("PUT", "url", r.URL)
```
epociask, line 188: let's remove this log stmt since its being handled in middleware now

server.go
```diff
@@ -169,29 +194,46 @@ func (d *DAServer) HandlePut(w http.ResponseWriter, r *http.Request) {
...
+	if comm, err = svr.store.Put(r.Context(), input); err != nil {
+		svr.WriteInternalError(w, err)
 		return
 	}
 
 	// write out encoded commitment
-	if _, err := w.Write(eigenda.Commitment.Encode(comm)); err != nil {
-		w.WriteHeader(http.StatusInternalServerError)
-		return
+	svr.WriteResponse(w, comm)
```
epociask, line 203: we should still be encoding the commitment

store/eigenda.go
```diff
@@ -4,60 +4,83 @@ import (
...
+
+	err = e.verifier.Verify(cert.BlobHeader.Commitment, encodedBlob)
 	if err != nil {
 		return nil, err
 	}
 
-	return blob, nil
+	switch domain {
+	case common.BinaryDomain:
+		return e.client.Codec.DecodeBlob(encodedBlob)
```
epociask, line 57: can we not return `decodedBlob` instead?

store/eigenda.go
```diff
@@ -4,60 +4,83 @@ import (
...
+	default:
+		return nil, fmt.Errorf("unexpected domain type: %d", domain)
+	}
 }
 
 // Put disperses a blob for some pre-image and returns the associated RLP encoded certificate commit.
 func (e EigenDAStore) Put(ctx context.Context, value []byte) (comm []byte, err error) {
+	if uint64(len(value)) > e.maxBlobSizeBytes {
+		return nil, fmt.Errorf("blob is larger than max blob size: blob length %d, max blob size %d", len(value), e.maxBlobSizeBytes)
+	}
```
epociask, line 69: like that we're introducing this invariant to the proxy

store/memory.go
```diff
@@ -86,44 +100,110 @@ func (e *MemStore) pruneExpired() {
...
+					// ReferenceBlockNumber: ,
+				},
+				// SignatoryRecordHash: ,
+				// Fee: ,
+				// ConfirmationBlockNumber: ,
+				BatchHeaderHash: mockBatchHeaderHash[:],
+			},
+			// BatchId: ,
+			// BlobIndex: ,
+			// InclusionProof: ,
```
epociask, line 187: think we'll need to fill these randomly as well since they're extracted sequencer side for encoding into the certificate thats sent on-chain and verified against the service manager

server.go
teddyknox: good catch

store/memory.go
teddyknox: Is there any point to filling this randomly if they're not going to validate onchain? Seems like empty values about as good in that case.

epociask:
https://github.com/Layr-Labs/eigenda-proxy/issues/30

teddyknox (approved):

PR #23, opened 2024-05-29
Use High-Level EigenDA Client

teddyknox:


eigenda/config.go
```diff
@@ -1,41 +1,37 @@
...
-	UseTlsFlagName                   = "eigenda-use-tls"
+	RPCFlagName                          = "eigenda-rpc"
+	StatusQueryRetryIntervalFlagName     = "eigenda-status-query-retry-interval"
+	StatusQueryTimeoutFlagName           = "eigenda-status-query-timeout"
+	DisableTlsFlagName                   = "eigenda-disable-tls"
+	ResponseTimeoutFlagName              = "eigenda-response-timeout"
+	CustomQuorumIDsFlagName              = "eigenda-custom-quorum-ids"
+	SignerPrivateKeyHexFlagName          = "eigenda-signer-private-key-hex"
+	PutBlobEncodingVersionFlagName       = "eigenda-put-blob-encoding-version"
+	DisablePointVerificationModeFlagName = "eigenda-disable-point-verification-mode"
```
epociask, line 23: Let's update `README.md` to reflect any of these new values which are required for operating the service (e.g, SignerPrivateKeyHexFlagName)

eigenda/config.go
```diff
@@ -88,27 +85,65 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
-			Usage:   "RPC endpoint of the EigenDA disperser.",
-			EnvVars: prefixEnvVars("TARGET_RPC"),
+			Name:     RPCFlagName,
+			Usage:    "RPC endpoint of the EigenDA disperser.",
+			EnvVars:  prefixEnvVars("TARGET_RPC"),
+			Required: true,
+		},
+		&cli.DurationFlag{
+			Name:     StatusQueryTimeoutFlagName,
+			Usage:    "Timeout for aborting an EigenDA blob dispersal if the disperser does not report that the blob has been confirmed dispersed.",
```
epociask, line 95: knit - using `finalized` is better reflective of the polling flow

eigenda/config.go
```diff
@@ -88,27 +85,65 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
+			EnvVars:  prefixEnvVars("TARGET_GRPC_DISABLE_TLS"),
+			Required: false,
 		},
 		&cli.DurationFlag{
-			Name:    StatusQueryRetryIntervalFlagName,
-			Usage:   "Wait time between retries of EigenDA blob status queries (made while waiting for a blob to be confirmed by).",
-			Value:   5 * time.Second,
-			EnvVars: prefixEnvVars("TARGET_STATUS_QUERY_INTERVAL"),
+			Name:     ResponseTimeoutFlagName,
+			Usage:    "The total amount of time that the client will waiting for a response from the EigenDA disperser.",
```
epociask, line 116: knit:
```suggestion
			Usage:    "The total amount of time that the client will wait for a response from the EigenDA disperser.",
```

eigenda/config.go
```diff
@@ -88,27 +85,65 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
+		&cli.UintSliceFlag{
+			Name:     CustomQuorumIDsFlagName,
+			Usage:    "The quorum IDs to write blobs to using this client. Should not include default quorums 0 or 1.",
+			Value:    cli.NewUintSlice(),
+			EnvVars:  prefixEnvVars("CUSTOM_QUORUM_IDS"),
+			Required: false,
+		},
+		&cli.StringFlag{
+			Name:     SignerPrivateKeyHexFlagName,
+			Usage:    "Signer private key in hex encoded format. This key should not be associated with an Ethereum address holding any funds.",
```
epociask, line 130: Let's put a major disclaimer about expected key mgmt practices in readme for this variable

go.mod
```diff
@@ -123,19 +128,24 @@ require (
...
+	google.golang.org/grpc v1.59.0 // indirect
+	google.golang.org/protobuf v1.33.0 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.2.1 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
 
 replace github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101315.1-rc.4
+
```
epociask, line 148: knit - for maintainability we could put a single comment explaining why we do these remappings

test/e2e_test.go
```diff
@@ -100,6 +114,10 @@ func createTestSuite(t *testing.T) (TestSuite, func()) {
 }
 
 func TestE2EPutGetLogicForEigenDAStore(t *testing.T) {
+	if !runTestnetIntegrationTests {
+		t.Skip("Skipping testnet integration test")
+	}
```
epociask, line 119: I like this a lot better

eigenda/config.go
```diff
@@ -88,27 +85,65 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
 		},
 		&cli.DurationFlag{
-			Name:    StatusQueryRetryIntervalFlagName,
-			Usage:   "Wait time between retries of EigenDA blob status queries (made while waiting for a blob to be confirmed by).",
-			Value:   5 * time.Second,
-			EnvVars: prefixEnvVars("TARGET_STATUS_QUERY_INTERVAL"),
+			Name:     ResponseTimeoutFlagName,
+			Usage:    "The total amount of time that the client will waiting for a response from the EigenDA disperser.",
+			Value:    10 * time.Second,
+			EnvVars:  prefixEnvVars("RESPONSE_TIMEOUT"),
```
epociask, line 118: knit - we prefix `TARGET` for all above env vars

eigenda/config.go
```diff
@@ -88,27 +85,65 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
+			Required: false,
+		},
+		&cli.UintSliceFlag{
+			Name:     CustomQuorumIDsFlagName,
+			Usage:    "The quorum IDs to write blobs to using this client. Should not include default quorums 0 or 1.",
```
epociask, line 123: is this verified in the client constructor?

eigenda/config.go
epociask: also might be worth disclosing to our customers how they can obtain authorization with EigenDA for reduced rate limiting requirements

eigenda/config.go
```diff
@@ -88,27 +85,65 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
+		&cli.StringFlag{
+			Name:     SignerPrivateKeyHexFlagName,
+			Usage:    "Signer private key in hex encoded format. This key should not be associated with an Ethereum address holding any funds.",
+			EnvVars:  prefixEnvVars("SIGNER_PRIVATE_KEY_HEX"),
+			Required: true,
+		},
+		&cli.UintFlag{
+			Name:     PutBlobEncodingVersionFlagName,
+			Usage:    "The blob encoding version to use when writing blobs from the high level interface.",
+			EnvVars:  prefixEnvVars("PUT_BLOB_ENCODING_VERSION"),
```
epociask, line 137: Should we validate this?

epociask:
Mostly knits and documentation stuff - otherwise code LGTM

epociask (commented):
Mostly knits and documentation stuff - otherwise code LGTM

eigenda/config.go
teddyknox: Core doesn't like us verifying the values here in the constructor because it involves hardcoding the 0 and 1 default quorum IDs.

eigenda/config.go
teddyknox: I believe the constructor verifies this.

eigenda/config.go
teddyknox: Why are we doing this?

teddyknox: Either should add to the prefix or remove imo

teddyknox: For now I'm removing

store/eigenda.go
```diff
@@ -30,12 +31,19 @@ func (e EigenDAStore) Get(ctx context.Context, key []byte) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to decode DA cert to RLP format: %w", err)
 	}
-	blob, err := e.client.RetrieveBlob(ctx, cert.BatchHeaderHash, cert.BlobIndex)
+	blob, err := e.client.GetBlob(ctx, cert.BatchHeaderHash, cert.BlobIndex)
```
afkbyte, line 34: I feel like there are still cases where we would want to use the disperser client rather than the high level to retrieve a blob for backwards compatibility but not sure if this is something we should address in this PR.

Say someone already running an chain with an integration using the disperser client upgrades to one using, if they wanted to retrieve and validate a historical batch we would error out.

we can also probably handle this in the integration itself -> if a Get from the Proxy Server fails then we retry with the old disperser interface.

afkbyte: I'm leaning towards handling it in the proxy server because the backwards compatibility handling is something we'd need to duplicate across integrations if we don't handle it here

epociask (approved):
LGTM

store/eigenda.go
teddyknox: Yes agreed. Right now I don't think we have any backwards compatibility concerns but in the future we should build backwards compatibility into this.

PR #25, opened 2024-06-05
chore: more efficient kzg commitment generation

epociask:


teddyknox (approved):
lol. Now we understand.

PR #31, opened 2024-06-07
Update EigenDA proxy to match OP v1.7.6 client expectation

teddyknox:
* moves LoadServer() and server.go into a new server package so that LoadServer() can be used in e2e_test.go (avoiding duplication of DI efforts)
* Removes GenericCommitmentByte from the returned commitment, since it is prepended by the OP client on the clients side..?
* Enable WithLogging() middleware
* Expand WithLogging() to support error logging, update handler signatures and implementations to tie into this
* Update path patterns to match the usage by the OP plasma client

### DIFF
--- a/eigenda/commitment.go
+++ b/eigenda/commitment.go
@@ -33,7 +33,7 @@ const (
 type Commitment []byte
 
 func (c Commitment) Encode() []byte {
-	return append([]byte{byte(op_plasma.GenericCommitmentType), byte(EigenDA), byte(EigenV0)}, c...)
+	return append([]byte{byte(EigenDA), byte(EigenV0)}, c...)
 }
 
 func StringToCommit(key string) (Commitment, error) {
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -139,6 +139,10 @@ func TestE2EPutGetLogicForEigenDAStore(t *testing.T) {
 }
 
 func TestE2EPutGetLogicForMemoryStore(t *testing.T) {
+	if runTestnetIntegrationTests {
+		t.Skip("Skipping non-testnet integration test")
+	}
+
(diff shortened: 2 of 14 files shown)


eigenda/commitment.go
```diff
@@ -33,7 +33,7 @@ const (
 type Commitment []byte
 
 func (c Commitment) Encode() []byte {
-	return append([]byte{byte(op_plasma.GenericCommitmentType), byte(EigenDA), byte(EigenV0)}, c...)
+	return append([]byte{byte(EigenDA), byte(EigenV0)}, c...)
```
epociask, line 36: We should be encoding the commitment server side

test/e2e_test.go
```diff
@@ -139,12 +129,39 @@ func TestE2EPutGetLogicForEigenDAStore(t *testing.T) {
...
 	t.Log("Getting input data from proxy server...")
 	preimage, err := daClient.GetInput(ts.ctx, commit)
-	assert.NoError(t, err)
-	assert.Equal(t, testPreimage, preimage)
+	require.NoError(t, err)
+	require.Equal(t, testPreimage, preimage)
+}
+
+func TestE2EPutGetLogicForMemoryStore(t *testing.T) {
+	if runTestnetIntegrationTests {
```
epociask, line 142: Shouldn't this be the opposite?

eigenda/commitment.go
teddyknox: I don't think this is the case. Let's hop on a call.

test/e2e_test.go
teddyknox: We want to run this test with the normal tests, and not run it with the testnet-integration tests.

epociask (approved):
LGTM

Issue #32, opened 2024-06-07
Add lightweight proxy client


epociask:

epociask:
Closed by https://github.com/Layr-Labs/eigenda-proxy/pull/33

Issue #30, opened 2024-06-06
Test domain filtering using an E2E test


epociask:

epociask:
closed by https://github.com/Layr-Labs/eigenda-proxy/pull/33

PR #33, opened 2024-06-08
feat: Proxy client

epociask:
Closes:
- https://github.com/Layr-Labs/eigenda-proxy/issues/32 
- https://github.com/Layr-Labs/eigenda-proxy/issues/30


client/client.go
```diff
@@ -0,0 +1,130 @@
...
+
+// ProxyClient is an interface for communicating with the EigenDA proxy server
+type ProxyClient interface {
+	Health() error
+	GetData(ctx context.Context, cert *common.Certificate, domain common.DomainType) ([]byte, error)
+	SetData(ctx context.Context, b []byte) (*common.Certificate, error)
+}
+
+// client is the implementation of ProxyClient
+type client struct {
```
teddyknox, line 33: * Why not expose the concrete type?
* Naming-wise, I think "eigenda" should either be in the package name or in the struct/interface name.

client/client.go
```diff
@@ -0,0 +1,130 @@
...
+	cfg        *Config
+	httpClient *http.Client
+}
+
+func New(cfg *Config) ProxyClient {
```
teddyknox, line 38: i.e. why not return a `client` type? (may need to make it a public type)

client/client.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
teddyknox, line 59: Would be nice to add a protocol for including error messages from the response body in the error output here.

client/client.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// TODO: Add support for custom http client option
+type Config struct {
+	URL string
+}
+
+// ProxyClient is an interface for communicating with the EigenDA proxy server
+type ProxyClient interface {
+	Health() error
```
teddyknox, line 27: Is this part of the OP plasma interface? I figured this endpoint was for docker healthchecks.

client/client.go
teddyknox: Is this client specific to OP stack or general?

client/client.go
```diff
@@ -0,0 +1,130 @@
...
+
+// GetData fetches blob data associated with a DA certificate
+func (c *client) GetData(ctx context.Context, cert *common.Certificate, domain common.DomainType) ([]byte, error) {
+	b, err := rlp.EncodeToBytes(cert)
+	if err != nil {
+		return nil, err
+	}
+
+	// encode prefix bytes
+	b = eigenda.GenericPrefix(eigenda.Commitment(b).Encode())
```
teddyknox, line 74: nit: not terribly ergonomic to have `Commitment(b)` do nothing to `b` without a corresponding call to `.Encode()`. What if it looked more like

```
type genericPrefix String
type commitment String
b = eigenda.NewGenericPrefix(eigenda.NewCommitment(b))
```

where these are functions and not type casting operators and b

client/client.go
```diff
@@ -0,0 +1,130 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to store data: %v", resp.StatusCode)
+	}
+
+	b, err = io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(b) < decodingOffset {
```
teddyknox, line 120: nit: might rename this to something more obvious like "minAllowedCommitmentLength" or something like that

client/client.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, fmt.Errorf("read certificate is of invalid length: %d", len(b))
+	}
+
+	var blob *common.Certificate
```
teddyknox, line 124: nit: rename `blob` var to `cert`

common/common.go
```diff
@@ -93,3 +110,26 @@ func ParseBytesAmount(s string) (uint64, error) {
 		return 0, fmt.Errorf("unsupported unit: %s", unit)
 	}
 }
+
+func SubsetExists[P comparable](arr, sub []P) bool {
```
teddyknox, line 114: where is this being used?

client/client.go
teddyknox: Update:

I'm now seeing that `GenericPrefix()` is already a function, but `Commitment()` is not. So my suggestion would be to make `Commitment()` into a simple function so the `Encode()` part can be dropped.

operator-setup
teddyknox, line 1: what does this submodule update accomplish? point it to my merged test srs points?

test/e2e_test.go
```diff
@@ -138,7 +142,59 @@ func TestE2EPutGetLogicForEigenDAStore(t *testing.T) {
...
+	t.Log("Waiting for client to establish connection with plasma server...")
+	// wait for server to come online after starting
+	wait.For(ts.ctx, time.Second*1, func() (bool, error) {
+		err := daClient.Health()
+		if err != nil {
+			return false, nil
+		}
+
+		return true, nil
+	})
```
teddyknox, line 166: I believe that createTestSuite() is already waiting for the server to come up.

https://github.com/Layr-Labs/eigenda-proxy/blob/f270ca10fe3c7be238db3e71ef7d367f6454eeb5/server/server.go#L114

test/e2e_test.go
```diff
@@ -165,3 +221,55 @@ func TestE2EPutGetLogicForMemoryStore(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, testPreimage, preimage)
 }
+
+func TestMemStoreWithProxyClient(t *testing.T) {
```
teddyknox, line 225: nit: Theoretically it should be possible to dry up these tests. Not necessary to do manually but maybe chatgpt can do.

teddyknox (approved):
So the idea here is that we have different versions of the proxy conform to the *effective* plasma interface of different versions of optimism, but then we maintain this client to hold a certain idealized interface constant through those changes. And this client can be used in Orbit, ZK Sync, CDK. Do I have that right?

Approving with comments.

operator-setup
epociask: precisely

test/e2e_test.go
epociask: this just does a time wait and verifies that no error is generated within 10 ms. Feel like doing the await gurantees client <-> server communication is ready versus instantiation the server construct

client/client.go
epociask: general

Issue #24, opened 2024-06-04
Generate structured commitments within memstore  


epociask:

teddyknox:
#22 updates the memstore to return KZG commitments in BlobInfo certs.

Issue #21, opened 2024-05-25
Fault/fraud testing


epociask:
Clients can provide optional role IDs (e.g, `sequencer`, `defensive_validator`) which the server can process for role specific interaction behaviors. 

## Interaction Behavior Types
- `honest`: server returns proper batches associated with a commitment 
- `interval_dishonesty`: server returns a malformed batch every `x` calls for some client role
- `dishonest`: server continuously lies and returns batches that are irrespective of the actual commitment 

## Representation
A json rep can be used to define the expected return behavior for the proxy server; e.g:
```
{
    "sequencer": {
        "mode": "interval_dishonesty",
        "interval": 10,
    },
    "defensive_validator": {
        "mode": "honest"
    },
}

```

teddyknox:
This is good, my only comment is that it should be named something testing specific like "fault-test-mode".

PR #37, opened 2024-06-10
Add query param to GET and PUT for commitment interpretation type

teddyknox:
* Creates type system for commitments complete with literals, Marshal() and Unmarshal() functions, and more. Hopefully allows us to encode the various commitment types such that we cannot make mistakes when using this type system (only when designing it).
* Uses this type system to define a "default" commitment type which is useful for allowing the caller to define any prefix bytes that should prepended after puts and L-trimmed before gets.
* Unifies the `eigenda` and `server` packages to deal with a cyclic dependency between the two.


server/memory_store.go
```diff
@@ -1,4 +1,4 @@
-package store
+package server
 
 import (
```
epociask, line 3: What's the rationale for overloading the server package with storage interface implementations. IMHO it's more clean to keep to get this abstractions separated by package

server/server.go
```diff
@@ -24,27 +21,34 @@ var (
 )
 
 const (
-	invalidDomain = "invalid domain type"
-)
+	invalidDomain         = "invalid domain type"
+	invalidCommitmentType = "invalid commitment type"
```
epociask, line 25: 'invalidCommit' would be aligned for naming consistency with line #24

commitments/op_commitment.go
```diff
@@ -0,0 +1,90 @@
...
+
+const (
+	// Keccak256Byte represents a commitment using Keccak256 hashing.
+	Keccak256Byte OPCommitmentType = 0
+	// DAServiceByte represents a commitment using a DA service.
+	DAServiceByte OPCommitmentType = 1
+)
+
+type OPCommitment struct {
+	keccak256Commitment []byte
```
epociask, line 18: What's the rationale for supporting this commitment type?

server/server.go
```diff
@@ -147,9 +151,14 @@ func (svr *Server) HandleGet(w http.ResponseWriter, r *http.Request) error {
 		svr.WriteBadRequest(w, invalidDomain)
 		return err
 	}
+	commitmentType, err := ReadCommitmentType(r)
```
epociask, line 154: Why are we reading a commit type here?

commitments/da_service_op_commitment.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (e DAServiceOPCommitment) IsEigenDA() bool {
+	return e.eigendaCommitment != nil
+}
+
+func (e DAServiceOPCommitment) MustEigenDAValue() EigenDACommitment {
+	if e.eigendaCommitment != nil {
+		return *e.eigendaCommitment
+	}
+	log.Panic("CommitmentEither does not contain a Keccak256Commitment value")
```
epociask, line 33: Why are we panicking? Should prolly be handling errors gracefully for client context

commitments/eigenda_commitment.go
```diff
@@ -0,0 +1,64 @@
...
+	"log"
+
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// Define the parent and child types
+type EigenDAEncodingCommitmentType byte
+
+const (
+	EigenDAEncoding0Byte EigenDAEncodingCommitmentType = 0
```
epociask, line 14: Knit - pretty verbose naming

server/commitment_type.go
```diff
@@ -0,0 +1,107 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/Layr-Labs/eigenda-proxy/commitments"
+)
+
+type CommitmentType uint8
+
+const (
+	OptimismCommitmentType CommitmentType = iota
```
epociask, line 14: Given this a commitment package you could just do Optimism since when imported would be referenced like commitment.Optimism

server/commitment_type.go
```diff
@@ -0,0 +1,107 @@
...
+		return OptimismCommitmentType
+	case "naked":
+		return NakedCommitmentType
+	default:
+		return UnknownCommitmentType
+	}
+}
+
+func StringToCommit(key string, c CommitmentType) ([]byte, error) {
+	if len(key) == 0 {
```
epociask, line 42: Should be < 2 since we're indexing on that offset

server/commitment_type.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func ReadCommitmentType(r *http.Request) (CommitmentType, error) {
+	query := r.URL.Query()
+	key := query.Get(CommitmentTypeKey)
+	if key == "" { // default
+		return OptimismCommitmentType, nil
+	}
+	dt := StrToCommitmentType(key)
```
epociask, line 101: Knit - more sense to be named ct

commitments/op_commitment.go
teddyknox: We may have to support it in the future for chains migrating from AltLayer's keccak256-based EigenDA PoC.

server/server.go
teddyknox: Domain and CommitmentType are analogous here imo

server/server.go
teddyknox: The idea is that in the "naked" mode the plasma server does not interpret commitments passed to `GetBlob()` in any way, so they are just passed right through to the store. This is not the behavior in "optimism" mode.

Thinking more about it, I'm of the opinion that we we shouldn't offer a true naked mode but should instead return EigenDA-version encoded commitments, i.e. `append([]byte{byte(EigenDACommitmentEncodingVersion)}, rlpEncodedDACert...)`.

commitments/da_service_op_commitment.go
teddyknox: This function should only be called when preceded by a call to `.IsEigenDACommitment()`. Prefixing the function name with `Must` is an idiom in golang to indicate that the function will panic if not called in the right context.

IMO returning an error as well is clunky and redundant with the existence of the IsEigenDACommitment() function.

commitments/eigenda_commitment.go
teddyknox: Updated to CertEncodingV0

server/commitment_type.go
teddyknox: Good idea, will do

server/commitment_type.go
teddyknox: True

commitments/da_service_op_commitment.go
epociask: so my issue with this is that we'll panic on the server handler routine which will result in a dropped connection with insufficient information being provided to the proxy client

test/e2e_test.go
```diff
@@ -114,162 +106,109 @@ func createTestSuite(t *testing.T, useMemory bool) (TestSuite, func()) {
 	}, kill
 }
 
-func TestHoleskyWithPlasmaClient(t *testing.T) {
-	if !runTestnetIntegrationTests {
-		t.Skip("Skipping testnet integration test")
```
epociask, line 119: Can we please revert these testing changes? Made the same adjustments and added parallelisation in the OP testing PR

server/memory_store.go
```diff
@@ -10,60 +10,49 @@ import (
...
 )
 
 const (
-	MemStoreFlagName   = "memstore.enabled"
-	ExpirationFlagName = "memstore.expiration"
-
 	DefaultPruneInterval = 500 * time.Millisecond
 )
 
-type MemStoreConfig struct {
```
epociask, line 30: Feel like we're gonna have to bring this back when supporting configurable interaction patterns to avoid parameter overloading the constructor

teddyknox: > configurable interaction patterns

what do you mean?

commitments/da_service_op_commitment.go
teddyknox: The pattern I'm advocating for:
```
var comm DAServiceOPCommitment
comm.Unmarshal(b)
if comm.IsEigenDA() {
    eigendaComm := comm.MustEigenDAValue()
    // do something
}
```

The below pattern is is everywhere in golang code, is not particularly, and is isomorphic to the above:
```
foo := getObjectOrNil()
if foo != nil {
    foo.bar() // possible NPE if we didn't make the surrounding check
}
```

server/memory_store.go
epociask: https://github.com/Layr-Labs/eigenda-proxy/issues/21

.gitignore
```diff
@@ -24,7 +24,7 @@ go.work
 .env
 
 ## kzg cache
-e2e/resources/kzg/SRSTables/
- 
+../test/resources/SRSTables/
```
epociask, line 27: regression

client/client.go
```diff
@@ -25,8 +23,8 @@ type Config struct {
 // ProxyClient is an interface for communicating with the EigenDA proxy server
 type ProxyClient interface {
 	Health() error
-	GetData(ctx context.Context, cert *common.Certificate, domain common.DomainType) ([]byte, error)
-	SetData(ctx context.Context, b []byte) (*common.Certificate, error)
+	GetData(ctx context.Context, cert []byte, domain server.DomainType, commitmentType server.CommitmentType) ([]byte, error)
+	SetData(ctx context.Context, b []byte, commitmentType server.CommitmentType) ([]byte, error)
```
epociask, line 27: knit - commitmentType could be better encapsulated under a reduced rep like `ct`

commitments/op_commitment.go
```diff
@@ -0,0 +1,90 @@
+package commitments
+
+import (
+	"fmt"
+	"log"
+)
+
+type OPCommitmentType byte
```
epociask, line 8: knit - suffixing `commitment` inside of all package files is redundant with package name. Could suffice as `commitments/op.go`, `commitments/eigenda.go`, etc

commitments/op_commitment.go
epociask: gotcha

commitments/op_commitment.go
```diff
@@ -0,0 +1,90 @@
...
+func (e OPCommitment) MustDAServiceValue() DAServiceOPCommitment {
+	if e.genericCommitment != nil {
+		return *e.genericCommitment
+	}
+	log.Panic("OPCommitment does not contain a DAServiceCommitment value")
+	return DAServiceOPCommitment{} // This will never be reached, but is required for compilation.
+}
+
+func (e OPCommitment) Marshal() ([]byte, error) {
+	if e.IsGenericCommitment() {
```
epociask, line 57: knit - control flow could be represented via switch

e2e/server_test.go
```diff
@@ -87,17 +87,17 @@ func TestProxyClient(t *testing.T) {
 	var testPreimage = []byte("inter-subjective and not objective!")
 
 	t.Log("Setting input data on proxy server...")
-	blobInfo, err := daClient.SetData(ts.Ctx, testPreimage)
+	blobInfo, err := daClient.SetData(ts.Ctx, testPreimage, server.DefaultCommitmentMode)
```
epociask, line 90: knit - having commitment mode as part of the server package could result in ambiguity when imported in rollup repos

e2e/setup.go
```diff
@@ -49,31 +47,26 @@ func CreateTestSuite(t *testing.T, useMemory bool) (TestSuite, func()) {
...
+	eigendaCfg := server.Config{
 		ClientConfig: clients.EigenDAClientConfig{
 			RPC:                      holeskyDA,
 			StatusQueryTimeout:       time.Minute * 45,
 			StatusQueryRetryInterval: time.Second * 1,
 			DisableTLS:               false,
 			SignerPrivateKeyHex:      pk,
 		},
-		CacheDir:               "../operator-setup/resources/SRSTables",
+		CacheDir:               "../test/resources/SRSTables/",
```
epociask, line 58: regression

server/domain_type.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"net/http"
+)
+
+var (
+	ErrInvalidDomainType = fmt.Errorf("invalid domain type")
+)
+
+// DomainType is a enumeration type for the different data domains for which a
```
epociask, line 12:
```suggestion
// DomainType is an enumeration type for the different data domains for which a
```

verify/verify_test.go
```diff
@@ -28,9 +27,9 @@ func TestVerification(t *testing.T) {
 	}
 
 	kzgConfig := &kzg.KzgConfig{
-		G1Path:          "../e2e/resources/kzg/g1.point",
-		G2PowerOf2Path:  "../e2e/resources/kzg/g2.point.powerOf2",
-		CacheDir:        "../e2e/resources/kzg/SRSTables",
+		G1Path:          "../operator-setup/resources/g1_abbr.point",
+		G2PowerOf2Path:  "../operator-setup/resources/g2_abbr.point.powerOf2",
+		CacheDir:        "../test/resources/SRSTables/",
```
epociask, line 32: where is this test directory?

server/server.go
```diff
@@ -184,8 +206,15 @@ func (svr *Server) HandlePut(w http.ResponseWriter, r *http.Request) error {
 		return err
 	}
 
+	comm, err = EncodeCommitment(comm, commitmentType)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return err
+	}
+
+	fmt.Printf("write cert: %x\n", comm)
```
epociask, line 215: looks like left over debugging

server/memory_store_test.go
```diff
@@ -49,21 +45,19 @@ func TestGetSet(t *testing.T) {
...
-	t.Parallel()
-
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
 	kzgConfig := &kzg.KzgConfig{
-		G1Path:          "../e2e/resources/kzg/g1.point",
-		G2PowerOf2Path:  "../e2e/resources/kzg/g2.point.powerOf2",
-		CacheDir:        "../e2e/resources/kzg/SRSTables",
+		G1Path:          "../operator-setup/resources/g1_abbr.point",
```
epociask, line 58: if we're gonna change from `e2e` to `operator-setup` then would it make sense to remove the SRS values stored in the `e2e` directory?

epociask (commented):
just a few comments - mostly knits

.gitignore
epociask: curios as to why the test repo lives as an entity a level above the project directory

server/memory_store.go
teddyknox: We can add this back when we have more than one field.

test/e2e_test.go
teddyknox: Done

.gitignore
teddyknox: Yeah not sure how that got in there.

client/client.go
teddyknox: wdym? shorter variable name?

commitments/op_commitment.go
teddyknox: np

commitments/op_commitment.go
teddyknox: I don't think so, you can't call conditionals like this in `case` statements.

server/memory_store_test.go
teddyknox: Openning a PR to move enough points for 2MiB blobs to this repo, and removing the operator-setup submodule.

epociask (approved):
LGTM 💯 🚀 🚢

PR #41, opened 2024-06-18
Update docs on how to run the proxy locally

Inkvi:
- The previous version of the docs used a parameter (`eigenda-use-tls`) that wasn't longer available.
- Specifying a valid `eigenda-max-blob-length` is necessary. Otherwise an error is thrown:
```
INFO [06-18|16:04:11.833] Initializing EigenDA proxy server...     role=eigenda_proxy
INFO [06-18|16:04:11.834] Using eigenda backend                    role=eigenda_proxy
INFO [06-18|16:04:11.836] failed to read G1 points cannot read file unexpected EOF role=eigenda_proxy
CRIT [06-18|16:04:11.836] Application failed                       role=eigenda_proxy message="failed to create store: cannot read file unexpected EOF"

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ An ephemeral memory store backend can be used for faster feedback testing when p
 ## Running Locally
 
 1. Compile binary: `make eigenda-proxy`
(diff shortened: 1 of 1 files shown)


epociask (approved):
LGTM - ty so much for the contribution 🎆

PR #47, opened 2024-06-20
feat: Fault Testing Support

epociask:
# Fault Testing Support 

## Fixes Issue

Fixes https://github.com/Layr-Labs/eigenda-proxy/issues/21

## Changes proposed
* Introduce configurable fault mode which allows for setting malicious return behaviors for memstore
* Updated `readme.md` with fault mode info and updated soft confirmation docs
* Updated put logic for eigenda to verify the blob length post iFFT encoding


server/load_store.go
```diff
@@ -23,32 +24,50 @@ func LoadStoreRouter(cfg CLIConfig, ctx context.Context, log log.Logger) (*store
...
+
+		memstore, err = store.NewMemStore(ctx, verifier, log, maxBlobLength, expiration, fc)
 		if err != nil {
 			return nil, err
 		}
 	} else {
 		log.Info("Using EigenDA backend")
 	}
 
+	if vCfg.Verify {
```
bxue-l2, line 65: nix, maybe move the code block up

server/server.go
```diff
@@ -144,12 +144,18 @@ func (svr *Server) Health(w http.ResponseWriter, r *http.Request) error {
...
 func (svr *Server) HandleGet(w http.ResponseWriter, r *http.Request) error {
-
+	ctx := context.Background()
 	ct, err := ReadCommitmentMode(r)
 	if err != nil {
 		svr.WriteBadRequest(w, invalidCommitmentMode)
 		return err
 	}
+
+	actor := ReadActor(r)
```
bxue-l2, line 154: I assume we only want memstore to behave like a malicious actor. Can we not touch the code path from other storage type.

client/client.go
```diff
@@ -60,6 +61,10 @@ func (c *client) Health() error {
 func (c *client) GetData(ctx context.Context, comm []byte) ([]byte, error) {
 	url := fmt.Sprintf("%s/get/0x%x?commitment_mode=simple", c.cfg.URL, comm)
 
+	if c.cfg.Actor != "" {
```
bxue-l2, line 64: One high level comment, the eigenda-proxy should be configured to work correctly most of time. Is it reasonable to make Actor a lower layer concept subject to a specific boundary.

server/load_store.go
epociask: good callout

server/server.go
epociask: we can but that'd increase the attack surface for critical system flows that likely shouldn't ever have modifiable return behaviors (i.e, comms with mainnet). we only should ever use this feature for testing sequencer behaviors when doing QA / experimentation

client/client.go
epociask: wdym?

server/server.go
bxue-l2: So can we move actor inside memstore go files, so it no longer appears in the top level codeabse

epociask: how would we do that when it needs to be read from the URL params?

bxue-l2: I guess my concern is that we only intend memstore to have the ability to behave maliciously, then we should localize the change to the memstore itsel. Because memstore is not intended for production use in any cases. Then I feel it is relatively safe. 

My understanding of of the code is that, it uses the context to determine proxy behavior. I worry it touches unnecesary part of code, making it bit harder to maintain. I worry we might accidentally inject some malicious behavior in the production path. Just my thoughts, I would invite @jianoaix to take a look.

server/config.go
```diff
@@ -30,12 +29,15 @@ const (
...
 	DisablePointVerificationModeFlagName = "eigenda-disable-point-verification-mode"
 	// Kzg flags
-	G1PathFlagName             = "eigenda-g1-path"
-	G2TauFlagName              = "eigenda-g2-tau-path"
-	CachePathFlagName          = "eigenda-cache-path"
-	MaxBlobLengthFlagName      = "eigenda-max-blob-length"
+	G1PathFlagName        = "eigenda-g1-path"
+	G2TauFlagName         = "eigenda-g2-tau-path"
+	CachePathFlagName     = "eigenda-cache-path"
+	MaxBlobLengthFlagName = "eigenda-max-blob-length"
```
jianoaix, line 35: Can we use "length" and "size" consistently? "length" is number of symbols from the field (32 bytes per symbol), and "size" is number of bytes

store/fault_mode.go
```diff
@@ -0,0 +1,79 @@
...
+	Actors map[string]Behavior
+}
+
+func (cfg *FaultConfig) AllPolicyExists() bool {
+	_, exists := cfg.Actors[AllPolicyKey]
+	return exists
+}
+
+func LoadFaultConfig(path string) (*FaultConfig, error) {
+	println(fmt.Sprintf("Loading config from %s", path))
```
jianoaix, line 52: Remove debug info here and below?

server/server.go
jianoaix: Is this server for production? Is production server taking the "actor" param from http request, based on which to decide the malicious behavior which is implemented in the store?

epociask: it is but these malicious behavior patterns are only expressed within the memstore implementation which will never be used in production environments.

PR #50, opened 2024-06-21
Readme update

teddyknox:
* convert list of cli flags in README to a table with columns ["cli flag name", "env var flag name", "type", "required?", "default value", "description"]
* update cli helpstring config based on the more thorough readme description of each flag
* update .env.example with the latest


epociask (approved):
LGTM

PR #65, opened 2024-07-23
fix: Resolve commitment decoding bug

epociask:
# Pull Request Template

## Fixes Issue

Fixes #

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/Layr-Labs/eigenda-proxy/commitments"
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -260,6 +261,10 @@ func ReadCommitmentMode(r *http.Request) (commitments.CommitmentMode, error) {
 	if key == "" { // default
 		commit := path.Base(r.URL.Path)
 		if len(commit) > 0 && commit != "put" { // provided commitment in request params
+			if !strings.HasPrefix(commit, "0x") {
```
bxue-l2, line 264: An related thing, when I use eigenda-proxy through curl. I have to put 
127.0.0.1:3100/put/
If I miss the last "/", the eigenda-proxy would not be able to respond. Is it expected based on some standard, or something we can fix

bxue-l2 (approved):

server/server.go
epociask: good callout - lemme add that in :)

epociask: its expected based on a standard - given that client interactions are done programmatically its probably ok to keep it this way

PR #66, opened 2024-07-24
chore: add `help` target to Makefile to display the usage of all targets

caseylove:
Similar to [Layr-Labs/eigenda#653](https://github.com/Layr-Labs/eigenda/pull/653)

This PR adds a new target `help` to display commonly used targets in this project Makefile. This is useful for new developers begin to contribute and get a helpful tool for building

image


samlaf (approved):
LGTM. I do something like this in incredible-squaring, eigensdk, etc.
Eventually I think we should all move to justfiles instead though: https://github.com/casey/just
optimism has recently moved to this and it's quite nice.

samlaf:
Closing due to inactivity.

PR #68, opened 2024-07-26
Update flags.go

InoMurko:
S3 is never read

# Pull Request Template

## Fixes Issue

Fixes #

### DIFF
--- a/server/flags.go
+++ b/server/flags.go
@@ -61,9 +61,11 @@ type CLIConfig struct {
 }
 
 func ReadCLIConfig(ctx *cli.Context) CLIConfig {
+	config := ReadConfig(ctx)
 	return CLIConfig{
-		EigenDAConfig: ReadConfig(ctx),
+		EigenDAConfig: config,
 		MetricsCfg:    opmetrics.ReadCLIConfig(ctx),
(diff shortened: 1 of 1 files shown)


InoMurko:
It probably makes sense to allow 

```
// PutWithoutKey ... 
func (r *Router) PutWithoutKey(ctx context.Context, value []byte) (key []byte, err error) {
	if r.mem != nil {
		r.log.Debug("Storing data to memstore")
		return r.mem.Put(ctx, value)
	}

	if r.eigenda != nil {
		r.log.Info("Storing data to eigenda backend")
		return r.eigenda.Put(ctx, value)
	}

	if r.s3 != nil {
		r.log.Info("Storing data to S3 backend")
		commitment := crypto.Keccak256(value)

		err = r.s3.Put(ctx, commitment, value)
		if err != nil {
			return nil, err
		}
	}

	return nil, errors.New("no DA storage backend found")

}
```
this chunk of code to do both. 
Namely, after it tries to do eigenda it regardles puts data on s3 bo (as backup). 
And when reading it tries to read from eigenda, if ti doesn't succeed, it tries reading from s3.

epociask (approved):
LGTM - ty for catching this

PR #70, opened 2024-07-30
Inomurko/iam

InoMurko:
# Pull Request Template

## Fixes Issue
https://github.com/Layr-Labs/eigenda-proxy/issues/69

Fixes #

## Changes proposed

- Adds a PATH within s3 bucket variable
- Adds the ability to authenticate iam or static

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -43,6 +43,7 @@ const (
 	S3EndpointFlagName        = "s3.endpoint"
 	S3AccessKeyIDFlagName     = "s3.access-key-id"     // #nosec G101
 	S3AccessKeySecretFlagName = "s3.access-key-secret" // #nosec G101
+	S3BackupFlagName          = "s3.backup"
 )
 
 const BytesPerSymbol = 31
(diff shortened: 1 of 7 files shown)


server/config.go
```diff
@@ -163,6 +166,15 @@ func ReadConfig(ctx *cli.Context) Config {
...
 }
 
+func toS3CredentialType(s string) store.S3CredentialType {
+	if s == string(store.S3CredentialStatic) {
+		return store.S3CredentialStatic
+	} else if s == string(store.S3CredentialIAM) {
+		return store.S3CredentialIAM
+	}
+	return store.S3CredentialUnknown
+}
```
epociask, line 176: knit - `switch` might be more idiomatic

epociask: also it might make more sense for this function to live in `store/s3.go`

store/s3.go
```diff
@@ -11,41 +11,63 @@ import (
...
 }
 
 type S3Store struct {
 	cfg    S3Config
 	client *minio.Client
-	stats *Stats
+	stats  *Stats
 }
 
 func NewS3(cfg S3Config) (*S3Store, error) {
```
epociask, line 37: knit - this function could be refactored to support less duplication if we were to add a receiver to S3 config for generating the client credentials `func (cfg *Config) Credentials() minio.credentials` and only have a single instillation of the client in the s3 constructor

epociask (approved):
LGTM - the changes are pretty lightweight. Left some comments but all knits 🙏🏻

PR #75, opened 2024-08-01
Optimize verifier

bxue-l2:
This PR removes kzgProver from the proxy, with a verifier. 

The verifier is light weighted and intended for such use case.


verify/verify_test.go
```diff
@@ -57,3 +60,53 @@ func TestCommitmentVerification(t *testing.T) {
...
+func TestCommitmentWithTooLargeBlob(t *testing.T) {
+
+	var dataRand [2000 * 32]byte
+	_, err := rand.Read(dataRand[:])
+	require.NoError(t, err)
+	data := dataRand[:]
+
+	kzgConfig := &kzg.KzgConfig{
+		G1Path:          "../resources/g1.point",
+		G2PowerOf2Path:  "../resources/g2.point.powerOf2",
```
epociask, line 73: wondering if we should remove this value entirely as well as `g2.point`

epociask (approved):
LGTM

verify/verify_test.go
bxue-l2: It is a good call out, I think we should just remove the the kzg verifier. It is intended for usage by  DA nodes

bxue-l2: It is going to be a breaking change. Maybe another PR

PR #91, opened 2024-08-17
chore: go fmt

epociask:
# Pull Request Template

## Fixes Issue

Fixes #


samlaf (approved):
We should add a workflow to force all PRs to run make fmt like we do in other repos. :)

PR #92, opened 2024-08-19
chore: bump eigenda dep to v8 and add linting

epociask:
# Pull Request Template

## Fixes Issue

Fixes #


e2e/server_test.go
```diff
@@ -108,7 +108,7 @@ func TestProxyClientWithLargeBlob(t *testing.T) {
 	}
 	daClient := client.New(cfg)
 	//  2MB blob
-	testPreimage := []byte(e2e.RandString(2000000))
+	testPreimage := []byte(e2e.RandString(4_000_000))
```
bxue-l2, line 111: grpc max is 4194304. 
current eigenda max blob size is 4194309.
this check can't detect grpc error case we wanted

bxue-l2 (approved):

e2e/server_test.go
epociask: this test is just verifying that a blob with a size close to the maximum limit can successfully be dispersed and read :)

PR #93, opened 2024-08-20
fix: rpc server metric label shortened to commitment header

samlaf:
# Pull Request Template

## Changes proposed

changed the way get request metrics are recorded to group by commitmentHeader only.
Otherwise we get a gazillion requests and metrics explode (which would cost a loot of money on datadog)

### Screenshots (Optional)

image


epociask (approved):
LGTM - great catch

PR #94, opened 2024-08-20
refactor(metrics): remove dependency on optimism metrics

samlaf:
# Pull Request Template

## Changes proposed

Main issue I wanted to resolve is changing the histogram buckets to be useful (see screenshot below).
We were using the optimism metrics which define buckets for their rpc use case, which don't make sense for our case.
So I just pulled in their code, and added a status code label to the requests_total metric.

### Screenshots (Optional)

image

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	Namespace = "eigenda_proxy"
+	namespace           = "eigenda_proxy"
+	httpServerSubsystem = "http_server"
 )
 
 // Config ... Metrics server configuration
@@ -28,8 +29,7 @@ type Config struct {
 type Metricer interface {
 	RecordInfo(version string)
 	RecordUp()
-	RecordRPCServerRequest(method string) func()
-	RecordRPCClientResponse(method string, err error)
+	RecordRPCServerRequest(method string) func(status string)
 
 	Document() []metrics.DocumentedMetric
 }
@@ -39,7 +39,8 @@ type Metrics struct {
 	Info *prometheus.GaugeVec
 	Up   prometheus.Gauge
 
-	metrics.RPCMetrics
+	HTTPServerRequestsTotal          *prometheus.CounterVec
+	HTTPServerRequestDurationSeconds *prometheus.HistogramVec
 
 	registry *prometheus.Registry
 	factory  metrics.Factory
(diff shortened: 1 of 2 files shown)


metrics/metrics.go
```diff
@@ -60,20 +60,40 @@ func NewMetrics(procName string) *Metrics {
...
+			Name:      "requests_total",
+			Help:      "Total requests to the HTTP server",
+		}, []string{
+			"method", "status",
+		}),
+		HTTPServerRequestDurationSeconds: factory.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: httpServerSubsystem,
+			Name:      "request_duration_seconds",
+			// TODO: we might want different buckets for different routes?
```
epociask, line 88: lets file an issue for this within the repo

server/server.go
```diff
@@ -70,9 +70,11 @@ func WithMetrics(handleFn func(http.ResponseWriter, *http.Request) error, m metr
 		// we want to group all requests by commitment header, otherwise the prometheus metric labels will explode
 		commitmentHeader := r.URL.Path[:3]
 		recordDur := m.RecordRPCServerRequest(commitmentHeader)
-		defer recordDur()
 
-		return handleFn(w, r)
+		err := handleFn(w, r)
```
epociask, line 74: hmm just realized something - the `commitmentHeader` will change based on the caller client. (e.g, nitro uses a generic commitment representation which only prepends a version byte).

samlaf: ehh you're saying nitro doesn't follow the 3 byte convention outlined in https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#commitment-schemas?

metrics/metrics.go
samlaf: within optimism repo you mean?
like a feature to pass your own buckets? Possibly, although their metrics server is also for rpc based requests, not http. So it just felt very different when I was reading it, as opposed to a typical REST endpoint which we have here.
Think its better to just fork and use our own stuff, we'll prob add a bunch of other things.

epociask: I mean our repo :) 
just so we can track

server/server.go
epociask: 🤦🏻‍♂️

epociask: will put up a PR shortly with readme updates - but no it doesn't

epociask: https://github.com/Layr-Labs/eigenda-proxy/pull/95

samlaf: Do you think its worth trying to track these bytes as labels then? Or should we just separate request metrics by put vs get?
Also thinking every request to a given eigenda-proxy request prob uses the same commitmentHeader?

metrics/metrics.go
samlaf: Ah gotcha! https://github.com/Layr-Labs/eigenda-proxy/issues/96

server/server.go
epociask: we should probably do (commitment mode, cert version)

epociask (approved):
LGTM

server/server.go
samlaf: Added this as comment in [7b5aab1](https://github.com/Layr-Labs/eigenda-proxy/pull/94/commits/7b5aab10baa50adb01c7bc49f190c1adf9a55d7e)
Resolving to merge for now.

PR #95, opened 2024-08-21
chore: Update commitment schemas in README

epociask:
## Changes proposed
* Updated readme documentation 
* Fix env ingestion for S3 fallback mode
* Minor refactors on CLI handling logic

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ In order to disperse to the EigenDA network in production, or at high throughput
 | `--s3.bucket` |  | `$EIGENDA_PROXY_S3_BUCKET` | Bucket name for S3 storage. |
 | `--s3.path` |  | `$EIGENDA_PROXY_S3_PATH` | Bucket path for S3 storage. |
 | `--s3.endpoint` |  | `$EIGENDA_PROXY_S3_ENDPOINT` | Endpoint for S3 storage. |
+| `--s3.backup` | `false` | `$EIGENDA_PROXY_S3_BACKUP` | Backup mode for S3. | whether to use S3 as a backup store to ensure resiliency in case of EigenDA read failure |
+| `--s3.timeout` | `5s` | `$EIGENDA_PROXY_S3_TIMEOUT` | timeout for S3 storage operations (e.g. get, put) |
 | `--help, -h` | `false` |  | Show help. |
 | `--version, -v` | `false` |  | Print the version. |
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -155,15 +155,26 @@ Container can be built via running `make docker-build`.
...
+### Optimism Commitment Mode
+For `alt-da` clients running on Optimism, the following commitment schema is supported:
 
 ```
- 0        1        2         3               N
- |--------|--------|---------|---------------|
-  commit   da layer  version  raw commitment
+ 0        1        2         3                N
+ |--------|--------|---------|----------------|
+  commit   da layer  version   raw commitment
```
samlaf, line 164: so what exactly are the 2 first bytes? And why do they not appear in https://specs.optimism.io/experimental/alt-da.html#input-commitment-submission

image

samlaf, line 164: image

So how come the version byte doesn't appear in the specs? https://specs.optimism.io/experimental/alt-da.html#input-commitment-submission

README.md
```diff
@@ -155,15 +155,26 @@ Container can be built via running `make docker-build`.
...
   type       type     byte
 ```
 
+### Generic Commitment Mode
```
samlaf, line 168: What does generic mean here? Is this just nitro? anything else uses this?

README.md
epociask: this is unique to certificate versioning so we can target different verification paths in the future

README.md
epociask: anything that's not OP

README.md
```diff
@@ -86,6 +88,9 @@ An optional `--eigenda-eth-confirmation-depth` flag can be provided to specify a
 
 An ephemeral memory store backend can be used for faster feedback testing when testing rollup integrations. To target this feature, use the CLI flags `--memstore.enabled`, `--memstore.expiration`.
 
+### S3 Fallback
+An optional S3 fallback `--s3.backup` can be triggered to ensure resiliency when **reading**. When enabled, a blob is persisted to S3 after being successfully dispersed using the keccak256 hash of the existing commitment for the entity key. In the event that blobs cannot be read from EigenDA, they will then be retrieved from S3.
```
samlaf, line 92: Dont understand what this means.
Where does "existing commitment" come from?
What does "entity key" mean? Not mentioned anywhere else in the README.

README.md
```diff
@@ -165,8 +172,12 @@ For `alt-da` clients running on Optimism, the following commitment schema is sup
   type       type     byte
 ```
 
-### Generic Commitment Mode
-For generic clients communicating with proxy, the following commitment schema is supported:
+Both `keccak256` (i.e, S3 storage using hash of pre-image for commitment value) and `generic` (i.e, EigenDA) are supported to ensure cross-compatibility with alt-da storage backends if desired by a rollup operator.
```
samlaf, line 175: "hash of pre-image for commitment value" sounds weird, since hash of pre-image is just the commitment value itself.
Can we say something like "using hash of the value as the S3 key" or something like that?

samlaf: maybe I'm missing something as well.

samlaf (approved):
LGTM. Left a few nit comments.

README.md
epociask: just S3 key and the commitment generated by eigenda

README.md
epociask: they're both pretty idempotent

PR #101, opened 2024-08-22
pin library tag to 0.8.1

bxue-l2:


epociask (commented):
this should already be done in main: https://github.com/Layr-Labs/eigenda-proxy/blob/main/go.mod

PR #104, opened 2024-08-25
feat: Support explict modes for read fallbacks and caching

epociask:
## Fixes Issue

Fixes NA

## Changes proposed
**TLDR** Modularized storage backend providers to allow for easy addition in the future. Additionally added explicit support for optional read fallback and caching that allows a user to construct custom configurations of proxy based on their individual risk tolerance / requirement for decentralization. E.g, I'm a customer and want caching with redis and read fallback with S3 - I can now do that (excluding redis cuz we haven't implemented yet). 

**Verbose Summary**
* Two interfaces to differentiate between storage backends
* Modularized routing logic to easily support the addition of new backend types in the future
* Support caching mode which takes in list of supported backend storage strings (i.e, read before referencing eigenda)
* Support fallback mode which takes in list of supported backend storage strings (i.e, read only when eigenda cannot be read from)
* Extended E2E tests to verify both system flows
* Extended config rep and added tests for verification 

## Note to reviewers
We can probably create a definition object for each storage backend to describe its allowed modes but this likely isn't necessary unless we envision support for 10+ backends with additional secondary modes outside of `read-fallback, cache`. We could also support many instance configurations (e.g, a separate S3 bucket for caching vs. fallback) if this is something that customers want down the line. All sensitive flows should be tested, but please review diligently. 

**NOTE:** This PR also allows us to scale memstore in the future as well. E.g, if we have redis we can configure it as a read-fallback target when doing node development to ensure we can still reference old memory state after rebooting the proxy instance.


store/router.go
```diff
@@ -42,112 +52,192 @@ func (r *Router) Get(ctx context.Context, key []byte, cm commitments.CommitmentM
...
+		if r.cacheEnabled() {
+			r.log.Debug("Retrieving data from cached backends")
+			data, err := r.multiSourceRead(ctx, key, false)
+			if err == nil {
+				return data, nil
+			}
+			r.log.Warn("Failed to read from cache targets", "err", err)
+		}
+
+		// 2 - read blob from EigenDA
```
hopeyen, line 76: The flow and all handling of `Get` makes sense to me and seems logically intact, but I feel like this flow can be organized more clearly 🤔 like, `multiSourceRead` could take in the source targets directly instead of having a boolean flag, and wrap around `eigenda.Get` and `.Verify` as well

README.md
```diff
@@ -88,8 +89,11 @@ An optional `--eigenda-eth-confirmation-depth` flag can be provided to specify a
 
 An ephemeral memory store backend can be used for faster feedback testing when testing rollup integrations. To target this feature, use the CLI flags `--memstore.enabled`, `--memstore.expiration`.
 
-### S3 Fallback
-An optional S3 fallback `--s3.backup` can be triggered to ensure resiliency when **reading**. When enabled, a blob is persisted to S3 after being successfully dispersed using the keccak256 hash of the existing commitment for the entity key. In the event that blobs cannot be read from EigenDA, they will then be retrieved from S3. 
+### Storage Fallback
+An optional storage fallback CLI flag `--routing.fallback-targets` can be leveraged to ensure resiliency when **reading**. When enabled, a blob is persisted to a fallback target after being successfully dispersed using the keccak256 hash of the existing EigenDA commitment for the fallback target key. This ensure second order keys are succinct. In the event that blobs cannot be read from EigenDA, they will then be retrieved in linear order from the provided fallback targets.
```
samlaf, line 93:
```suggestion
An optional storage fallback CLI flag `--routing.fallback-targets` can be leveraged to ensure resiliency when **reading**. When enabled, a blob is persisted to a fallback target after being successfully dispersed. Fallback targets use the keccak256 hash of the existing EigenDA commitment as their key, for succinctness. In the event that blobs cannot be read from EigenDA, they will then be retrieved in linear order from the provided fallback targets. 
```

samlaf: First reading was a bit confusing so rearranged. Not sure what you meant by "second order key". second order key sounds like the second key of a multi-map read to me, like dict[key1][key2]

store/eigenda.go
```diff
@@ -13,10 +13,12 @@ import (
 )
 
 type EigenDAStoreConfig struct {
-	MaxBlobSizeBytes     uint64
+	MaxBlobSizeBytes uint64
+	// the # of Ethereum blocks to wait after the L1BlockReference # before attempting to verify
```
samlaf, line 17: is L1BlockReference an eigenda or rollup reference here?
Would this flag being > 0 lead to potentially blocking behavior of some endpoints?

store/eigenda.go
```diff
@@ -13,10 +13,12 @@ import (
...
+	// & accredit a blob
 	EthConfirmationDepth uint64
 
-	// The total amount of time that the client will spend waiting for EigenDA to confirm a blob
+	// total duration time that client waits for blob to confirm
```
samlaf, line 21: is StatusQueryTimeout related to EthConfirmationDepth above? If so, maybe a more informative variable name would help, to show they are related? I remember being confused by this variable first time I read the proxy code, and still am right now.

store/store.go
```diff
@@ -2,16 +2,78 @@ package store
 
 import "context"
 
+type Backend uint8
```
samlaf, line 5:
```suggestion
type BackendType uint8
```

store/store.go
```diff
@@ -2,16 +2,78 @@ package store
...
+
+const (
+	EigenDA Backend = iota
+	Memory
+	S3
+	Redis
+
+	Unknown
+)
```
samlaf, line 14:
```suggestion
const (
	EigenDABackendType BackendType = iota
	MemoryBackendType
	S3BackendType
	RedisBackendType

	UnknownBackendType
)
```

samlaf: Prefer to be full explicit with naming. When I first saw "EigenDA" being returned from the Backend() function, I assumed it was returning a struct, which didn't make much sense.

store/store.go
```diff
@@ -2,16 +2,78 @@ package store
...
+	switch s {
+	case "EigenDA":
+		return EigenDA
+	case "Memory":
+		return Memory
+	case "S3":
+		return S3
+	case "Redis":
+		return Redis
+	case "Unknown":
```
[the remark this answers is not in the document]
samlaf: agree that we can prob just panic, since this is misconfugiration at startup only right?

samlaf (commented):
LGTM. added some small nits

store/store.go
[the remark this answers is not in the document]
epociask: we can do that, only added this to appease linting

server/load_store.go
```diff
@@ -42,37 +42,84 @@ func LoadStoreRouter(ctx context.Context, cfg CLIConfig, log log.Logger) (*store
...
+	fallbacks := make([]store.PrecomputedKeyStore, len(cfg.EigenDAConfig.FallbackTargets))
 
-	if err != nil {
-		return nil, err
+	for i, f := range cfg.EigenDAConfig.FallbackTargets {
+		b := store.StringToBackend(f)
+
+		switch b {
+		case store.S3:
+			fallbacks[i] = s3
```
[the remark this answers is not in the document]
epociask: eventually yes but that requires modularizing backend configs which falls out of scope for this PR

store/eigenda.go
epociask: they are separate - `StatusQueryTimeout` just refers to how long to wait for a blob to confirm before timing out client side

store/eigenda.go
epociask: It's an eigenda reference for the L1 block at which the blob batch was bridged to

utils/utils.go
```diff
@@ -8,6 +8,26 @@ import (
 
 // Helper utility functions //
 
+func ContainsDuplicates[P comparable](s []P) bool {
+	seen := make(map[P]interface{})
```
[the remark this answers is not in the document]
epociask: true - but this verification is only performed upon application bootstrap. will incorporate the change though

server/config.go
```diff
@@ -164,21 +179,31 @@ func ReadConfig(ctx *cli.Context) Config {
...
-	default:
-		return store.S3CredentialUnknown
+// checkTargets ... verifies that a backend target slice is constructed correctly
+func (cfg *Config) checkTargets(targets []string) error {
+	if len(targets) == 0 {
+		return nil
 	}
+
+	if utils.ContainsDuplicates(targets) {
+		return fmt.Errorf("duplicate fallback targets provided")
```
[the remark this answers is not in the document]
epociask: nice catch

store/router.go
epociask: true but you'd have to overload the caller with mutex mgmt - gonna leave as if ok. don't see how this approach would present cleaner abstraction.

README.md
epociask: nice - appreciate the recommendation

store/store.go
epociask: [265b7b9](https://github.com/Layr-Labs/eigenda-proxy/pull/104/commits/265b7b99a0f235eda7e3e3c4dfb9a2a89a883364)

README.md
epociask: [265b7b9](https://github.com/Layr-Labs/eigenda-proxy/pull/104/commits/265b7b99a0f235eda7e3e3c4dfb9a2a89a883364)

utils/utils.go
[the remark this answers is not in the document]
epociask: [265b7b9](https://github.com/Layr-Labs/eigenda-proxy/pull/104/commits/265b7b99a0f235eda7e3e3c4dfb9a2a89a883364)

server/load_store.go
```diff
@@ -42,37 +42,84 @@ func LoadStoreRouter(ctx context.Context, cfg CLIConfig, log log.Logger) (*store
...
+		case store.EigenDA, store.Memory:
+			return nil, fmt.Errorf("EigenDA cannot be used as a fallback target")
+
+		case store.Redis:
+			return nil, fmt.Errorf("redis is not supported yet")
+
+		case store.Unknown:
+			fallthrough
+
+		default:
```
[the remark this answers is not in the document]
epociask: [265b7b9](https://github.com/Layr-Labs/eigenda-proxy/pull/104/commits/265b7b99a0f235eda7e3e3c4dfb9a2a89a883364)

store/store.go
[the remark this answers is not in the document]
epociask: [265b7b9](https://github.com/Layr-Labs/eigenda-proxy/pull/104/commits/265b7b99a0f235eda7e3e3c4dfb9a2a89a883364)

store/eigenda.go
samlaf: Ah can we change the name then to be more explicit? EigendaBatchL1BLockReference or something?
Otherwise it's confusing as could be interpreted like the L1origin of L2 blocks, among other things

store/eigenda.go
samlaf: so why not call it `ConfirmationTimeout` then?

store/eigenda.go
epociask: the variable name is embedded into the disperser response object 😅

epociask: but can definitely update the comment

store/router.go
hopeyen: sounds good!

PR #105, opened 2024-08-27
change disperseBlob response timeout to 60sec

bxue-l2:
Some users might take longer than 10sec to disperse blobs, when it timed out, the workload on disperser is completed prematurely, see(https://github.com/Layr-Labs/eigenda/blob/master/api/clients/eigenda_client.go#L45) and (https://github.com/Layr-Labs/eigenda/blob/master/disperser/apiserver/server.go#L283) whenever context is timedout

Although processing speed on disperser is less than 500ms, the transmission latency might high


epociask (approved):
LGTM

epociask:
Actually lets also update the default values in the `README` table a well 🙏🏻

PR #107, opened 2024-08-28
feat: very basic grafana dashboard

### LINKED ISSUE #97: Create starter grafana dashboard
This will be a nice started that users can start with, instead of everyone having to create their own from scratch.

samlaf:
## Fixes Issue

Fixes #97 

This is currently a very very basic grafana dashboard (empty because proxy wasn't actually running :\):
image

Still not sure what the best way to import this into another project's infra. Would have to play around with it. Leaving here as a draft for now.

### DIFF
--- a/grafana-dashboard.json
+++ b/grafana-dashboard.json
@@ -1,0 +1,243 @@
+{
+    "annotations": {
+        "list": [
+            {
+                "builtIn": 1,
+                "datasource": {
+                    "type": "grafana",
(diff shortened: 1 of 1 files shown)


samlaf:
Closing as I see this was added in some other PR: https://github.com/Layr-Labs/eigenda-proxy/blob/main/monitor/grafana/dashboards/simple_dashboard.json

PR #108, opened 2024-08-29
chore: Update large blob test to use 16mb

epociask:
## Fixes Issue

Fixes #

### DIFF
--- a/e2e/server_test.go
+++ b/e2e/server_test.go
@@ -110,8 +110,8 @@ func TestProxyServerWithLargeBlob(t *testing.T) {
 		URL: ts.Address(),
 	}
 	daClient := client.New(cfg)
-	//  4MB blob
-	testPreimage := []byte(e2e.RandString(4_000_000))
+	//  16MB blob
+	testPreimage := []byte(e2e.RandString(16_000_000))
 
 	t.Log("Setting input data on proxy server...")
 	blobInfo, err := daClient.SetData(ts.Ctx, testPreimage)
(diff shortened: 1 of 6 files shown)


store/router.go
```diff
@@ -189,7 +189,7 @@ func (r *Router) multiSourceRead(ctx context.Context, commitment []byte, fallbac
 			continue
 		}
 		// verify cert:data using EigenDA verification checks
-		err = r.eigenda.Verify(key, data)
+		err = r.eigenda.Verify(commitment, data)
```
bxue-l2, line 192: Is it true that the only cached source is eigenda?

e2e/server_test.go
```diff
@@ -110,8 +110,8 @@ func TestProxyServerWithLargeBlob(t *testing.T) {
 		URL: ts.Address(),
 	}
 	daClient := client.New(cfg)
-	//  4MB blob
-	testPreimage := []byte(e2e.RandString(4_000_000))
+	//  16MB blob
+	testPreimage := []byte(e2e.RandString(16_000_000))
```
bxue-l2, line 114: eigenda accounting uses 2^24, 16777216

bxue-l2:
If possible, can you also update the readme, and default argument, they are current 4MiB

store/router.go
bxue-l2: I am not super familiar with multisource

bxue-l2 (approved):

e2e/server_test.go
epociask: right - the default codec client will pad to the next power of two when doing dispersal

store/router.go
epociask: nah the cached sourced can be arbitrary,  but we'll need to do the eigenda verifications against it to ensure the data hasn't been tampered with

PR #109, opened 2024-08-30
fix: request record label with commitment mode and version

hopeyen:
## Changes proposed

- Modified metrics labels for `http_server_request_total` and `http_server_request_duration` (HTTP server request counts and durations) to include `commitment_mode` and `commitment_version`
- Added `http_server_requests_bad_header` metrics with `method, error_type` label for requests that can't be parsed with `commitment_mode` and `commitment_version`. 

## Note to reviewers

I can combine the new metrics to `RecordRPCServerRequest` and record commitment mode and version (or errors of reading them) after handling the request if preferred


README.md
```diff
@@ -100,6 +100,8 @@ An optional storage caching CLI flag `--routing.cache-targets` can be leveraged
 
 To the see list of available metrics, run `./bin/eigenda-proxy doc metrics`
 
+To quickly set up monitoring dashboard, add eigenda-proxy metrics endpoint to a reachable prometheus server config as a scrape target, add prometheus datasource to Grafana to, and import the existing [Grafana dashboard JSON file](./grafana_dashboard.json)
+
 ## Deployment Guide
```
epociask, line 105: if we're gonna provide an example dashboard should we also add a docker compose that allows users to optionally stand up observability?

metrics/metrics.go
```diff
@@ -29,7 +29,8 @@ type Config struct {
 type Metricer interface {
 	RecordInfo(version string)
 	RecordUp()
-	RecordRPCServerRequest(method string) func(status string)
+	RecordBadRequestHeader(method string, errorType string)
```
epociask, line 32: technically the commitment is passed via a URL parameter and not a request header - tangentially we're already recording response status codes which should encapsulate this metric. Is there any reason to have this?

server/server.go
```diff
@@ -66,16 +66,21 @@ func NewServer(host string, port int, router *store.Router, log log.Logger,
...
+		ct, err := ReadCommitmentMode(r)
+		if err != nil {
+			m.RecordBadRequestHeader(r.Method, "invalid commitment mode")
+			return err
+		}
+		vb, err := ReadCommitmentVersion(r, ct)
+		if err != nil {
+			m.RecordBadRequestHeader(r.Method, "invalid commitment version")
+			return err
+		}
```
epociask, line 79: nice - looks like this operation is done twice now though. Maybe we could extend the `handleFunc` to take these fields in to avoid redundant operations.

README.md
hopeyen: I added a docker compose with eigenda-proxy, prometheus, and grafana!
was thinking to also add a bash script service that just send some test traffic to the running eigenda-proxy, but had some trouble with building and executing it 🤔 (lost 2 hours on this:( lmk if you think it's worth spending more time on)

metrics/metrics.go
hopeyen: yep, i was thinking about this too. I've combined the new metrics to `RecordRPCServerRequest` and record commitment mode and version (or errors of reading them) after handling the request; I modified `handleFunc` to read the schemes and return them after handling so the work isn't doubled, and the reason to have `handleFunc` to return them instead of reading them in is so that the server can easily write bad requests response to the caller

server/server.go
```diff
@@ -156,47 +149,47 @@ func (svr *Server) Health(w http.ResponseWriter, _ *http.Request) error {
 	return nil
 }
 
-func (svr *Server) HandleGet(w http.ResponseWriter, r *http.Request) error {
-	ct, err := ReadCommitmentMode(r)
+func (svr *Server) HandleGet(w http.ResponseWriter, r *http.Request) (commitments.CommitmentMode, commitments.CommitmentVersion, error) {
```
epociask, line 152: Let's encapsulate the `(mode, version)` into some metadata struct

server/server.go
```diff
@@ -290,6 +298,19 @@ func ReadCommitmentMode(r *http.Request) (commitments.CommitmentMode, error) {
...
+	if len(commitment) < 3 {
+		return 0, fmt.Errorf("commitment is too short")
+	}
+
+	if mode == commitments.OptimismAltDA || mode == commitments.OptimismGeneric {
+		return commitment[2], nil
+	}
+	// the only other mode is simple, which take first byte as version
+	return commitment[0], nil
+}
```
epociask, line 312: By commitment version do we mean certificate version?

README.md
epociask: eh I'd just let it be :)

server/server.go
hopeyen: yep, I can update to certificate version for clarify

epociask (approved):
LGTM! Congrats on your first proxy PR

metrics/metrics.go
```diff
@@ -90,7 +99,7 @@ func NewMetrics(subsystem string) *Metrics {
 			Buckets: prometheus.ExponentialBucketsRange(0.05, 1200, 20),
 			Help:    "Histogram of HTTP server request durations",
 		}, []string{
-			"method", // no status on histograms because those are very expensive
+			"method", "commitment_mode", "commitment_version", // no status on histograms because those are very expensive
```
bxue-l2, line 102: do you mean certVersion?

server/server.go
```diff
@@ -294,6 +306,31 @@ func ReadCommitmentMode(r *http.Request) (commitments.CommitmentMode, error) {
 	return commitments.OptimismAltDA, nil
 }
 
+func ReadCommitmentVersion(r *http.Request, mode commitments.CommitmentMode) (byte, error) {
```
bxue-l2, line 309: I don't think cert version is a part of request. From what I understand, the request body only contains the user's data. After the dispersal is returned from EigenDA (after Put), the call returns a DA cert,

https://github.com/Layr-Labs/eigenda-proxy/blob/main/server/server.go#L214 (commitment is actually the DA cert)

then it prepend meta information in front of the DA cert. https://github.com/Layr-Labs/eigenda-proxy/blob/main/commitments/mode.go#L65

then DA cert version is actually hard coded for now. https://github.com/Layr-Labs/eigenda-proxy/blob/main/commitments/da_service_op.go#L41

hopeyen: The cert version I'm getting is the `version byte` described in the docs here: https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#commitment-schemas 
I understand that the version is currently hard coded, but was thinking this could support future changes when there's more DA types.

metrics/metrics.go
hopeyen: yes! updated to `DA_cert_version`

epociask (approved):
LGTM

PR #112, opened 2024-09-05
fix: Commitment decoding bug && server handler tests

epociask:
## Fixes Issue

Fixes #

## Changes proposed
* Created `IRouter` and provide mock generation for interface
* Add Make target for generating go mocks in repo
* Added handler tests
* Fixed commitment decoding bug which was resulting in handler process segfaulting


store/router.go
```diff
@@ -11,6 +13,14 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
+type IRouter interface {
```
hopeyen, line 16: nice abstraction! Are you thinking of adding a different implementation of Router? like cachingRouter/fallbackRouter that specifically add caching/fallback?

epociask: hmm that could be nice to have but fail to the see current use cases or necessity of decoupling here

epociask: dependent on how need shifts in the future it could be valuable though

teddyknox (approved):

PR #115, opened 2024-09-06
Update Makefile: add disperse-test-blob

samlaf:
This'll make it easier to test that the da-proxy is working manually. eg:
```
make run-memstore-server
make disperse-test-blob
```


epociask (approved):
LGTM

PR #121, opened 2024-09-16
chore(breaking): add EIGENDA_PROXY prefix to memstore flags

samlaf:
Needed for consistency with all other flags. For some reason MEMSTORE flags are the only ones without the EIGENDA_PROXY flag.
image

## Breaking
This is a breaking change for anyone who was running the proxy using the env var flags. They will need to add the prefix so we should communicate this change.

### DIFF
--- a/.env.example.holesky
+++ b/.env.example.holesky
@@ -62,10 +62,10 @@ EIGENDA_PROXY_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
 # EIGENDA_PROXY_LOG_PID=false
 
 # Whether to use mem-store for DA logic
-# MEMSTORE_ENABLED=false
+# EIGENDA_PROXY_MEMSTORE_ENABLED=false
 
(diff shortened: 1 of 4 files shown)


hopeyen (commented):
would you be able to also update the .env.example files? like https://github.com/Layr-Labs/eigenda-proxy/blob/ae7a0da530ec89a8432f3d3d6d65af95da46ee02/.env.example.holesky#L65

bxue-l2:
We can make it non-breaking by having both ENV available.

samlaf:
> bfe39a6

Done: [bfe39a6](https://github.com/Layr-Labs/eigenda-proxy/pull/121/commits/bfe39a6d132a5343913cb97b809d98b8ea127e3f)

epociask (approved):
LGTM - breaking the env namespace should be fine given customers should never use them in a production setting

bxue-l2:
> LGTM - breaking the env namespace should be fine given customers should never use them in a production setting

good point

PR #126, opened 2024-09-17
Read only mode

hopeyen:
- Added flag `EIGENDA_PROXY_READ_ONLY`
- Generate random private key if not provided
- Add readonly config to EigenDA router that simply return bad request when calling `PUT` methods

### DIFF
--- a/.env.example.holesky
+++ b/.env.example.holesky
@@ -87,3 +87,6 @@ EIGENDA_PROXY_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
           
 # endpoint for S3 storage
 # EIGENDA_PROXY_S3_ENDPOINT
+
+# set to true to disable writing to EigenDA
+# EIGENDA_PROXY_READ_ONLY=false
--- a/Makefile
+++ b/Makefile
@@ -19,12 +19,6 @@ HOLESKYTEST = TESTNET=true go test -timeout 50m -v ./e2e  -parallel 4 -deploy-co
 eigenda-proxy:
 	env GO111MODULE=on GOOS=$(TARGETOS) GOARCH=$(TARGETARCH) go build -v $(LDFLAGS) -o ./bin/eigenda-proxy ./cmd/server
 
-.PHONY: eigenda-proxy-codesign
-proxy-build-sign-run:
-	env GO111MODULE=on GOOS=$(TARGETOS) GOARCH=$(TARGETARCH) go build -v $(LDFLAGS) -o ./bin/eigenda-proxy ./cmd/server
-	codesign --sign - --force --preserve-metadata=entitlements,requirements,flags,runtime ./bin/eigenda-proxy
-	ENV_PATH=.env ./bin/eigenda-proxy --addr 127.0.0.1 --port 3100
-
 .PHONY: docker-build
 docker-build:
 	@docker build -t $(IMAGE_NAME) .
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,12 @@ HOLESKYTEST = TESTNET=true go test -timeout 50m -v ./e2e  -parallel 4 -deploy-co
 eigenda-proxy:
 	env GO111MODULE=on GOOS=$(TARGETOS) GOARCH=$(TARGETARCH) go build -v $(LDFLAGS) -o ./bin/eigenda-proxy ./cmd/server
 
+.PHONY: eigenda-proxy-codesign
+proxy-build-sign-run:
+	env GO111MODULE=on GOOS=$(TARGETOS) GOARCH=$(TARGETARCH) go build -v $(LDFLAGS) -o ./bin/eigenda-proxy ./cmd/server
+	codesign --sign - --force --preserve-metadata=entitlements,requirements,flags,runtime ./bin/eigenda-proxy
+	ENV_PATH=.env ./bin/eigenda-proxy --addr 127.0.0.1 --port 3100
+
 .PHONY: docker-build
 docker-build:
 	@docker build -t $(IMAGE_NAME) .
--- a/e2e/setup.go
+++ b/e2e/setup.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 	"os"
 	"testing"
@@ -93,10 +94,15 @@ type TestSuite struct {
 func CreateTestSuite(t *testing.T, testCfg *Cfg) (TestSuite, func()) {
 	ctx := context.Background()
 
-	// load signer key from environment
+	// load signer key from environment or generate a random one
 	pk := os.Getenv(privateKey)
 	if pk == "" && !testCfg.UseMemory {
-		t.Fatal("SIGNER_PRIVATE_KEY environment variable not set")
+		t.Logf("SIGNER_PRIVATE_KEY environment variable not set. Generated random private key")
+		randomBytes := make([]byte, 32)
+		if _, err := rand.Read(randomBytes); err != nil {
+			t.Fatalf("Failed to generate random private key: %v", err)
+		}
+		pk = hex.EncodeToString(randomBytes)
 	}
 
 	// load node url from environment
--- a/server/config.go
+++ b/server/config.go
@@ -165,6 +165,17 @@ func (cfg *Config) VerificationCfg() *verify.Config {
 
 // ReadConfig ... parses the Config from the provided flags or environment variables.
 func ReadConfig(ctx *cli.Context) Config {
+	signerPrivateKeyHex := func() string {
+		if key := ctx.String(SignerPrivateKeyHexFlagName); key != "" {
+			return key
+		}
+		randomBytes := make([]byte, 32)
+		if _, err := rand.Read(randomBytes); err != nil {
+			return ""
+		}
+		return hex.EncodeToString(randomBytes)
+	}()
+
 	cfg := Config{
 		RedisCfg: store.RedisConfig{
 			Endpoint: ctx.String(RedisEndpointFlagName),
@@ -183,22 +194,13 @@ func ReadConfig(ctx *cli.Context) Config {
 			Timeout:          ctx.Duration(S3TimeoutFlagName),
 		},
 		ClientConfig: clients.EigenDAClientConfig{
-			RPC:                      ctx.String(EigenDADisperserRPCFlagName),
-			StatusQueryRetryInterval: ctx.Duration(StatusQueryRetryIntervalFlagName),
-			StatusQueryTimeout:       ctx.Duration(StatusQueryTimeoutFlagName),
-			DisableTLS:               ctx.Bool(DisableTLSFlagName),
-			ResponseTimeout:          ctx.Duration(ResponseTimeoutFlagName),
-			CustomQuorumIDs:          ctx.UintSlice(CustomQuorumIDsFlagName),
-			SignerPrivateKeyHex: func() string {
-				if key := ctx.String(SignerPrivateKeyHexFlagName); key != "" {
-					return key
-				}
-				randomBytes := make([]byte, 32)
-				if _, err := rand.Read(randomBytes); err != nil {
-					panic(fmt.Errorf("failed to generate random private key: %w", err))
-				}
-				return hex.EncodeToString(randomBytes)
-			}(),
+			RPC:                          ctx.String(EigenDADisperserRPCFlagName),
+			StatusQueryRetryInterval:     ctx.Duration(StatusQueryRetryIntervalFlagName),
+			StatusQueryTimeout:           ctx.Duration(StatusQueryTimeoutFlagName),
+			DisableTLS:                   ctx.Bool(DisableTLSFlagName),
+			ResponseTimeout:              ctx.Duration(ResponseTimeoutFlagName),
+			CustomQuorumIDs:              ctx.UintSlice(CustomQuorumIDsFlagName),
+			SignerPrivateKeyHex:          signerPrivateKeyHex,
 			PutBlobEncodingVersion:       codecs.BlobEncodingVersion(ctx.Uint(PutBlobEncodingVersionFlagName)),
 			DisablePointVerificationMode: ctx.Bool(DisablePointVerificationModeFlagName),
 		},
@@ -512,7 +514,7 @@ func CLIFlags() []cli.Flag {
 		},
 		&cli.BoolFlag{
 			Name:    ReadOnlyFlagName,
-			Usage:   "Whether the proxy should operate in read-only mode.",
+			Usage:   "Whether to run the proxy in read-only mode; set to true to only allow retrieval from EigenDA.",
 			Value:   false,
 			EnvVars: prefixEnvVars("READ_ONLY"),
 		},
--- a/server/config.go
+++ b/server/config.go
@@ -30,6 +30,7 @@ const (
 	SignerPrivateKeyHexFlagName          = "eigenda-signer-private-key-hex"
 	PutBlobEncodingVersionFlagName       = "eigenda-put-blob-encoding-version"
 	DisablePointVerificationModeFlagName = "eigenda-disable-point-verification-mode"
+	ReadOnlyFlagName                     = "read-only"
 
 	// kzg flags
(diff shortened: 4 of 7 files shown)


Makefile
```diff
@@ -19,6 +19,12 @@ HOLESKYTEST = TESTNET=true go test -timeout 50m -v ./e2e  -parallel 4 -deploy-co
 eigenda-proxy:
 	env GO111MODULE=on GOOS=$(TARGETOS) GOARCH=$(TARGETARCH) go build -v $(LDFLAGS) -o ./bin/eigenda-proxy ./cmd/server
 
+.PHONY: eigenda-proxy-codesign
+proxy-build-sign-run:
+	env GO111MODULE=on GOOS=$(TARGETOS) GOARCH=$(TARGETARCH) go build -v $(LDFLAGS) -o ./bin/eigenda-proxy ./cmd/server
+	codesign --sign - --force --preserve-metadata=entitlements,requirements,flags,runtime ./bin/eigenda-proxy
```
samlaf, line 25: Why do you need to sign if you’re just running this locally?

bxue-l2 (dismissed):
lgtm

Makefile
hopeyen: I needed to sign even if I'm running this locally, for every time I build. If this isn't needed in general, I can remove it and check my local env?

samlaf: I'm building and running this fine (both the binary and the container) without signing. Don't think we should be signing here indeed. Curious what your env config is that requires this though, once you figure it out. I've personally never signed a binary so also curious what exactly this does.

server/config.go
```diff
@@ -179,13 +183,22 @@ func ReadConfig(ctx *cli.Context) Config {
...
+			SignerPrivateKeyHex: func() string {
+				if key := ctx.String(SignerPrivateKeyHexFlagName); key != "" {
+					return key
+				}
+				randomBytes := make([]byte, 32)
+				if _, err := rand.Read(randomBytes); err != nil {
+					panic(fmt.Errorf("failed to generate random private key: %w", err))
+				}
+				return hex.EncodeToString(randomBytes)
+			}(),
```
samlaf, line 201: personally not a big fan of inlined functions like this here. I'd prefer to move this up above the config defn and keep the config params easy to read.

Also if we ever want to change the signature and return an err instead of panicing that will be possible.

server/config.go
```diff
@@ -496,6 +510,12 @@ func CLIFlags() []cli.Flag {
 			Value:   cli.NewStringSlice(),
 			EnvVars: prefixEnvVars("CACHE_TARGETS"),
 		},
+		&cli.BoolFlag{
+			Name:    ReadOnlyFlagName,
+			Usage:   "Whether the proxy should operate in read-only mode.",
+			Value:   false,
+			EnvVars: prefixEnvVars("READ_ONLY"),
+		},
```
samlaf, line 518: Why are we adding this? Who is this feature useful for? Can we add some description here or somewhere else as to the intent of this feature.

store/router.go
```diff
@@ -117,6 +119,10 @@ func (r *Router) Get(ctx context.Context, key []byte, cm commitments.CommitmentM
 
 // Put ... inserts a value into a storage backend based on the commitment mode
 func (r *Router) Put(ctx context.Context, cm commitments.CommitmentMode, key, value []byte) ([]byte, error) {
+	if r.readonly {
+		return nil, errors.New("read-only router does not support Put")
+	}
```
samlaf, line 124: What will this return to the client? Will the Put handler return a 500 with this error?

One other idea is to just disable the put route/handler entirely when in read mode. Have you considered that?

hopeyen: Yes, this will return 500 with this error.

I've considered this approach and tried implementing this approach prior to the current PR. There's restriction from IRouter interface requiring the `[put](https://github.com/Layr-Labs/eigenda-proxy/blob/d124ef8bd80eb97f4feadd6901cc3947d6350874/store/router.go#L18)` handler, but I didn't think it made sense to remove `put` from the interface? 

Also thought about making a separate router type that contains exactly this super simple `put` handler, and create Router type based on the readonly config, but there was a lot of code repetition for all the other functions, and I didn't find a way to easily share functions between two child struct types (the regular `Router` and a `ReadOnly` 🤔  And thought about a wrapper type, but didn't think it was worth it at the end and a bit counter-intuitive . 

Please suggest and I'm happy to learn and redo:)

server/config.go
```diff
@@ -179,13 +183,22 @@ func ReadConfig(ctx *cli.Context) Config {
...
-			ResponseTimeout:              ctx.Duration(ResponseTimeoutFlagName),
-			CustomQuorumIDs:              ctx.UintSlice(CustomQuorumIDsFlagName),
-			SignerPrivateKeyHex:          ctx.String(SignerPrivateKeyHexFlagName),
+			RPC:                      ctx.String(EigenDADisperserRPCFlagName),
+			StatusQueryRetryInterval: ctx.Duration(StatusQueryRetryIntervalFlagName),
+			StatusQueryTimeout:       ctx.Duration(StatusQueryTimeoutFlagName),
+			DisableTLS:               ctx.Bool(DisableTLSFlagName),
+			ResponseTimeout:          ctx.Duration(ResponseTimeoutFlagName),
+			CustomQuorumIDs:          ctx.UintSlice(CustomQuorumIDsFlagName),
+			SignerPrivateKeyHex: func() string {
```
bxue-l2: actually, if someone forgot to set the key, and does not enable read only. Then the server can send unauthenticated data

server/config.go
hopeyen: This is a feature included in the v1.5.0 [document](https://docs.google.com/document/d/1vcL2zpmjNijFGyyMTOz3xgjy9zMya9kGcGtNfDK5VS4/edit#task=SZUBJBN_rUxUnced). I think this was requested by celo. The use case is they only need retrieval when running proxy, and it made sense to develop a read mode that does not require a key.

I've modified the usage string to say `Whether to run the proxy in read-only mode; set to true to only allow retrieval from EigenDA.` Please let me know if this is sufficient

server/config.go
hopeyen: Yes, I realized this as well, I was able to request for dispersal even with random private keys. This was confusing because I thought disperser had authentication checks with ratelimits

server/config.go
hopeyen: Moved this up as suggested!

For returning an error, I decided to return an empty string instead when random generation fails. This leads to the same behavior as the original implementation and doesn't change the current function signature `ReadConfig(ctx *cli.Context) Config {...}`. I'm not aware of a way to return an err without including it in the return type, but happy to learn more if you could point me to an example?

Makefile
hopeyen: okay interesting 🤔 I've removed this, and my env config is really simple, uses `EIGENDA_PROXY_SIGNER_PRIVATE_KEY_HEX, EIGENDA_PROXY_ETH_RPC, EIGENDA_PROXY_EIGENDA_DISPERSER_RPC, EIGENDA_PROXY_SERVICE_MANAGER_ADDR, EIGENDA_PROXY_LOG_LEVEL, EIGENDA_PROXY_LOG_PID, EIGENDA_PROXY_METRICS_ADDR, EIGENDA_PROXY_METRICS_ENABLED, EIGENDA_PROXY_METRICS_PORT`

server/config.go
```diff
@@ -161,6 +165,17 @@ func (cfg *Config) VerificationCfg() *verify.Config {
...
+	signerPrivateKeyHex := func() string {
+		if key := ctx.String(SignerPrivateKeyHexFlagName); key != "" {
+			return key
+		}
+		randomBytes := make([]byte, 32)
+		if _, err := rand.Read(randomBytes); err != nil {
+			return ""
+		}
+		return hex.EncodeToString(randomBytes)
+	}()
```
epociask, line 177: how do we handle the case in-which read mode is disable and the user didn't provide a valid key? should we fail loudly assuming this could be an error case that someone running in production would need to know about?

epociask: knit - would prefer this to live as a separate function within this file to avoid overlaying additional programming abstractions (i.e, callbacks) that aren't generally adopted within the service

store/router.go
epociask, line 124: what if I just wanna run proxy for S3 support using the `keccak256` backend and would like to disable reading from eigenda?

server/config.go
samlaf, line 177:
```suggestion
	signerPrivateKeyHex := ctx.String(SignerPrivateKeyHexFlagName)
	if signerPrivateKeyHex == "" {
		randomBytes := make([]byte, 32)
		if _, err := rand.Read(randomBytes); err != nil {
			return ""
		}
		signerPrivateKeyHex = hex.EncodeToString(randomBytes)
	}
```
Can we do something simpler like this? Really don't like inline functions when not needed.

samlaf: Also not sure about returning "" when rand fails... maybe panic was right. Not super sure what the intended behavior is here so I'll leave that to you.

server/config.go
samlaf: Oh yeah no definitely if you want to return an error you need to change the function signature to add a second error return type. I commented on the changes in a separate comment, so feel free to resolve this convo.

.env.example.holesky
```diff
@@ -87,3 +87,6 @@ EIGENDA_PROXY_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
           
 # endpoint for S3 storage
 # EIGENDA_PROXY_S3_ENDPOINT
+
+# set to true to disable writing to EigenDA
+# EIGENDA_PROXY_READ_ONLY=false
```
epociask, line 92: Thinking about this a bit more pragmatically - we have a LOT of knobs for this service with the env namespace only growing. Would it be better to process this internally using the `SignerPrivateKey` where we treat it as `readonly <-> SignerPrivateKey=""` or unset by the user? 

A lil hesitant to introduce more env flags, thoughts? 
cc - @samlaf, @bxue-l2

e2e/setup.go
```diff
@@ -93,10 +94,15 @@ type TestSuite struct {
...
 	pk := os.Getenv(privateKey)
 	if pk == "" && !testCfg.UseMemory {
-		t.Fatal("SIGNER_PRIVATE_KEY environment variable not set")
+		t.Logf("SIGNER_PRIVATE_KEY environment variable not set. Generated random private key")
+		randomBytes := make([]byte, 32)
+		if _, err := rand.Read(randomBytes); err != nil {
+			t.Fatalf("Failed to generate random private key: %v", err)
+		}
+		pk = hex.EncodeToString(randomBytes)
 	}
```
epociask, line 106: What does generating a unique key do? Looking at the disperser server [implementation](https://github.com/Layr-Labs/eigenda/blob/master/disperser/apiserver/server.go), rate limits are only enforced when dispersing.

epociask: Also authorization headers are only signed by the client when [dispersing](https://github.com/Layr-Labs/eigenda/blob/master/api/clients/disperser_client.go#L109-L202) as well

store/router.go
samlaf: Gotcha! This works I think. We should be returning a 400 instead though. I noticed we return 500s in a bunch of places where we shouldn't (see https://github.com/Layr-Labs/eigenda-proxy/issues/127 for eg)

e2e/setup.go
epociask: All this to say - why can't the key just be a hard-set constant?

server/config.go
```diff
@@ -496,6 +512,12 @@ func CLIFlags() []cli.Flag {
 			Value:   cli.NewStringSlice(),
 			EnvVars: prefixEnvVars("CACHE_TARGETS"),
 		},
+		&cli.BoolFlag{
+			Name:    ReadOnlyFlagName,
+			Usage:   "Whether to run the proxy in read-only mode; set to true to only allow retrieval from EigenDA.",
+			Value:   false,
+			EnvVars: prefixEnvVars("READ_ONLY"),
+		},
```
epociask, line 520: let's also update the readme documentation so users know how to target this functionality

.env.example.holesky
samlaf: Yeah I certainly don't like the fact that we have growing flags without obvious links between themselves. I don't really like the privateKey being empty turning readonly mode though, as that is very surprising. One thing we could do eventually is move to urfave v3 and use mutuallyExclusiveFlags (https://pkg.go.dev/github.com/urfave/cli/v3#MutuallyExclusiveFlags). So basically urfave would force user to either be in readonly mode, or to provide a signerPrivateKey, but can't have both.

samlaf: Regarding the UX problem of celo wanting to run in read-only mode and not having to provide a random signerPrivateKey, seems like the main source of the problem is in the eigenda-client? https://github.com/Layr-Labs/eigenda/blob/97dd2a3da09fd62ff3670afd96c2d402f1ca8453/api/clients/config.go#L55 seems like having a non-empty signerPrivateKey is enforced? Perhaps we would need a read-only eigenda-client?

store/router.go
hopeyen: I'm a bit confused, the read-only mode implemented here will disable all `put` methods no matter what. Are you saying the read-only mode should handle something more granular with the cache/fallback targets?

.env.example.holesky
hopeyen: Yeah, I agree with all the concerns here. I would suggest having a separate refactor effort for the flags.

For comment on eigenda-client, I thought so too, but didn't want to touch `eigenda` code 😬  maybe I should've taken the initiative there which makes more sense. I can go ahead and put a PR there and see what corresponding changes would be good here

hopeyen: reevaluating the approach. Maybe it makes more sense to make a retrieval client instead. Currently `EigenDAClient` only contains a dispersal client that has a separate definition for `RetrieveBlob` than the one for retrieval client that has a more complex/decentralized retrieval logic. 

thinking about read-only's UX, perhaps we want to simply create an `RetrievalClient` instead of an `EigenDAClient` and let `retrieveBlob` serve `Get` and return error for `Put`. But then, why don't the read-only user simply spin up a retrieval client? lmk what you guys think, can also bring this up in the sync

server/config.go
bxue-l2: disperser has a free tier for testnet. but it does not have a free tier for mainnet.

server/config.go
bxue-l2: I think the problem now is that if some op-nodes just want to run the derivation pipeline, there is no way to start the proxy without putting in some fake privateKey.

.env.example.holesky
bxue-l2: We didn't have time to discuss that in the sync. In the current `EigenDAClient`, it looks we didn't put much consideration for clients that require read-only (like op-node or nitro node which only runs derivation). Instead of creating patch here and there, it is good to start from a high level design first, which I think Hope is pointing out.

On spinning up a retrieval client. In OP, the protocol requires there to be a proxy server, so the user can't spin up a client, the proxy has to. 

On refactor, we could create a client struct that wrap around both node and disperser client. What do people think

server/config.go
hopeyen: the server won't be able to send unauthenticated data on the mainnet due to rate-limits, right? what's the desirable behavior in this case? (I'm going to update this random generation to use a constant address like all zeros)

.env.example.holesky
hopeyen: Thanks Bowen, I think taking a step back and agree on high level design makes sense. I think the current PR might be a short term temporary solution if we are okay with this, but I guess no one is rushing to use this, so I'm okay to close this.

For the client struct, are you saying this client struct would live in `eigenda` or in `eigenda-proxy`? what's the reason to wrap around a node for a read only mode 🤔? what about waiting for contract storage changes for operator set and implement distributed retrieval for the retriever first?

bxue-l2: ignore my suggestion, it is probably too low level of a suggestion.

hopeyen: okay, sorry if my questions were overwhelming. I'm more than happy to learn and understand what you think is best

hopeyen:
gonna close for now and create a separate one after updating eigenda-core's `EigenDAClient`

PR #125, opened 2024-09-17
Fix HTTP `PUT` requests panicking with nil-pointer deref when no S3 configured

ezdac:
## Problem
If the CLI is not configured with S3 storage options, HTTP handler for `PUT`ing a keyed commitment will panic with a nil pointer dereference:
```
f1
[...]

PR #129, opened 2024-09-17
Fix s3 panic from keccak commitment put requests

### LINKED ISSUE #128: Bug: HTTP PUT requests panic with nil-pointer deref when no S3 configured
See #125 

Creating this issue to add more context, and also to propose another solution.

Proposed solution in #125 is to use reflection, however https://groups.google.com/g/golang-nuts/c/wnH302gBa4I/m/i_cPI8Gu9P8J states
```
This is a common source of confusion.  The basic answer is to never store something in an interface if you don't expect the methods to be called on it.  The language may allow it, but that violates the semantics of the interface.  To expound, a nil value should usually not be stored in an interface unless it is of a type that has explicitly handled that case in its pointer-valued methods and has no value-receiver methods.
```
which indicates that a better solution would be to set the interface itself to nil, as opposed to setting it to a nil pointer of the S3Store struct value, and then using reflection to check that the value is not nil.

samlaf:
## Fixes Issue

Fixes #128 

## Changes proposed

2 main things in this PR:
1. add a test to reproduce the panic
2. fix the test (https://github.com/Layr-Labs/eigenda-proxy/commit/d8c605b980b43a63559c278836afbe58de66910a)

## Note to reviewers

I'm fine with using either this fix or the one in #125  
But if we want the one in #125 then we should still include the test from this PR.

### DIFF
--- a/e2e/optimism_test.go
+++ b/e2e/optimism_test.go
@@ -123,7 +123,9 @@ func TestOptimismKeccak256Commitment(gt *testing.T) {
 
 	testCfg := e2e.TestConfig(useMemory())
 	testCfg.UseKeccak256ModeS3 = true
-	proxyTS, shutDown := e2e.CreateTestSuite(gt, testCfg)
+
+	tsConfig := e2e.TestSuiteConfig(gt, testCfg)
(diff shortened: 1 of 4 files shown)


bxue-l2 (approved):
lgtm

Issue #130, opened 2024-09-18
Upstream features to eigenda-client


samlaf:
I noticed there's some stuff we're dealing with inside the eigenda-proxy which should be in the eigenda-client. Thinking we should upstream them to eigenda-client to keep the da-proxy's logic only server functionality focused. Features that could be upstreamed:
- [x] wait for confirmation depth (eigenda-client only has wait for finality or don't wait at all, and we implement logic to wait for n number of blocks for confirmation. shouldn't this be part of eigenda-client?)
- [ ] we do batch verification inside eigenda-proxy. shouldn't this be part of eigenda-client like kzg verification?

There's prob more, but this is what I can think of right now.
```[tasklist]
### Tasks
- [ ] https://github.com/Layr-Labs/eigenda-proxy/issues/187
```

litt3:
Remaining work is captured in https://github.com/Layr-Labs/eigenda/issues/1000

PR #131, opened 2024-09-18
fix: negative confDepth bug in verifier

### LINKED ISSUE #123: Potential bug when confirmationDepth of -1
The [default value for confirmationDepth](https://github.com/Layr-Labs/eigenda-proxy/blob/main/server/config.go#L462) is -1. This is used to indicate to wait for finality. However, I believe that this causes a bug in getContextBlock(), where confirmationDepth is used as is, and -1 is subtracted from it before [subtracting that result from the blockNumber](https://github.com/Layr-Labs/eigenda-proxy/blob/09497a94ab764dd5a04d75d3b92ae4cb7917390a/verify/cert.go#L118).

This gives a blockNumber of `curBlockNumber+2`, which when we later query for that block number, results in a `header not found` error, and `internal server error`:
image

samlaf:
## Fixes Issue

Fixes #123 

## Changes proposed

Still think that this confirmation depth feature should be in the eigenda-client itself and not here (see issue #130), but this at least fixes the bug.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -203,7 +203,15 @@ func ReadConfig(ctx *cli.Context) Config {
 		FallbackTargets:        ctx.StringSlice(FallbackTargets),
 		CacheTargets:           ctx.StringSlice(CacheTargets),
 	}
-	cfg.ClientConfig.WaitForFinalization = (cfg.EthConfirmationDepth < 0)
+	// the eigenda client can only wait for 0 confirmations or finality
+	// the da-proxy has a more fine-grained notion of confirmation depth
+	// we use -1 to let the da client wait for finality, and then need to set the confirmation depth
+	// for the da-proxy to 0 (because negative confirmation depth doesn't mean anything and leads to errors)
+	// TODO: should the eigenda-client implement this feature for us instead?
+	if cfg.EthConfirmationDepth < 0 {
+		cfg.ClientConfig.WaitForFinalization = true
+		cfg.EthConfirmationDepth = 0
+	}
 
 	return cfg
 }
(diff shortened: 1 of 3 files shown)


server/config.go
```diff
@@ -203,7 +203,15 @@ func ReadConfig(ctx *cli.Context) Config {
 		FallbackTargets:        ctx.StringSlice(FallbackTargets),
 		CacheTargets:           ctx.StringSlice(CacheTargets),
 	}
-	cfg.ClientConfig.WaitForFinalization = (cfg.EthConfirmationDepth < 0)
+	// the eigenda client can only wait for 0 confirmations or finality
+	// the da-proxy has a more fine-grained notion of confirmation depth
+	// we use -1 to let the da client wait for finality, and then need to set the confirmation depth
+	// for the da-proxy to 0 (because negative confirmation depth doesn't mean anything and leads to errors)
+	// TODO: should the eigenda-client implement this feature for us instead?
```
epociask, line 210: wouldn't that also require the client to do the cert verifications as well?

epociask (approved):
just one general comment - otherwise LGTM! Wonderful catch on this 🚀

server/config.go
samlaf: Let's chat about this! I don't understand why you are saying this (don't have a full model of how all of these things are tied together), but I created https://github.com/Layr-Labs/eigenda-proxy/issues/130 specifically to upstream bunch of these things to the da-client. Is that possible? Is there a reason we haven't done that?

samlaf: Merging. Let's continue discussion in that issue.

PR #132, opened 2024-09-18
dep: update eigenda to latest version for latest eigenda-client logging behavior

samlaf:
Need access to https://github.com/Layr-Labs/eigenda/pull/764

## Fixes Issue

Fixes #

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,10 +3,11 @@ module github.com/Layr-Labs/eigenda-proxy
 go 1.21
 
 require (
-	github.com/Layr-Labs/eigenda v0.8.1
+	github.com/Layr-Labs/eigenda v0.8.4-0.20240918205210-351526af1e80
 	github.com/consensys/gnark-crypto v0.12.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -3,10 +3,11 @@ module github.com/Layr-Labs/eigenda-proxy
 go 1.21
 
 require (
-	github.com/Layr-Labs/eigenda v0.8.1
+	github.com/Layr-Labs/eigenda v0.8.4-0.20240918205210-351526af1e80
```
bxue-l2, line 6: we should update to a proper release tag when doing the proxy release

bxue-l2 (approved):

go.mod
bxue-l2: so this implies, eigenda proxy release has to delay before eigenda-core release

samlaf: gotcha. you'll keep track of this I'm guessing?

PR #133, opened 2024-09-19
Fix metric middleware panic

samlaf:
Recent metric changes were causing panics that weren't caught in tests because the tests never use the metrics and logs middlewares. Added testing for those, and fixed the panic.

## Fixes Issue

Fixes #

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -99,7 +99,7 @@ func NewMetrics(subsystem string) *Metrics {
 			Buckets: prometheus.ExponentialBucketsRange(0.05, 1200, 20),
 			Help:    "Histogram of HTTP server request durations",
 		}, []string{
-			"method", "commitment_mode", "DA_cert_version", // no status on histograms because those are very expensive
+			"method", // no status on histograms because those are very expensive
 		}),
 		registry: registry,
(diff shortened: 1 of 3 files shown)


metrics/metrics.go
```diff
@@ -99,7 +99,7 @@ func NewMetrics(subsystem string) *Metrics {
 			Buckets: prometheus.ExponentialBucketsRange(0.05, 1200, 20),
 			Help:    "Histogram of HTTP server request durations",
 		}, []string{
-			"method", "commitment_mode", "DA_cert_version", // no status on histograms because those are very expensive
+			"method", // no status on histograms because those are very expensive
```
samlaf, line 102: @hopeyen had to remove this because it was causing a revert in the metric middleware, since you weren't passing these labels (because we don't have access to them before the request when this is called):
```
panic: inconsistent label cardinality: expected 3 label values but got 1 in []string{"GET"}
```

There might be another way to keep these in, but I suggest merging this first and then figuring out that in your other PR which should also change the behavior of the handlers to use custom errors instead of relying on the values returned along with an err.

hopeyen (approved):
Hmm I see, I can take a look at adding them back in after this gets merged

PR #134, opened 2024-09-19
fix: custom error wrapped around error and commitment meta

hopeyen:
Added a `HandlerError` type to contain error and commitment meta (mode and version) for metrics

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -1,0 +1,33 @@
+package server
+
+import (
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda-proxy/commitments"
+)
+
+// HandlerError is an interface that can be either a simple error or a MetaError
+type HandlerError interface {
+	error
+	isHandlerError() // marker method
+}
+
+// MetaError includes both an error and commitment metadata
+type MetaError struct {
+	Err  error
+	Meta commitments.CommitmentMeta
+}
+
+func (me MetaError) Error() string {
+	return fmt.Sprintf("Error: %s (Mode: %s, CertVersion: %b)",
+		me.Err.Error(),
+		me.Meta.Mode,
+		me.Meta.CertVersion)
+}
+
+func (MetaError) isHandlerError() {}
+
+// NewMetaError creates a new MetaError
+func NewMetaError(err error, meta commitments.CommitmentMeta) HandlerError {
+	return MetaError{Err: err, Meta: meta}
+}
--- a/server/errors.go
+++ b/server/errors.go
@@ -6,12 +6,6 @@ import (
 	"github.com/Layr-Labs/eigenda-proxy/commitments"
 )
 
-// HandlerError is an interface that can be either a simple error or a MetaError
-type HandlerError interface {
-	error
-	isHandlerError() // marker method
-}
-
 // MetaError includes both an error and commitment metadata
 type MetaError struct {
 	Err  error
@@ -25,9 +19,7 @@ func (me MetaError) Error() string {
 		me.Meta.CertVersion)
 }
 
-func (MetaError) isHandlerError() {}
-
 // NewMetaError creates a new MetaError
-func NewMetaError(err error, meta commitments.CommitmentMeta) HandlerError {
+func NewMetaError(err error, meta commitments.CommitmentMeta) MetaError {
 	return MetaError{Err: err, Meta: meta}
 }
--- a/server/server.go
+++ b/server/server.go
@@ -60,15 +60,23 @@ func NewServer(host string, port int, router store.IRouter, log log.Logger,
 }
 
 // WithMetrics is a middleware that records metrics for the route path.
-func WithMetrics(handleFn func(http.ResponseWriter, *http.Request) (commitments.CommitmentMeta, error),
+func WithMetrics(handleFn func(http.ResponseWriter, *http.Request) (commitments.CommitmentMeta, HandlerError),
 	m metrics.Metricer) func(http.ResponseWriter, *http.Request) error {
 	return func(w http.ResponseWriter, r *http.Request) error {
 		recordDur := m.RecordRPCServerRequest(r.Method)
 
 		meta, err := handleFn(w, r)
+		if err.Err != nil {
+			if err.Meta != nil {
+				recordDur(w.Header().Get("status"), string(err.Meta.Mode), string(err.Meta.CertVersion))
+			} else {
+				recordDur(w.Header().Get("status"), string("NoCommitmentModeSet"), string("NoCertVersionSet"))
+			}
+			return err.Err
+		}
 		// we assume that every route will set the status header
 		recordDur(w.Header().Get("status"), string(meta.Mode), string(meta.CertVersion))
-		return err
+		return nil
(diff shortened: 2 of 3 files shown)


server/server_test.go
```diff
@@ -131,9 +131,9 @@ func TestGetHandler(t *testing.T) {
 
 			meta, err := server.HandleGet(rec, req)
 			if tt.expectError {
-				require.Error(t, err)
+				require.Error(t, err.Err)
```
samlaf, line 134: Do you need to unwrap the err.Err like this? The whole point of custom errors is that `err` itself is an error, you shouldn't need to unwrap it I would think?

server/server.go
```diff
@@ -60,15 +60,23 @@ func NewServer(host string, port int, router store.IRouter, log log.Logger,
...
 
 		meta, err := handleFn(w, r)
+		if err.Err != nil {
+			if err.Meta != nil {
+				recordDur(w.Header().Get("status"), string(err.Meta.Mode), string(err.Meta.CertVersion))
+			} else {
+				recordDur(w.Header().Get("status"), string("NoCommitmentModeSet"), string("NoCertVersionSet"))
+			}
+			return err.Err
+		}
```
samlaf, line 76: when can the meta not be set? Would be good to document this.
Actually thinking more not a big fan of this. We should only return a metaError if the meta fields are set. If they aren't then we can return a simple error. And then we can switch on the error type instead of doing "duck-typing" like this.

server/server.go
```diff
@@ -363,3 +371,20 @@ func (svr *Server) GetStoreStats(bt store.BackendType) (*store.Stats, error) {
...
+	return fmt.Sprintf("HandlerError: %s", e.Err.Error())
+}
+
+// Helper function to create a new HandlerError
+func NewHandlerError(err error, meta *commitments.CommitmentMeta) HandlerError {
+	return HandlerError{
+		Err:  err,
+		Meta: meta,
+	}
+}
```
samlaf, line 390: Let's move this to some other file `errors.go`

server/server.go
```diff
@@ -363,3 +371,20 @@ func (svr *Server) GetStoreStats(bt store.BackendType) (*store.Stats, error) {
...
 	return nil, fmt.Errorf("store not found")
 }
+
+type HandlerError struct {
+	Err  error
+	Meta *commitments.CommitmentMeta
+}
+
+func (e *HandlerError) Error() string {
+	return fmt.Sprintf("HandlerError: %s", e.Err.Error())
```
samlaf, line 381: Should we also print the meta info in here?

server/server.go
```diff
@@ -363,3 +371,20 @@ func (svr *Server) GetStoreStats(bt store.BackendType) (*store.Stats, error) {
 
 	return nil, fmt.Errorf("store not found")
 }
+
+type HandlerError struct {
+	Err  error
+	Meta *commitments.CommitmentMeta
```
samlaf, line 377: let's not use a pointer here. scares me in case this gets changed somewhere else.

samlaf (commented):
I like this! Just leaving a few nits

server/server.go
hopeyen: sounds great, I've removed the pointer

server/server.go
hopeyen: yes, great idea, thanks for the suggestion!

server/server.go
hopeyen: I was thinking this would take care of the case when readCommitmentMode has returned error, but I realized that now that `commitmentMeta {}` won't match to `nil`:( 
I've taken a different approach now by matching like `if meta, ok := err.(MetaError); ok {...}`. let me know if you think it could be improved further

server/server_test.go
hopeyen: Thanks for pointing this out. I needed to unwrap with HandlerError defined with `Meta *commitments.CommitmentMeta`, but updating to `Meta commitments.CommitmentMeta` allowed me to directly access it like an error.

samlaf: oh interesting... and weird! But glad this works now

server/server.go
samlaf: No this is perfect! Just separate file makes it easier but no need to complicate things with separate package until this error is needed in some other package.

server/errors.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda-proxy/commitments"
+)
+
+// HandlerError is an interface that can be either a simple error or a MetaError
+type HandlerError interface {
+	error
+	isHandlerError() // marker method
+}
```
samlaf, line 13: Why are you defining an interface here? Don't think we should be doing this. We want to adhere to the error interface that golang defines, and then we just define custom errors by defining structs that implement the Error() function, and that's it. Then we have access to https://pkg.go.dev/errors#Is and a bunch of goodies.

server/errors.go
```diff
@@ -0,0 +1,33 @@
...
+
+func (me MetaError) Error() string {
+	return fmt.Sprintf("Error: %s (Mode: %s, CertVersion: %b)",
+		me.Err.Error(),
+		me.Meta.Mode,
+		me.Meta.CertVersion)
+}
+
+func (MetaError) isHandlerError() {}
```
samlaf, line 28: This is not used anywhere, also shouldn't be exist. See https://pkg.go.dev/errors#Is

server/server.go
samlaf: Yep this is almost there! Just want to use go's native error library's `is` function (https://pkg.go.dev/errors#Is) instead of type checking like this, which allows us to wrap a MetaError in another error if ever we need that, and will still work.

I recommend https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully if you want to learn more about golang's error handling philosophy. :)

samlaf (commented):
Thanks for the changes! Almost there, just a few more things to follow golang's error standards, and then we're good to merge :)

server/server.go
hopeyen: thank you for the resources! Isn't the current `errors.As` an native error function just like `errors.Is`? My interpretation between these two options were that `errors.Is` checks for exact values of the error while `errors.As` checks for the type of the error, and both options are recursive for wrapped types.

Using `As` made more sense to me, but per your suggestion, if I were to move from `As` to `Is`, it seems I will have to define a function like `func (m MetaError) Is(target error) bool { return target.Err != nil && target.Meta != nil}`. is this what you had in mind?

server/errors.go
hopeyen: sorry it got leftover from experimentations. removed this and the interface as pointed out in the above comment

server/server.go
samlaf: Oh nvm `errors.As` is the right option here. `errors.Is/As` are the same, just that is only checks the type, whereas As unwraps it into a struct that is then usable. Since we want to use that struct, As is perfect here. I was probably still looking at the old code that was just type casting instead of using errors.As.

samlaf (approved):
Thanks for these changes :) Good to merge

PR #138, opened 2024-09-20
Fix config checker logic

samlaf:
Since we merged the PR that fixed the `confDepth=-1` bug, now confDepth is always a non-negative number, so the conf checker is broken. Running `make run-memstore-server` results in this CRIT:
```
CRIT [09-19|21:18:50.564] Application failed                       role=eigenda_proxy message="eth confirmation depth is set for certificate verification, but Eth RPC or SvcManagerAddr is not set"
```

I changed the checks a bit arbitrarily... not super sure if it's correct, @epociask please take a good look. When using the memstore we don't need ethRPC or svsManagerAddr do we?

Tests are failing, waiting for correct semantic before changing them. Also removed the -v from `make test` because that prints too much garbage and is impossible to find the actual error when a test fails.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -26,6 +26,7 @@ const (
 	PutBlobEncodingVersionFlagName       = "eigenda-put-blob-encoding-version"
 	DisablePointVerificationModeFlagName = "eigenda-disable-point-verification-mode"
 
+	// cert verification flags
 	CertVerificationEnabledFlagName = "eigenda-cert-verification-enabled"
 	EthRPCFlagName                  = "eigenda-eth-rpc"
 	SvcManagerAddrFlagName          = "eigenda-svc-manager-addr"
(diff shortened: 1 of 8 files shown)


server/config.go
```diff
@@ -246,16 +246,19 @@ func (cfg *Config) Check() error {
...
-	}
+		if cfg.EthRPC == "" {
+			return fmt.Errorf("eth rpc is not set")
+		}
 
-	if cfg.EthConfirmationDepth >= 0 && (cfg.SvcManagerAddr == "" || cfg.EthRPC == "") {
-		return fmt.Errorf("eth confirmation depth is set for certificate verification, but Eth RPC or SvcManagerAddr is not set")
+		if cfg.SvcManagerAddr == "" {
+			return fmt.Errorf("svc manager address is not set")
+		}
```
epociask, line 261: currently cert verification is an optional feature but this logic makes it mandatory

samlaf: Added a `eigenda-cert-verification-enabled` flag in [6befa7c](https://github.com/Layr-Labs/eigenda-proxy/pull/138/commits/6befa7c9af47c1e8a9795372217668444f898989)
Please make sure the changes make sense.

epociask (commented):
Let's update the README to reflect this new flag type

samlaf:
> ad6094d

Done: [ad6094d](https://github.com/Layr-Labs/eigenda-proxy/pull/138/commits/ad6094d4d2225905f74e8064ccaa06f5eec8879a)

samlaf:
> Let's update the README to reflect this new flag type

Done: [ad6094d](https://github.com/Layr-Labs/eigenda-proxy/pull/138/commits/ad6094d4d2225905f74e8064ccaa06f5eec8879a)

epociask (approved):
LGTM

Issue #127, opened 2024-09-17
Potential bug: submitting oversized blob results in internal server error


samlaf:
Running the [TestProxyServerWithOversizedBlob](https://github.com/Layr-Labs/eigenda-proxy/blob/b1ec82177365669c17b67288b12209cae27d8df3/e2e/server_test.go#L127) e2e test, and it passes but the logs look like:
```
...
/Users/samlaf/devel/eigenlayer/da-integrations/eigenda-proxy/e2e/server_test.go:144: Setting input data on proxy server...
t=2024-09-17T14:23:00-0700 lvl=info msg=request role=eigenda_proxy method=POST url="/put/?commitment_mode=simple"
t=2024-09-17T14:23:00-0700 lvl=debug msg="Storing data to EigenDA backend" role=eigenda_proxy
t=2024-09-17T14:23:00-0700 lvl=error msg="internal server error" role=eigenda_proxy err="put request failed with commitment [] (commitment mode simple): blob is larger than max blob size: blob length 170000000, max blob size 16777216"
t=2024-09-17T14:23:00-0700 lvl=error msg="put request failed with commitment [] (commitment mode simple): blob is larger than max blob size: blob length 170000000, max blob size 16777216" role=eigenda_proxy
--- PASS: TestProxyServerWithOversizedBlob (3.93s)
```

(running on main might see different error msgs since I was running this on the `samlaf--devex-improvements` branch with more expressive error msgs, but the essential idea remains the same)

### Issue
don't think we should be logging an `internal server error` for an error that sounds to me (?) like it should be a 400?

PR #140, opened 2024-09-20
fix: Return 400 when blobs exceed max limit

epociask:
## Fixes Issue

Fixes https://github.com/Layr-Labs/eigenda-proxy/issues/127

### DIFF
--- a/e2e/server_test.go
+++ b/e2e/server_test.go
@@ -1,6 +1,8 @@
 package e2e_test
 
 import (
+	"fmt"
+	"net/http"
 	"strings"
 	"testing"
 	"time"
@@ -185,7 +187,7 @@ func TestProxyServerWithOversizedBlob(t *testing.T) {
 	}
 
 	require.True(t, oversizedError)
-	require.ErrorIs(t, err, store.ErrProxyOversizedBlob, store.ErrEigenDAOversizedBlob)
+	require.Contains(t, err.Error(), fmt.Sprint(http.StatusBadRequest))
 
 }
 
--- a/e2e/server_test.go
+++ b/e2e/server_test.go
@@ -1,8 +1,6 @@
 package e2e_test
 
 import (
-	"fmt"
-	"net/http"
 	"strings"
 	"testing"
 	"time"
@@ -187,7 +185,7 @@ func TestProxyServerWithOversizedBlob(t *testing.T) {
 	}
 
 	require.True(t, oversizedError)
-	require.Contains(t, err.Error(), fmt.Sprint(http.StatusBadRequest))
+	require.ErrorIs(t, err, store.ErrProxyOversizedBlob, store.ErrEigenDAOversizedBlob)
 
 }
 
--- a/e2e/server_test.go
+++ b/e2e/server_test.go
@@ -1,6 +1,8 @@
 package e2e_test
 
 import (
+	"fmt"
+	"net/http"
 	"strings"
 	"testing"
 	"time"
@@ -185,6 +187,7 @@ func TestProxyServerWithOversizedBlob(t *testing.T) {
 	}
 
 	require.True(t, oversizedError)
+	require.Contains(t, err.Error(), fmt.Sprint(http.StatusBadRequest))
 
 }
 
--- a/server/server.go
+++ b/server/server.go
@@ -15,7 +15,6 @@ import (
 	"github.com/Layr-Labs/eigenda-proxy/commitments"
 	"github.com/Layr-Labs/eigenda-proxy/metrics"
 	"github.com/Layr-Labs/eigenda-proxy/store"
-	"github.com/Layr-Labs/eigenda-proxy/utils"
 	"github.com/ethereum-optimism/optimism/op-service/rpc"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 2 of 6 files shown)


e2e/server_test.go
```diff
@@ -185,6 +187,7 @@ func TestProxyServerWithOversizedBlob(t *testing.T) {
 	}
 
 	require.True(t, oversizedError)
+	require.Contains(t, err.Error(), fmt.Sprint(http.StatusBadRequest))
```
samlaf, line 190: Can't you use `require.ErrorIs(t, err, http.StatusBadRequest)` here?
https://pkg.go.dev/github.com/stretchr/testify/require#ErrorIs

server/server.go
```diff
@@ -218,6 +219,12 @@ func (svr *Server) HandlePut(w http.ResponseWriter, r *http.Request) (commitment
 	commitment, err := svr.router.Put(r.Context(), meta.Mode, comm, input)
 	if err != nil {
 		err = fmt.Errorf("put request failed with commitment %v (commitment mode %v): %w", comm, meta.Mode, err)
+
+		if utils.ErrorMessageContainsAny(err, store.OversizedEigenDAError, store.OversizedProxyError) {
+			svr.WriteBadRequest(w, err)
```
samlaf, line 224: nit unrelated to this PR but I don't like that right now we double log every error:
1. once in the svr.WriteBadRequest/InternalError/etc
2. second time in the logger middleware

Would be good to find a nice way to clean this up, and ideally only print in the logging middleware, including the correct status. Moving to a framework like https://echo.labstack.com/docs/middleware/logger is also a possibility.

server/server.go
```diff
@@ -218,6 +219,12 @@ func (svr *Server) HandlePut(w http.ResponseWriter, r *http.Request) (commitment
 	commitment, err := svr.router.Put(r.Context(), meta.Mode, comm, input)
 	if err != nil {
 		err = fmt.Errorf("put request failed with commitment %v (commitment mode %v): %w", comm, meta.Mode, err)
+
+		if utils.ErrorMessageContainsAny(err, store.OversizedEigenDAError, store.OversizedProxyError) {
```
samlaf, line 223:
```suggestion
                // we add here any error that should be returned as a 400 instead of a 500.
                // currently only includes oversized requests
		if utils.ErrorMessageContainsAny(err, store.OversizedEigenDAError, store.OversizedProxyError) {
```

utils/utils.go
```diff
@@ -8,6 +8,21 @@ import (
...
 // Helper utility functions //
 
+func ErrorMessageContainsAny(err error, msgs ...string) bool {
+	if err == nil {
+		return false
+	}
+
+	errMsg := err.Error()
+	for _, str := range msgs {
+		if strings.Contains(errMsg, str) {
```
samlaf, line 18: use https://pkg.go.dev/errors#Is instead of comparing the error strings. This gives better type checking.

store/memory.go
```diff
@@ -134,7 +134,7 @@ func (e *MemStore) Get(_ context.Context, commit []byte) ([]byte, error) {
 func (e *MemStore) Put(_ context.Context, value []byte) ([]byte, error) {
 	time.Sleep(e.config.PutLatency)
 	if uint64(len(value)) > e.config.MaxBlobSizeBytes {
-		return nil, fmt.Errorf("blob is larger than max blob size: blob length %d, max blob size %d", len(value), e.config.MaxBlobSizeBytes)
+		return nil, fmt.Errorf("%s: blob length %d, max blob size %d", OversizedProxyError, len(value), e.config.MaxBlobSizeBytes)
```
samlaf, line 137:
```suggestion
		return nil, fmt.Errorf("%w: blob length %d, max blob size %d", OversizedProxyError, len(value), e.config.MaxBlobSizeBytes)
```

samlaf: wrap errors instead of just concatenating strings. Will require making OverSizedProxyError an actual error, but this way we can use golang's stdlib's error functionality.

samlaf (commented):
Like the approach! But don't like the stringified errors. Just change the strings to actual errors everywhere, and use `errors.Is` to compare errors, and then we're gucci. Eventually I'd like to use custom errors everywhere instead of random fmt.Errorf's but this is a bigger endeavor.

store/memory.go
epociask: [97c51a0](https://github.com/Layr-Labs/eigenda-proxy/pull/140/commits/97c51a0994e1a02077ce6f6a432a1ff6e44d8fd4)

utils/utils.go
epociask: [97c51a0](https://github.com/Layr-Labs/eigenda-proxy/pull/140/commits/97c51a0994e1a02077ce6f6a432a1ff6e44d8fd4)

server/server.go
epociask: [97c51a0](https://github.com/Layr-Labs/eigenda-proxy/pull/140/commits/97c51a0994e1a02077ce6f6a432a1ff6e44d8fd4)

server/server.go
epociask: think we can achieve this quickly by just printing in the logging middleware directly and removing the logs from the writer functions - but a fancy middleware wrapper would be nice hehe

e2e/server_test.go
epociask: don't think we can since it assumes the string contents to be identical

samlaf: It says that it works the same way as `errors.Is`, so this should work now that we're wrapping errors I would think? But either works honestly

samlaf (approved):

PR #142, opened 2024-09-20
refactor: use echo webserver framework

samlaf:


samlaf:
Closing for now. we discussed and we want to minimize external dependencies.

PR #144, opened 2024-09-20
show response commitment

bxue-l2:
Current behavior after put, it shows an empty string
Screenshot 2024-09-20 at 1 19 23 PM

This fix prints the response commitment
Screenshot 2024-09-20 at 1 36 23 PM

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -260,7 +260,7 @@ func (svr *Server) HandlePut(w http.ResponseWriter, r *http.Request) (commitment
 		}
 	}
 
-	svr.log.Info(fmt.Sprintf("write commitment: %x\n", comm))
+	svr.log.Info(fmt.Sprintf("response commitment: %x\n", responseCommit))
 	// write out encoded commitment
 	svr.WriteResponse(w, responseCommit)
 	return meta, nil


server/server.go
```diff
@@ -260,7 +260,7 @@ func (svr *Server) HandlePut(w http.ResponseWriter, r *http.Request) (commitment
 		}
 	}
 
-	svr.log.Info(fmt.Sprintf("write commitment: %x\n", comm))
```
samlaf, line 263: oh yeah I noticed this as well. Not super sure what the intent of this was. If this is only meant for keccak commitments, then we should print it in the if clause above probably? Or was this just a mistake @epociask ?

bxue-l2: I think there is a bit confusion on the term commitment schema, https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#commitment-schemas. The readme section actually means returned commitment-schemas. 

In the commented log above, it only prints comm used when you send data to eigenda-proxy. If it is optimism alt da, the commitment is left empty, so nothing gets printed. 

I often find it useful to debug, if I have the response commitment, which is something rollup will commit onchain. So I am adding this change

samlaf: The responseCommitment should be same as comm when keccak is used right? So there's literally no point in logging `comm`? If that's the case, then let's just merge this. But if there is a difference between `comm` and `responseCommit`, then perhaps we should log both?

bxue-l2: no really, for keccak, you append a version byte 0 before the hash, https://github.com/Layr-Labs/eigenda-proxy/blob/main/commitments/op.go#L110

bxue-l2: since it is a byte, printing will just show 00 in the front

samlaf: oh comm is the commit type?? I thought it was the commitment for keccak commitments, where the commitment is supplied in the POST request?

samlaf (approved):

PR #146, opened 2024-09-22
refactor: clean up flags and configs

samlaf:
We had a single global flags file that defined flags for every subsystem. It was very hard to know which flag is needed for which, and hard to maintain. Broke it down by moving flags/configs to each respective subsystem package. There's still a lot more cleanup to do, but I feel this is good enough to be reviewed and merged, and iterated on. Hard to get everything write at once.

What I like:
- modular: each package defines its own config/flags
- cleanup surfaced some other dependency problems which were hidden. I think this kind of modular approach will force us to think more structurally and find the right abstractions. still a wip though.

What I don't like:
- the way I'm dealing with flags that are defined in 2 packages (for eg maxBlobLength is needed in memstore and verifier config) is kind of ugly (currently memstore reads config from verifier flag)
- e2e test setup is now more ugly and has some repeated hardcoded constants. This is because we removed the indirection from global-flags (which we were using in e2e test setup) to configs. Now each config needs to be properly setup in the default e2e test setup... perhaps we should introduce one more level of indirection (op does this) with flags -> CLIconfig -> config (instead of the current flags -> config). This would allow the e2e tests to be setup with CLIconfigs (these are more easily human understandable, as opposed to the configs that are currently needed)


e2e/setup.go
```diff
@@ -53,15 +59,15 @@ func TestConfig(useMemory bool) *Cfg {
 }
 
 func createRedisConfig(eigendaCfg server.Config) server.CLIConfig {
```
epociask, line 61: Knit - is eigendaCfg overly verbose?

flags/eigendaflags/cli.go
```diff
@@ -0,0 +1,121 @@
...
+	return []cli.Flag{
+		&cli.StringFlag{
+			Name:     DisperserRPCFlagName,
+			Usage:    "RPC endpoint of the EigenDA disperser.",
+			EnvVars:  withEnvPrefix(envPrefix, "DISPERSER_RPC"),
+			Category: category,
+		},
+		&cli.DurationFlag{
+			Name:     StatusQueryTimeoutFlagName,
+			Usage:    "Duration to wait for a blob to finalize after being sent for dispersal. Default is 30 minutes.",
```
epociask, line 45: Knit - hasn't this value been changed given the recent decrease in bridging frequency?

store/precomputed_key/s3/cli.go
```diff
@@ -0,0 +1,92 @@
...
+		&cli.StringFlag{
+			Name:    AccessKeySecretFlagName,
+			Usage:   "access key secret for S3 storage",
+			Value:   "",
+			EnvVars: withEnvPrefix(envPrefix, "ACCESS_KEY_SECRET"),
+		},
+		&cli.StringFlag{
+			Name:    BucketFlagName,
+			Usage:   "bucket name for S3 storage",
+			EnvVars: withEnvPrefix(envPrefix, "BUCKET"),
```
epociask, line 59: Knit - we explicitly set value in some locations but not all

epociask (approved):
LGTM - just a few knits

flags/eigendaflags/cli.go
samlaf: That's a good question. Do right now bridging frequency is:
- preprod: 1 min
- testnet: 10 min (still lol... but guessing they'll change it to 3 min also eventually..?)
- mainnet: 3 min

Finalization itself requires waiting for 24 minutes though. So not even sure how 30 minutes was enough.
Feel like the msg in this flag should also reflect whether the WaitForFinalization flag is true or false?

store/precomputed_key/s3/cli.go
samlaf: [f80fb01](https://github.com/Layr-Labs/eigenda-proxy/pull/146/commits/f80fb01d04f424265e8f04a099a7b2dd52e447b6)

Issue #141, opened 2024-09-20
server doesn't respect da-server REST spec


samlaf:
The [Spec](https://specs.optimism.io/experimental/alt-da.html#da-server) for keyed puts is
```
Request:
  POST /put/<hex_encoded_commitment>
  Content-Type: application/octet-stream
  Body: <preimage_bytes>

Response:
  200 OK
```
but we actually return the commitment back, the same way we do for keyless puts.

Doesn't really matter but feel like we should implement the spec exactly?

PR #147, opened 2024-09-23
chore: Don't return commitment for OP Keccak256 mode on PUT requests

epociask:
## Fixes Issue
https://github.com/Layr-Labs/eigenda-proxy/issues/141 

## Changes proposed
- Fix aforementioned issue 
- Write to redundant or secondary storage providers when reading from EigenDA


server/server.go
```diff
@@ -261,8 +261,10 @@ func (svr *Server) HandlePut(w http.ResponseWriter, r *http.Request) (commitment
 	}
 
 	svr.log.Info(fmt.Sprintf("response commitment: %x\n", responseCommit))
-	// write out encoded commitment
-	svr.WriteResponse(w, responseCommit)
+	// write commitment to resp body if not in OptimismKeccak mode
+	if meta.Mode != commitments.OptimismKeccak {
```
samlaf, line 265: Don’t we still need to write empty header with 200 status? Or is that done somewhere else? Sorry on phone

samlaf (approved):
Approving but do look at my question. If it’s a non issue then feel free to resolve and merge.

server/server.go
epociask: it'll default to 200

PR #148, opened 2024-09-23
Added --s3.enable-tls flag

jcortejoso:
## Changes proposed

Adding a new `--s3.enable-tls` command line flag to optionally enable SSL connection with the S3 endpoint.
Default value for the new flag is `false` to keep current behaviour. 

I have tested with GCP GCS (`--s3.endpoint=storage.googleapis.com`), that requires a secure connection.

### DIFF
--- a/store/precomputed_key/s3/cli.go
+++ b/store/precomputed_key/s3/cli.go
@@ -33,14 +33,14 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 		&cli.StringFlag{
 			Name:     EndpointFlagName,
 			Usage:    "endpoint for S3 storage",
-			EnvVars:  withEnvPrefix(envPrefix, "S3_ENDPOINT"),
+			EnvVars:  withEnvPrefix(envPrefix, "ENDPOINT"),
 			Category: category,
 		},
 		&cli.BoolFlag{
 			Name:     EnableTLSFlagName,
 			Usage:    "enable TLS connection to S3 endpoint",
 			Value:    false,
-			EnvVars:  withEnvPrefix(envPrefix, "S3_ENABLE_TLS"),
+			EnvVars:  withEnvPrefix(envPrefix, "ENABLE_TLS"),
 			Category: category,
 		},
 		&cli.StringFlag{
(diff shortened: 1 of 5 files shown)


server/config.go
```diff
@@ -55,6 +55,7 @@ const (
 	S3BucketFlagName          = "s3.bucket"          // #nosec G101
 	S3PathFlagName            = "s3.path"
 	S3EndpointFlagName        = "s3.endpoint"
```
epociask, line 57: this is awesome - thanks! can you please update the table in the `README.md`?

server/config.go
```diff
@@ -321,6 +323,12 @@ func s3Flags() []cli.Flag {
 			Value:   "",
 			EnvVars: prefixEnvVars("S3_ENDPOINT"),
 		},
+		&cli.BoolFlag{
+			Name:    S3DisableTLSFlagName,
+			Usage:   "whether to disable TLS for S3 storage",
+			Value:   true,
+			EnvVars: prefixEnvVars("S3_DISABLE_TLS"),
+		},
```
samlaf, line 331: I personally don't like having a BoolFlag set to true by default, because then actually setting the flag `--s3.disable-tls` is useless, so the only flag config one would ever use is `--s3.disable-tls=false` which feels counterintuitive.

I'd personally prefer to flip to `--s3.enable-tls`, but @epociask wdyt?

jcortejoso: I tried to keep a similar syntax to `--eigenda-disable-tls` but it's true and logic your point. I'll follow your suggestion, and we can revert or change to something else if @epociask prefers in other way.

store/precomputed_key/s3/cli.go
```diff
@@ -35,9 +36,16 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			EnvVars:  withEnvPrefix(envPrefix, "S3_ENDPOINT"),
 			Category: category,
 		},
+		&cli.BoolFlag{
+			Name:     EnableTLSFlagName,
+			Usage:    "enable TLS connection to S3 endpoint",
+			Value:    false,
+			EnvVars:  withEnvPrefix(envPrefix, "S3_ENABLE_TLS"),
```
samlaf, line 43: WithEnvPrwrix adds the S3 prefix. Should be removed. Seems like we forgot to remove it on some other flags as well. Do you mind changing those as well?

epociask (approved):
LGTM

PR #149, opened 2024-09-24
feat: add request-ids to each request

samlaf:
Needed to be able to trace requests and debug more easily. Still some things to fix like the ability for eigenda-client to use a custom logger based on the request-id in context.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Layr-Labs/eigenda-proxy/commitments"
 	"github.com/Layr-Labs/eigenda-proxy/metrics"
 	"github.com/Layr-Labs/eigenda-proxy/store"
+	"github.com/Layr-Labs/eigenda-proxy/utils"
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -273,17 +277,14 @@ func (svr *Server) WriteResponse(w http.ResponseWriter, data []byte) {
 }
 
 func (svr *Server) WriteInternalError(w http.ResponseWriter, err error) {
-	svr.log.Error("internal server error", "err", err)
```
samlaf, line 276: Removing these because of duplicate logs (also printed in middleware). However this means we lose the error/info + status message.......

samlaf: Thinking switching to echo webserver as in https://github.com/Layr-Labs/eigenda-proxy/pull/142 would be a really good idea at this point.... @epociask let's discuss tomorrow

samlaf:
Closing as no longer needed. We can reopen if this becomes a requirement again in the future.

PR #150, opened 2024-09-24
Update Optimism dependency to OP-Stack 1.9.2

karlb:
Updates:
* optimism v1.9.0 -> v1.9.2
* op-geth to the commit 110c433a2469 which is used by optimism v1.9.2
* removed unnecessary version replacements * go-kzg-4844 now uses v1.0.0 * pebble now uses v1.1.2

Necessary code changes:
* opio is now ctxinterrupt
* plasma is now alt-da


karlb:
Is the holesky-test supposed to work for me in CI? I only see the `SIGNER_PRIVATE_KEY environment variable not set` error and don't think I caused that by the changes in this PR.

samlaf:
@karlb just merged #153 which should allow your PR to read the secret key. Tried changing the target branch but it seems like the workflow is still running in the context of your PR's outdated workflow. Can you try rebasing on top of main to get access to new workflow?

karlb:
@samlaf I did the rebase and the holesky-test now succeeds. The other tests need approval to run in CI, but I don't think I've broken them during the rebase, so they should also pass.

samlaf (approved):
Thanks @karlb !!

PR #155, opened 2024-09-27
fix: don't allow to configure fallback or cache with nil Redis|S3 ins…

mcortesi:
To avoid user confussion, check that variables are not nil before adding them to potential sources.

If not, they will when we actually try to use them

This is because `populateTargets()` accepts `redis` and `s3` variables to be `nil`.


samlaf (commented):
LGTM but @epociask is the codeowner for this feature.

epociask (approved):
this is a good callout, thanks for catching :)

PR #157, opened 2024-09-27
feat: Avoid writing twice to the same storage backend

mcortesi:
`handleRedundantWrites` will do a write for each source in `caches` and `fallbacks`.

While probably doesn't make much sense to have the same type of store be `cache` and `fallback` they can technically be the same. 

So maybe check that `caches ++ fallbacks` is an unique set.

Alternative, on configuration processing check that a source can't be at same time `cache` and `fallback`

### DIFF
--- a/store/router.go
+++ b/store/router.go
@@ -157,8 +157,7 @@ func (r *Router) handleRedundantWrites(ctx context.Context, commitment []byte, v
 		r.fallbackLock.RUnlock()
 	}()
 
-	sources := r.caches
-	sources = append(sources, r.fallbacks...)
+	sources := r.uniqueSources()
 
 	key := crypto.Keccak256(commitment)
 	successes := 0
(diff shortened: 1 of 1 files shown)


samlaf:
So we treat sources and fallbacks the same when writing? So I guess they are only differentiated by their read sequence logic? This looks fine to me, though I do have a strong preference for the config approach (KISS).

```
Alternative, on configuration processing check that a source can't be at same time cache and fallback
```

I'll let @epociask review this though given he wrote this feature.

epociask:
Hi @mcortesi - there's already a config invariant to ensure that this case can never be true: https://github.com/Layr-Labs/eigenda-proxy/blob/main/server/config.go#L115-L120

mcortesi:
@epociask amazin, haven't seen that.. maybe a comment on the code will help for the unaware reader

epociask:
Closing PR since verifications are performed at the cfg invariant enforcement

PR #159, opened 2024-09-28
fix: verifier srs flags paths

samlaf:
Made a mistake in https://github.com/Layr-Labs/eigenda-proxy/pull/146

paths should have remained the same given that they are relative to where the binary is run from, not where the go file lives. Fixed and added a comment to that effect.

### DIFF
--- a/verify/cli.go
+++ b/verify/cli.go
@@ -86,18 +86,18 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			Category: category,
 		},
 		&cli.StringFlag{
-			Name:     G2TauFlagName,
-			Usage:    "Directory path to g2.point.powerOf2 file.",
-			EnvVars:  withEnvPrefix(envPrefix, "TARGET_G2_TAU_PATH"),
+			Name:    G2TauFlagName,
+			Usage:   "Directory path to g2.point.powerOf2 file.",
+			EnvVars: withEnvPrefix(envPrefix, "TARGET_G2_TAU_PATH"),
 			// we use a relative path so that the path works for both the binary and the docker container
 			// aka we assume the binary is run from root dir, and that the resources/ dir is copied into the working dir of the container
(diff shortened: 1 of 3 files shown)


verify/cli.go
```diff
@@ -80,23 +80,27 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			Name:    G1PathFlagName,
 			Usage:   "Directory path to g1.point file.",
 			EnvVars: withEnvPrefix(envPrefix, "TARGET_KZG_G1_PATH"),
-			// TODO: should use absolute path wrt root directory to prevent future errors
-			//       in case we move this file around
-			Value:    "../resources/g1.point",
+			// we use a relative path so that the path works for both the binary and the docker container
+			// aka we assume the binary is run from root dir, and that the resources/ dir is copied into the working dir of the container
```
bxue-l2, line 84: I thought binary is run from the workdir, which is /app

bxue-l2 (approved):

verify/cli.go
```diff
@@ -80,23 +80,27 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
...
+			Value:    "resources/g2.point.powerOf2",
 			Category: category,
 		},
 		&cli.StringFlag{
 			Name:     CachePathFlagName,
 			Usage:    "Directory path to SRS tables for caching.",
 			EnvVars:  withEnvPrefix(envPrefix, "TARGET_CACHE_PATH"),
-			Value:    "../resources/SRSTables/",
+			// we use a relative path so that the path works for both the binary and the docker container
+			// aka we assume the binary is run from root dir, and that the resources/ dir is copied into the working dir of the container
```
bxue-l2, line 102: same as above,

verify/cli.go
samlaf: That’s the binary in the container. By binary here I just meant if we make + .//bin/… or go run

PR #160, opened 2024-09-29
tests: dont wait for finality in e2e tests

### LINKED ISSUE #154: Make holesky-test suite faster
~testnet has 10 min batching intervals, which I think (?) is why the holesky-e2e tests currently take  roughly 20 mins to run. preprod cluster has 1 min batching, which should reduce the ci time for that test.~

holesky e2e test suite takes >15mins to run. Makes waiting for ci unbearably long, and devex experience pretty bad. We should find a way to optimize these tests.

samlaf:
## Fixes Issue

Fixes #154 

## Changes proposed

not sure if this even works, but trying to not wait for finality to speed up the e2e holesky tests.
There's currently no way to run the tests locally since we don't have the private key we use in the tests stored anywhere accessible (can't read github secrets). Might be nice to put it in our shared password vault.

UPDATE: seems like e2e tests still took 14 mins without waiting for finality... so there's prob something else cause them to be super slow.

### DIFF
--- a/e2e/setup.go
+++ b/e2e/setup.go
@@ -120,6 +120,7 @@ func TestSuiteConfig(t *testing.T, testCfg *Cfg) server.CLIConfig {
 			StatusQueryRetryInterval: pollInterval,
 			DisableTLS:               false,
 			SignerPrivateKeyHex:      pk,
+			WaitForFinalization:      false,
 		},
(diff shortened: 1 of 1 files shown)


samlaf:
Closing this as stale, also hasn't been a major issue lately afaiu.

PR #163, opened 2024-10-01
make backward compatible

bxue-l2:
Make env backward compatible.

Method: Adding multiple EnvVars

The urfave doc shows, https://cli.urfave.org/v2/examples/flags/#values-from-the-environment
If EnvVars contains more than one string, the first environment variable that resolves is used.

If this approach looks fine, I will go ahead to change the template env, https://github.com/Layr-Labs/eigenda-proxy/blob/main/.env.example.holesky

Also add a new column to he configuration option in the readme [page](https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#configuration-options). but @samlaf , do you want to take the rest?

sample

Screenshot 2024-10-01 at 11 59 10 AM

### DIFF
--- a/flags/eigendaflags/cli.go
+++ b/flags/eigendaflags/cli.go
@@ -28,7 +28,7 @@ func withFlagPrefix(s string) string {
 }
 
 func withEnvPrefix(envPrefix, s string) []string {
-	return []string{envPrefix + "_EIGENDA_" + s}
+	return []string{envPrefix + "_EIGENDA_CLIENT_" + s, envPrefix + "_" + s}
 }
 
 // CLIFlags ... used for EigenDA client configuration
--- a/store/precomputed_key/s3/cli.go
+++ b/store/precomputed_key/s3/cli.go
@@ -32,7 +32,7 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 		&cli.StringFlag{
 			Name:     EndpointFlagName,
 			Usage:    "endpoint for S3 storage",
-			EnvVars:  withEnvPrefix(envPrefix, "S3_ENDPOINT"),
+			EnvVars:  withEnvPrefix(envPrefix, "ENDPOINT"),
 			Category: category,
 		},
 		&cli.StringFlag{
--- a/verify/cli.go
+++ b/verify/cli.go
@@ -41,7 +41,7 @@ func withFlagPrefix(s string) string {
 }
 
 func withEnvPrefix(envPrefix, s string) []string {
-	return []string{envPrefix + "_EIGENDA_" + s}
+	return []string{envPrefix + "_EIGENDA_CLIENT_" + s, envPrefix + "_" + s}
 }
 
 // CLIFlags ... used for Verifier configuration


flags/eigendaflags/cli.go
```diff
@@ -28,7 +28,7 @@ func withFlagPrefix(s string) string {
 }
 
 func withEnvPrefix(envPrefix, s string) []string {
-	return []string{envPrefix + "_EIGENDA_" + s}
+	return []string{envPrefix + "_EIGENDA_CLIENT_" + s, envPrefix + "_" + s}
```
samlaf, line 31: Add comment explaining that second one is just for backward compatibility. We should also prob add a deprecation notice? Maybe in the category string?

Also are we sure this makes everything backward compatible? Did some of them not have just EIGeNDA prefix for eg? Also flags themselves can’t have multiple options so those won’t be backward compatible unless we add more flags (which I am against… but wdyt @jianoaix )

bxue-l2: I went through flags one by one, they are backward compatible, except for memstore. But since memstore is only used in testnet, so it is fine to break it.

samlaf: Ah true for memstore. But I’m pretty sure the eigenDA flags were —eigenda- before whereas now they are —eigenda.

bxue-l2: Adding a deprecation note is good. We could put it in the readme part, https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#configuration-options

bxue-l2: so in the readme, we specify the canonical format of the env variable, but also note we embrace backward compatibility

samlaf: That’s fine. Long term I personally really don’t like that README thing because it’s very annoying to generate, update and keep in sync. Would prefer if it was autogenerated (maybe we can just run the binary with —help and dump the help output in the readme if that’s what we want)

jianoaix: SG. Also +1 to add deprecation notice. Not sure if the flag framework supports auto notice; if not, maybe we can emit logs if the old flag is used to start the binary.

jianoaix: I think in Python they have a pattern (actually even library like https://pypi.org/project/deprecation/) using a `@deprecated` decorator  to annotate the things you want to deprecate. We may at least 1) document, e.g. specifying when it's deprecated, and when it'll be removed with a release version; 2) log the attempts to use deprecated API/flags.

samlaf: @bxue-l2 thought about this more and I think I agree with @jianoaix now. Doing multiple env_vars on the same flag is not the way to go, because that makes it very hard to give deprecation notices.

If we create new flags instead, then we can use flag actions to print a deprecation notice on the old flags. And when we release 2.0.0 we can remove the deprecated flags.

bxue-l2: I agree. I can close this PR. There 2 solutions I can see is to 
Plan A: revert the flags PRs
Plan B:
1. hand pick the old env variables (i.e. those that have new names)
2. keep the new names, create a sections about the list of flags for old names

samlaf: Agree. I think Plan B works best, because the flags PR contains a lot of other flag refactors... will be hard to dismantle properly. I'm thinking we create a new category EIGENDA_DEPRECATED_FLAGS or something (so that these flags don't clutter the --help output), and add all the backward compatible deprecated flags there. Should be in its own file probably, such that when we bump to 2.0.0 we can just delete that file to easily remove all of them.

PR #167, opened 2024-10-04
Disable chunk encoding for `put` requests to Google Cloud Storage

jcortejoso:
## Fixes Issue

Fixes #

## Changes proposed

The PR fixes the upload process on Google Storage (using its S3 compatibility API). Right now (with `main` config) the objects uploaded to Google Storage can have multiple `<size>;chunk-signature=<bytes>` strings in the file, which breaks the verification of the file on the get() stage.
Adding `DisableContentSha256 = true` to `putObjectOptions` used by minio client, the upload seems to be done properly and the chunk-signatures not included.

More info [here](https://github.com/minio/minio-go/issues/1922).

### DIFF
--- a/store/precomputed_key/s3/s3.go
+++ b/store/precomputed_key/s3/s3.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"io"
+	"net/url"
 	"path"
 	"time"
 
@@ -14,6 +15,7 @@ import (
 	"github.com/minio/minio-go/v7"
 
 	"github.com/minio/minio-go/v7/pkg/credentials"
+	"github.com/minio/minio-go/v7/pkg/s3utils"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


store/precomputed_key/s3/s3.go
```diff
@@ -97,7 +99,16 @@ func (s *Store) Get(ctx context.Context, key []byte) ([]byte, error) {
...
-	_, err := s.client.PutObject(ctx, s.cfg.Bucket, path.Join(s.cfg.Path, hex.EncodeToString(key)), bytes.NewReader(value), int64(len(value)), minio.PutObjectOptions{})
+	endpointURL, err := url.Parse(s.cfg.Endpoint)
+	if err != nil {
+		return err
+	}
+
+	putObjectOptions := minio.PutObjectOptions{}
+	if s3utils.IsGoogleEndpoint(*endpointURL) {
+		putObjectOptions.DisableContentSha256 = true // Avoid chunk signatures on GCS: https://github.com/minio/minio-go/issues/1922
+	}
```
epociask, line 110: its more efficient to do this check within the constructor and embed `putObjectOptions` as a field into the storage struct versus as an additional logic flow that adds latency to every insertion

samlaf (commented):
LGTM but I'll let @epociask approve

bxue-l2:
looks good to me too

epociask (approved):
LGTM

PR #168, opened 2024-10-04
fix: change put route put/ -> put

### LINKED ISSUE #88: PUT endpoint without trailing slash does nothing
The [OP alt-DA spec](https://specs.optimism.io/experimental/alt-da.html#da-server) requires
```
Request:
  POST /put
  Content-Type: application/octet-stream
  Body: <preimage_bytes>

Response:
  200 OK
  Content-Type: application/octet-stream
  Body: <hex_encoded_commitment>
```

The EigenDA proxy does not react on that URL and silently returns nothing, but it does work when appending a slash to the URL (which is not included in the spec).

```
karl@PVTY25Y2GX ~> curl -X POST http://127.0.0.1:3100/put --data foo -s | wc
       0       0       0
karl@PVTY25Y2GX ~> curl -X POST http://127.0.0.1:3100/put/ --data foo -s | wc
       0      13     187
```

Suggested changes:
* Respond with commitment at `/put`
* Respond with HTTP code 404 on unknown URLs

PS: The response does also not look hex-encoded, but that is a separate issue

samlaf:
## Fixes Issue

Fixes #88 

## Note to reviewers

Updated to optimism 1.9.3 which required also updating to go1.22


samlaf:
cc @karlb for review. Just applied the patch you had shared with me here.

samlaf:
Not sure why the test is still failing. @karlb said he would take a look tomorrow.

server/server.go
```diff
@@ -99,8 +97,12 @@ func WithLogging(
...
 	mux := http.NewServeMux()
 
-	mux.HandleFunc(GetRoute, WithLogging(WithMetrics(svr.HandleGet, svr.m), svr.log))
-	mux.HandleFunc(PutRoute, WithLogging(WithMetrics(svr.HandlePut, svr.m), svr.log))
+	mux.HandleFunc("/get/", WithLogging(WithMetrics(svr.HandleGet, svr.m), svr.log))
+	// we need to handle both: see https://github.com/ethereum-optimism/optimism/pull/12081
+	// /put is for generic commitments, and /put/ for keccak256 commitments
+	// TODO: we should probably separate their handlers?
+	mux.HandleFunc("/put", WithLogging(WithMetrics(svr.HandlePut, svr.m), svr.log))
+	mux.HandleFunc("/put/", WithLogging(WithMetrics(svr.HandlePut, svr.m), svr.log))
```
samlaf, line 105: @epociask can you review that this comment and the routes make sense. tests are passing but still want to be 100% certain since I still don't get our routing logic with the different paths. Should probably document this api at some point..

go.mod
```diff
@@ -1,17 +1,19 @@
...
+go 1.22
+
+toolchain go1.22.0
 
 require (
 	github.com/Layr-Labs/eigenda v0.8.4
 	github.com/consensys/gnark-crypto v0.12.1
-	github.com/ethereum-optimism/optimism v1.9.2
-	github.com/ethereum/go-ethereum v1.14.8
+	github.com/ethereum-optimism/optimism v1.9.4-0.20240927020138-a9c7f349d10b
```
bxue-l2, line 10: @samlaf Remember to change to a canonical tag after 1.9.4 is released by OP

bxue-l2 (approved):
lgtm

PR #177, opened 2024-10-08
Improve local test setup

### LINKED ISSUE #175: improve e2e tests
bunch of small improvements that we can make to e2e test suite:

- want to run tests manually from vscode 
   - run minio and redis as testcontainers instead of starting them from makefile command
   - use test labels instead of requiring env var to be set

- to run holesky tests locally
   - need a signer private key in password vault that we can use, instead of relying on github secrets cc @bxue-l2

samlaf:
## Fixes Issue

Fixes #175 

at least fixes everything with e2e tests. holesky tests are still not runnable locally until we get that private key stored somewhere.


Makefile
```diff
@@ -51,15 +32,11 @@ clean:
 test:
 	go test ./... -parallel 4 
 
-e2e-test: stop-minio stop-redis run-minio run-redis
```
epociask, line 54: nice I did this cuz of time press and wasn't able to codify the resources - happy to see it get fixed 💯

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
epociask, line 1: so which IDE do you use again?

e2e/main_test.go
```diff
@@ -5,6 +5,9 @@ import (
 	"testing"
 )
 
+// only difference between integrationTests and testnetIntegrationTests is
+// that integration tests are run against memstore whereas testnetintegration tests
+// are run against eigenda backend talking to testnet disperser
```
epociask, line 10: this does change the conditional behavior of some of the tests - e.g, in `TestProxyServerCaching` we only assert to read metrics with EigenDA when referencing memstore since we don't profile the `eigenDAClient` interactions

e2e/setup.go
```diff
@@ -37,6 +40,66 @@ const (
...
+func init() {
+	err := startMinioContainer()
+	if err != nil {
+		panic(err)
+	}
+	err = startRedisContainer()
+	if err != nil {
+		panic(err)
+	}
+}
```
epociask, line 59: knit - using an init pattern like this could make our code super hard to import - at least this library

e2e/setup.go
```diff
@@ -26,6 +27,8 @@ import (
 	opmetrics "github.com/ethereum-optimism/optimism/op-service/metrics"
 
 	"github.com/stretchr/testify/require"
+	miniotc "github.com/testcontainers/testcontainers-go/modules/minio"
+	redistc "github.com/testcontainers/testcontainers-go/modules/redis"
```
epociask, line 31: nice didn't even know that this existed

epociask (commented):
generally lgtm - just a few minor q's and knits

e2e/setup.go
```diff
@@ -37,6 +40,66 @@ const (
...
+	}
+}
+
+// startMinioContainer starts a MinIO container and returns the container instance and its endpoint
+func startMinioContainer() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	minioContainer, err := miniotc.Run(ctx,
+		"minio/minio:latest",
```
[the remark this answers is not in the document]
samlaf: good point: [7596b17](https://github.com/Layr-Labs/eigenda-proxy/pull/177/commits/7596b17d1b210cb5f550cf53d6beedec9fe0ba58)

.vscode/settings.json
samlaf: vscode... are you asking if I use some fork of vscode? I just use the normal vscode

e2e/setup.go
samlaf: yeah they have a bunch. I typically reach out for these premades when they exist.

e2e/setup.go
samlaf: agree that this sucks... also it could at some point be needed to have new clean instances per test. wanted to talk with you about how to best refactor this.

I didn't see an obvious place where to start these containers. The separation between creating the config (one fct that sets all the config vars) and starting the endpoints makes it hard... because the actual endpoint of the containers are only known after they are started, so we'd need to start them in the "createConfig" function, which feels wrong... perhaps we need yet another function like `startDependencyContainers()` or something? And we can call it from the test main function?

samlaf: On another note though I'm not sure why anyone would want to import this package... do you have anything in mind or just saying in case?

e2e/main_test.go
samlaf: [7ccf0c1](https://github.com/Layr-Labs/eigenda-proxy/pull/177/commits/7ccf0c191981d6fdd52daf74859800067d7b7a97)

.vscode/settings.json
epociask: it was supposed to be a joke

e2e/setup.go
epociask: just saying as an in-case - maybe one day we try importing this into node software directly for cross testing framework integration

epociask (approved):
LGTM - just very knits

e2e/setup.go
```diff
@@ -37,6 +40,66 @@ const (
...
+		panic(err)
+	}
+	err = startRedisContainer()
+	if err != nil {
+		panic(err)
+	}
+}
+
+// startMinioContainer starts a MinIO container and returns the container instance and its endpoint
+func startMinioContainer() error {
```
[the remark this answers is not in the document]
samlaf: [e37411f](https://github.com/Layr-Labs/eigenda-proxy/pull/177/commits/e37411f81c2a904efe327f64b621ef58c078ecda)

e2e/setup.go
```diff
@@ -37,6 +40,66 @@ const (
...
+
+	endpoint, err := minioContainer.Endpoint(ctx, "")
+	if err != nil {
+		return fmt.Errorf("failed to get MinIO endpoint: %w", err)
+	}
+
+	minioEndpoint = strings.TrimPrefix(endpoint, "http://")
+	return nil
+}
+
```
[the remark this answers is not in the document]
samlaf: [e37411f](https://github.com/Layr-Labs/eigenda-proxy/pull/177/commits/e37411f81c2a904efe327f64b621ef58c078ecda)

e2e/setup.go
samlaf: Agree here, but want to move on to other things for now and worry about this when it actually becomes a problem (perhaps an obvious refactor will also surface once we have an actual need for this). Added a TODO comment: [a9ae57b](https://github.com/Layr-Labs/eigenda-proxy/pull/177/commits/a9ae57b85a3da6a845895b699fb22684f004f728)

PR #179, opened 2024-10-09
Allow Google storage endpoints without http/s

jcortejoso:
## Fixes Issue

The line `endpointURL, err := url.Parse(cfg.Endpoint)` does not properly parse the URL if the parameter does not include the protocol (http/https). This is problematic because the protocol cannot be included when initializing the S3/MinIO client; otherwise, you get the following error:

```
t=2024-10-09T08:31:27+0000 lvl=crit msg="Application failed" role=eigenda_proxy message="failed to create store: failed to create S3 store: Endpoint url cannot have fully qualified paths."
```

## Changes proposed

A small refactor to simplify the Google Storage endpoint checker.

## Note to reviewers

For the simple tests added, I’ve only focused on the `isGoogleEndpoint()` method. While https://storage.googleapis.com is not a valid URL for `minio.New()`, I skipped checking for that in this function. If you'd prefer, I can include a check in `server/config.go` to ensure that the protocol is absent in the config provided.
Sorry about the noise/multiple PRs.

### DIFF
--- a/store/precomputed_key/s3/s3.go
+++ b/store/precomputed_key/s3/s3.go
@@ -6,8 +6,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"io"
-	"net/url"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/Layr-Labs/eigenda-proxy/store"
@@ -15,7 +15,6 @@ import (
 	"github.com/minio/minio-go/v7"
 
 	"github.com/minio/minio-go/v7/pkg/credentials"
-	"github.com/minio/minio-go/v7/pkg/s3utils"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


store/precomputed_key/s3/s3.go
```diff
@@ -58,13 +57,13 @@ type Store struct {
 	stats            *store.Stats
 }
 
+func isGoogleEndpoint(endpoint string) bool {
+	return strings.Contains(endpoint, "storage.googleapis.com")
+}
+
```
samlaf, line 63: seems fine. if it's always going to be https://storage.googleapis.com, maybe trim the protocol instead of doing this?

jcortejoso: So right now it requires to be submitted without `https` for minio/client (i.e.: `--s3.endpoint=storage.googleapis.com`).

Another solution would be ensuring that the protocol is present before using `url.Parse()` and keep `s3utils.IsGoogleEndpoint`. In minio, they use a [helper method](https://github.com/minio/minio-go/blob/cca410398dd20409180d4ff2d017e34bea17c27c/utils.go#L107) for this.

samlaf: When/why do you ever need the full protocol url? And why does minio not allow it (I’ve noticed this also and found it annoying). I’m just curious at this point, but I’ll approve this, sounds fine for now

samlaf (approved):

store/precomputed_key/s3/s3.go
jcortejoso: So it's just needed for `url.Parse()` here. I think it may depend on each client implementation, but seems `minio-go` client don't handle if the endpoint has the protocol on it.

PR #181, opened 2024-10-11
feat: add openapi spec

samlaf:
Thinking this would be a nice addition to have, and would enable:
1) being precise about eigenda-proxy endpoint
2) enforcing the spec + testing our implementation using https://specmatic.io/documentation.html
3) proposing new features such as fallback mechanism

## Fixes Issue

Fixes #

## Note to reviewers

This is a wip. I've only tested the /put/0x... keccak endpoint using specmatic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ stop-redis:
 	fi
 
 run-memstore-server:
-	./bin/eigenda-proxy --memstore.enabled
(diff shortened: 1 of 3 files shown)


samlaf:
Closing as no longer needed. This would require too much work to keep in sync at this point.

Issue #164, opened 2024-10-02
Update storage router to write blobs when reading


epociask:
**Problem**
Currently secondary storage locations are only utilized after dispersing a blob through proxy. This doesn't compensate for reading or verifier/syncing nodes who operate irrespective of the batcher. Furthermore the operation of writing to secondary storage targets increases response latency. 

**Proposed Solution**
Add a secondary storage routine which transits blobs seen either via dispersal or retrieval from EigenDA to cache and fallback targets if and only if they haven't been persisted to the secondary storage target before.

samlaf:
Thinking through this: is this feature actually useful? Derivation only needs to read this once right? Unless there’s a reorg or a challenge. But is it worth paying the extra write cost in the happy path for these rare cases?

Issue #96, opened 2024-08-21
Metrics: Support configurable prometheus histogram bucket sizes for request duration metric


samlaf:
See https://github.com/Layr-Labs/eigenda-proxy/pull/94#discussion_r1723900596

Metrics for request durations currently have a hardcoded histogram bucket size, which is currently tailored towards eigenda latencies).
Would be nice to let users set them as needed, such that different configurations of the proxy server could use different buckets sizes (memstore, s3, and eigenda backends have very different latencies).

PR #182, opened 2024-10-12
feat: Optional async decoupling for secondary writes and reworked E2E metric assertions

### LINKED ISSUE #113: Remove stats profiling in favor of prometheus

epociask:
## Fixes Issue
Related to https://github.com/Layr-Labs/eigenda-proxy/issues/164 but doesn't fix it. 

Fixes #113 and #96 

## Changes proposed
- Optional async decoupling for write routines to reduce dispersal latencies when writing to EigenDA. User can optionally configure the number of go routines. 
- Decompose router abstraction into primary router with access to a secondary one which manages secondary writing/reading
- Remove stats struct profiling in-favor of in-memory implementation of Metricer interface for E2E assertions 
- Add retry logic for secondary insertions
- Decomposed tests into cleaner and more concise modules
- Added async benchmark function 

### Benchmarking
As expected, latency decreases when using async writes to secondary storage. 

0 threads: 
```
     240	   6362160 ns/op	  412053 B/op	     624 allocs/op
```

5 threads:
```
     758	   1508104 ns/op	  218021 B/op	     620 allocs/op
```

10 threads:
```
    1742	    914375 ns/op	  167984 B/op	     619 allocs/op
```


store/secondary.go
```diff
@@ -0,0 +1,183 @@
...
+	defer r.fallbackLock.RUnlock()
+
+	return len(r.fallbacks) > 0
+}
+
+// handleRedundantWrites ... writes to both sets of backends (i.e, fallback, cache)
+// and returns an error if NONE of them succeed
+// NOTE: multi-target set writes are done at once to avoid re-invocation of the same write function at the same
+// caller step for different target sets vs. reading which is done conditionally to segment between a cached read type
+// vs a fallback read type
```
samlaf, line 81: can you rephrase? very long sentence, I don't understand it.

store/secondary.go
```diff
@@ -0,0 +1,183 @@
...
+		err := src.Put(ctx, key, value)
+		if err != nil {
+			r.log.Warn("Failed to write to redundant target", "backend", src.BackendType(), "err", err)
+		} else {
+			successes++
+		}
+	}
+
+	if successes == 0 {
+		return errors.New("failed to write blob to any redundant targets")
```
samlaf, line 107: should we use an errgroup here instead? If there are a bunch of sources, we should prob write to them in parallel?

store/secondary.go
```diff
@@ -0,0 +1,183 @@
...
+type ISecondary interface {
+	Fallbacks() []PrecomputedKeyStore
+	Caches() []PrecomputedKeyStore
+	Enabled() bool
+	Ingress() chan<- PutNotif
+	CachingEnabled() bool
+	FallbackEnabled() bool
+	HandleRedundantWrites(ctx context.Context, commitment []byte, value []byte) error
+	MultiSourceRead(context.Context, []byte, bool, func([]byte, []byte) error) ([]byte, error)
+	StreamProcess(context.Context)
```
samlaf, line 22: add comments to explain the interface. Not sure how this works from first reading. Ingress, StreamProcess, etc. are not super descriptive. Might want to use longer more descriptive names?

server/load_store.go
```diff
@@ -116,10 +116,22 @@ func LoadStoreRouter(ctx context.Context, cfg CLIConfig, log log.Logger) (store.
...
+	if err != nil {
+		return nil, err
+	}
+
+	if secondary.Enabled() { // only spin-up go routines if secondary storage is enabled
+		log.Debug("Starting secondary stream processing routines")
+
+		for i := 0; i < 10; i++ {
+			go secondary.StreamProcess(ctx)
+		}
```
samlaf, line 132: am I understanding correctly that you're spinning up 10 workers to pull from the queue? Are we sure that 1 is not enough? Why is 10 sufficient? Can we make it a config parameter instead. Also probably change the name StreamProcess to RequestProcessor or something more explicit?

samlaf: Also maybe if you want to go this route to expose an `init()` function or `startWorkers(numWorkers uint64)` function which spins up the goroutines?

store/router.go
```diff
@@ -98,8 +91,8 @@ func (r *Router) Get(ctx context.Context, key []byte, cm commitments.CommitmentM
 		}
 
 		// 3 - read blob from fallbacks if enabled and data is non-retrievable from EigenDA
-		if r.fallbackEnabled() {
-			data, err = r.multiSourceRead(ctx, key, true)
+		if r.secondary.FallbackEnabled() {
+			data, err = r.secondary.MultiSourceRead(ctx, key, true, r.eigenda.Verify)
```
samlaf, line 95: Thinking through fallbacks some more. Do we really need them? I can't think of a use case when I would want a fallback instead of a cache. Why would I only want to read after eigenDA, instead of always read before eigenDA? Would greatly simplify the code if we could just get rid of fallbacks. But I might be missing something here still..

samlaf (commented):
I like the refactor! Think there are ways to make the API for secondary struct easier to understand and use though..! I'm not a big fan of async APIs. See my reasoning here: https://www.notion.so/eigen-labs/Bls-Agg-Service-2-0-synchronous-API-98abef46040a48fc8d044e7de0781839

README.md
```diff
@@ -65,6 +65,7 @@ In order to disperse to the EigenDA network in production, or at high throughput
 | `--s3.enable-tls` |  | `$EIGENDA_PROXY_S3_ENABLE_TLS` | Enable TLS connection to S3 endpoint. |
 | `--routing.fallback-targets` | `[]` | `$EIGENDA_PROXY_FALLBACK_TARGETS` | Fall back backend targets. Supports S3. | Backup storage locations to read from in the event of eigenda retrieval failure. |
 | `--routing.cache-targets` | `[]` | `$EIGENDA_PROXY_CACHE_TARGETS` | Caching targets. Supports S3. | Caches data to backend targets after dispersing to DA, retrieved from before trying read from EigenDA. |
+| `--routing.concurrent-write-threads` | `0` | `$EIGENDA_PROXY_CONCURRENT_WRITE_THREADS` | Number of threads spun-up for async secondary storage insertions. (<=0) denotes single threaded insertions where (>0) indicates decoupled writes. |
```
samlaf, line 68: I hadn't raised this before but I don't really like the use of "routing" in this context. What was your motivation for calling it routing? Routing for me is the thing the http library does. This is more of a storage layer abstraction. Shouldn't we just call these flags `--storage.concurrent-write-threads`? or `--data.` as in the three-tier architecture.

commitments/mode.go
```diff
@@ -8,7 +8,7 @@ import (
 type CommitmentMeta struct {
 	Mode CommitmentMode
 	// CertVersion is shared for all modes and denotes version of the EigenDA certificate
-	CertVersion byte
+	CertVersion uint8
```
samlaf, line 11: yes, thank you for this

metrics/memory.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/ethereum-optimism/optimism/op-service/metrics"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+func keyLabels(labels []string) (common.Hash, error) {
+	sort.Strings(labels) // in-place sort strings so keys are order agnostic
+
+	encodedBytes, err := rlp.EncodeToBytes(labels)
```
samlaf, line 17: why use rlp instead of something more efficient/common?

metrics/memory.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return common.Hash{}, err
+	}
+
+	hash := crypto.Keccak256Hash(encodedBytes)
```
samlaf, line 22: also prob don't need a crypto hash here.

metrics/memory.go
```diff
@@ -0,0 +1,119 @@
...
+	// secondary metrics
+	SecondaryRequestsTotal *MetricCountMap
+}
+
+func NewInMemoryMetricer() *InMemoryMetricer {
+	return &InMemoryMetricer{
+		HTTPServerRequestsTotal: NewCountMap(),
+		SecondaryRequestsTotal:  NewCountMap(),
+	}
+}
```
samlaf, line 89: add docstrings above struct and constructor explaining that this is meant to be used for tests only, and that it's needed because prometheus structs don't allow Getting their values (basically what your Find method above implements)

metrics/memory.go
```diff
@@ -0,0 +1,119 @@
...
+	uint64Val, ok := value.(uint64)
+	if !ok {
+		return fmt.Errorf("could not read uint64 from sync map")
+	}
+
+	mcm.m.Store(key.Hex(), uint64Val+uint64(1))
+	return nil
+}
+
+func (mcm *MetricCountMap) Find(values ...string) (uint64, error) {
```
samlaf, line 59: Why `Find` and not `Get`, or `Read`? Find makes me think of finding the index, not the value.

store/store.go
```diff
@@ -75,6 +73,8 @@ type Store interface {
 
 type GeneratedKeyStore interface {
 	Store
+	// Stats returns the current usage metrics of the key-value data store.
+	Stats() *Stats
```
samlaf, line 77: Why is this still needed? the PR readme says that this PR fixes #113 but then here you still use Stats? When are the metrics used vs Stats here?

store/secondary.go
```diff
@@ -0,0 +1,182 @@
...
+type ISecondary interface {
+	AsyncEntry() bool
+	Enabled() bool
+	Topic() chan<- PutNotify
+	CachingEnabled() bool
+	FallbackEnabled() bool
+	HandleRedundantWrites(ctx context.Context, commitment []byte, value []byte) error
+	MultiSourceRead(context.Context, []byte, bool, func([]byte, []byte) error) ([]byte, error)
+	WriteSubscriptionLoop(ctx context.Context)
+}
```
samlaf, line 33: This probably works. I personally feel like we're getting wayyyy to complex way too early, for something that's super simple: we only have s3 and redis options, but now we have all these indirection layers, caching vs fallback, interfaces over interfaces. Makes it very hard to understand the code, review it, find bugs, benchmark/profile and optimize, etc. Let's merge this, but I'd like to try to keep the complexity lower in the future, or at least think through ways of simplifying the architecture.

store/store.go
epociask: Stats is used for Memstore

README.md
epociask: https://github.com/Layr-Labs/eigenda-proxy/issues/191

store/store.go
epociask: but is removed entirely now

metrics/memory.go
samlaf: I meant can use a standard (non-cryptographic) hash function. But I'm bikeshedding

e2e/setup.go
```diff
@@ -113,6 +111,7 @@ type Cfg struct {
 	UseS3Caching       bool
 	UseRedisCaching    bool
 	UseS3Fallback      bool
+	WriteThreadCount   int
```
samlaf, line 114: prob makes the comment above ("at most one of the below options should be true") not true anymore. Maybe update the comment to say "below 3/4 options"

e2e/setup.go
```diff
@@ -152,24 +152,23 @@ func createS3Config(eigendaCfg server.Config) server.CLIConfig {
...
 	}
 }
 
-func TestSuiteConfig(t *testing.T, testCfg *Cfg) server.CLIConfig {
+func TestSuiteConfig(testCfg *Cfg) server.CLIConfig {
 	// load signer key from environment
 	pk := os.Getenv(privateKey)
 	if pk == "" && !testCfg.UseMemory {
-		t.Fatal("SIGNER_PRIVATE_KEY environment variable not set")
+		panic("SIGNER_PRIVATE_KEY environment variable not set")
```
samlaf, line 165: why panic instead of t.Fatal?

e2e/setup.go
```diff
@@ -180,7 +179,10 @@ func TestSuiteConfig(t *testing.T, testCfg *Cfg) server.CLIConfig {
 	}
 
 	maxBlobLengthBytes, err := utils.ParseBytesAmount("16mib")
-	require.NoError(t, err)
+	if err != nil {
+		panic(err)
+	}
```
samlaf, line 184: same here, why change to the wordier if/panic?

e2e/setup.go
```diff
@@ -324,11 +332,32 @@ func createS3Bucket(bucketName string) {
...
+	var result any
+	switch any(new(T)).(type) {
+	case *string:
+		result = str
+	case *[]byte:
+		result = []byte(str)
+	}
+
+	return result.(T)
+}
```
samlaf, line 363: Think this is overcomplicating things and not necessary. Generics typically when behavior is the same, but types are different. Here we just want 2 different things. Why not just have 2 separate functions RandByte and RandString? Super straightforward, follows stdlib pattern. At your call site if I see the generic I'm not sure what's happening so I need to click through and then parse this complex function, for something that should be completely trivial and uninteresting.

server/middleware.go
```diff
@@ -33,6 +33,7 @@ func withMetrics(
 		// we assume that every route will set the status header
 		versionByte, err := parseVersionByte(w, r)
 		if err != nil {
+			recordDur(w.Header().Get("status"), string(mode), string(versionByte))
```
samlaf, line 36: This w.Header().Get("status") thing is buggy, I have an issue about this somewhere. But let's keep it like this for now and fix it in another PR.

samlaf (approved):
LGTM. Have a few small nits but want to get this PR merged asap!

e2e/setup.go
epociask: because we can longer pass `t` through via the constructor using this new pattern type

e2e/setup.go
epociask: [2d3559f](https://github.com/Layr-Labs/eigenda-proxy/pull/182/commits/2d3559f260ad130dbefd835f3ab13407a5494381)

e2e/setup.go
epociask: [2d3559f](https://github.com/Layr-Labs/eigenda-proxy/pull/182/commits/2d3559f260ad130dbefd835f3ab13407a5494381)

README.md
```diff
@@ -208,7 +208,11 @@ The `raw commitment` is an RLP-encoded [EigenDA certificate](https://github.com/
 
 ### Unit
 
-Unit tests can be ran via invoking `make test`.
+Unit tests can be ran via invoking `make test`.  Please make sure to have all test containers downloaded locally before running via:
+```
+docker pull redis
+docker pull minio
+```
```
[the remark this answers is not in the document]
epociask: [2d3559f](https://github.com/Layr-Labs/eigenda-proxy/pull/182/commits/2d3559f260ad130dbefd835f3ab13407a5494381)

PR #185, opened 2024-10-14
refactor: cleaner routing using gorilla mux regexps

samlaf:
This is still WIP, only refactored the GET methods so far, and still having trouble adapting the tests. But curious for opinions/criticisms at this point.

## Fixes Issue

Fixes #

## Changes proposed

Our routing is a bit convoluted and hard to follow. I'm working on an openapi spec in #181, which we could use with tests to enforce our actual golang router implements exactly. But our code is hard to read because the logic for the 3 types of commitments (simple, op generic, op keccak) is sprinkled all throughout different helper functions. Typically it's better to [push ifs up and fors down](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html).

go1.22's [revamped the stdlib router](https://go.dev/blog/routing-enhancements), but still doesn't have prefix or regexp abilities, which is why I opted to using gorilla mux, which is a minimal (in the sense that it's not an entire web framework we are taking on), but still very extensive routing library. In my opinion routing paths should be easy to read like openapi specs, not be sprinkled throughout if statements left and write inside helper functions. This PR makes routes like:
image


server/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+	return func(w http.ResponseWriter, r *http.Request) error {
+		recordDur := m.RecordRPCServerRequest(r.Method)
+
+		err := handleFn(w, r)
+		if err != nil {
+			var metaErr MetaError
+			if errors.As(err, &metaErr) {
+				recordDur(w.Header().Get("status"), string(metaErr.Meta.Mode), string(metaErr.Meta.CertVersion))
+			} else {
+				recordDur(w.Header().Get("status"), string("NoCommitmentMode"), string("NoCertVersion"))
```
epociask, line 29: what if we instead move these strings to introduce some `unknown` type for commitment mode and cert version enums?

server/routing.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+func (svr *Server) registerRoutes(r *mux.Router) {
+	subrouterGET := r.Methods("GET").PathPrefix("/get").Subrouter()
+	// simple commitments (for nitro)
+	subrouterGET.HandleFunc("/"+
+		"{optional_prefix:(?:0x)?}"+ // commitments can be prefixed with 0x
+		"{version_byte_hex:[0-9a-fA-F]{2}}"+ // should always be 0x00 for now but we let others through to return a 404
+		"{raw_commitment_hex:[0-9a-fA-F]*}",
+		withLogging(withMetrics(svr.handleGetSimpleCommitment, svr.m, commitments.SimpleCommitmentMode), svr.log),
```
epociask, line 18: what benefits do we reap by using the gorilla library? I.e, why use a heavy framework vs. what we had before. Generally agree with the structure of the refactoring but still a bit curious if we need to introduce a heavy framework with more developer complexity.

server/handlers.go
```diff
@@ -0,0 +1,209 @@
...
+func (svr *Server) handleHealth(w http.ResponseWriter, _ *http.Request) error {
+	w.WriteHeader(http.StatusOK)
+	return nil
+}
+
+// =================================================================================================
+// GET ROUTES
+// =================================================================================================
+
+func (svr *Server) handleGetSimpleCommitment(w http.ResponseWriter, r *http.Request) error {
```
epociask, line 25: knit - docstring

server/routing.go
epociask: also from a philosophical standpoint I've never seen ethereum projects use third party frameworks like this in RPC server implementations

server/handlers.go
```diff
@@ -0,0 +1,209 @@
...
+	// 	err = fmt.Errorf("error parsing version byte: %w", err)
+	// 	http.Error(w, err.Error(), http.StatusBadRequest)
+	// 	return err
+	// }
+	commitmentMeta := commitments.CommitmentMeta{
+		Mode:        commitments.OptimismKeccak,
+		CertVersion: byte(commitments.CertV0),
+	}
+
+	rawCommitmentHex, ok := mux.Vars(r)["raw_commitment_hex"]
```
epociask, line 62: lets capture "raw_commitment_hex" behind an env constant since its rawly declared 8 other times

server/handlers.go
```diff
@@ -0,0 +1,209 @@
...
+	if !ok {
+		return fmt.Errorf("commitment not found in path: %s", r.URL.Path)
+	}
+	commitment, err := hex.DecodeString(rawCommitmentHex)
+	if err != nil {
+		return fmt.Errorf("failed to decode commitment %s: %w", rawCommitmentHex, err)
+	}
+
+	svr.log.Info("Processing op keccak commitment", "commitment", rawCommitmentHex, "commitmentMeta", commitmentMeta)
```
epociask, line 153: we use this log for both dispersal and retrieval - should we add a method type to the string to allow for observable differentiation?

server/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		recordDur(w.Header().Get("status"), string(mode), string(versionByte))
+		return nil
+	}
+}
+
+// withLogging is a middleware that logs information related to each request.
+// It does not write anything to the response, that is the job of the handlers.
+// Currently we cannot log the status code because go's default ResponseWriter interface does not expose it.
+// TODO: implement a ResponseWriter wrapper that saves the status code: see https://github.com/golang/go/issues/18997
```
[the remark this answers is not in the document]
samlaf: https://github.com/Layr-Labs/eigenda-proxy/issues/186

server/handlers.go
samlaf: [a971fa4](https://github.com/Layr-Labs/eigenda-proxy/pull/185/commits/a971fa44c834fb776bb39d9ce64aaa8e2a9a22d4)

server/handlers.go
samlaf: [06034ff](https://github.com/Layr-Labs/eigenda-proxy/pull/185/commits/06034fffb9ac25d821e45b8af7737aaa07d8ddfe)

server/handlers.go
samlaf: [a36001e](https://github.com/Layr-Labs/eigenda-proxy/pull/185/commits/a36001e07f71ea74b1323339cb8e84a34f7fcd22)

server/middleware.go
samlaf: [b91ab13](https://github.com/Layr-Labs/eigenda-proxy/pull/185/commits/b91ab13c275626a51e78649c5f07a253dd2954fe)
Changed them both to just say "unknown" instead, but prefer not to introduce a type for this simple case. If it was used elsewhere I'd be fine with it, but here I think its clearer to just have the string written, instead of seeing a variable and then needing to click-follow in the IDE to see what string is going to appear in the actual prometheus metric.

server/routing.go
samlaf: Oh I've got many mouthfuls to say about this haha:
1. re ethereum projects not using web frameworks: not quite sure I haven't done an extensive survey. A lot of them are simple json-rpc, so they use a json-rpc library instead of a rest api framework, but the altda decided to go with rest api. Also if I had to guess I'd say if other projects that are (proxy) servers like ours (like flashbots builder say) are not using web framework either out of ignorance, or because their routing is super simple. But our routing is a bit more involved, so basic stdlib http router (even after go1.22 improvements) isn't enough.
2. gorilla is not a heavy framework. it's actually extremely lightweight. If I had my way I would actually use a full fledged framework with middlewares, like gin or echo (note that even these are "lightweight" frameworks). But gorilla mux is purely a routing library, not a framework. See limitations section at the bottom of [this](https://eli.thegreenplace.net/2021/rest-servers-in-go-part-3-using-a-web-framework/#Limitations) for a comparison.
3. I already tried to explain the benefits of using gorilla's router instead of stdlib's router. I think this [article](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html) summarizes my thoughts regarding the problem with our current implementation better than I ever could. Basically our routing ifs/switches are sprinkled everywhere in the proxy. This makes for very difficult reading. There's really only 3 types of commitments, so we should make the choice one of which "row" we're in, at the top routing layer, and then the rest of the code should be aware of which route it's in and be able to specialize. See for eg this [switch statement](https://github.com/Layr-Labs/eigenda-proxy/blob/refactor--better-routing-using-gorilla-mux-patterns/store/router.go#L123); don't think we should have this that far down code paths.

server/handlers.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+// =================================================================================================
+// POST ROUTES
+// =================================================================================================
+
+// handlePostSimpleCommitment handles the POST request for simple commitments.
+func (svr *Server) handlePostSimpleCommitment(w http.ResponseWriter, r *http.Request) error {
+	svr.log.Info("Processing simple commitment")
```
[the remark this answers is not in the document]
samlaf: [465fb84](https://github.com/Layr-Labs/eigenda-proxy/pull/185/commits/465fb84201021c3c2ee70e1b1cc05a980260bc00)
moved all the logs to the shared handlers (adds a redundant reencoding to hex, but at least its simpler this way)

server/handlers.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("failed to decode commitment %s: %w", rawCommitmentHex, err)
+	}
+
+	svr.log.Info("Processing op keccak commitment POST", "commitment", rawCommitmentHex, "commitmentMeta", commitmentMeta)
+	return svr.handlePostShared(w, r, commitment, commitmentMeta)
+}
+
+// handlePostOPGenericCommitment handles the POST request for optimism generic commitments.
+func (svr *Server) handlePostOPGenericCommitment(w http.ResponseWriter, r *http.Request) error {
+	svr.log.Info("Processing simple commitment")
```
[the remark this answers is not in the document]
samlaf: [465fb84](https://github.com/Layr-Labs/eigenda-proxy/pull/185/commits/465fb84201021c3c2ee70e1b1cc05a980260bc00)
moved all the logs to the shared handlers (adds a redundant reencoding to hex, but at least its simpler this way)

epociask (approved):
LGTM

PR #190, opened 2024-10-24
fix: fetch quorum parameters from edasm at RBN instead of latest block

samlaf:
## Fixes Issue

Think this is a bug... but not 100% sure. So please verify that this is correct.

Fixes #

### DIFF
--- a/verify/verifier.go
+++ b/verify/verifier.go
@@ -2,10 +2,12 @@ package verify
 
 import (
 	"fmt"
+	"math/big"
 
 	"github.com/consensys/gnark-crypto/ecc"
 	"github.com/consensys/gnark-crypto/ecc/bn254"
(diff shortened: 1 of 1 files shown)


epociask (approved):
LGTM - this is more future proofing then a current bug fix since these thresholds are hardcoded into the `EigenDAServiceManagerStorage` [contract](https://github.com/Layr-Labs/eigenda/blob/master/contracts/src/core/EigenDAServiceManagerStorage.sol)

samlaf:
> LGTM - this is more future proofing then a current bug fix since these thresholds are hardcoded into the `EigenDAServiceManagerStorage` [contract](https://github.com/Layr-Labs/eigenda/blob/master/contracts/src/core/EigenDAServiceManagerStorage.sol)

Ah gotcha then I need to update the comment I wrote.

PR #192, opened 2024-10-25
refactor: use batch confirmation that was upstreamed to eigenda client

samlaf:
This PR depends on https://github.com/Layr-Labs/eigenda/pull/821 being merged first.

Main simplification that this allows is getting rid of the very complex retry logic that we had where an inner functionw as returning an error that was being caught by a very outward for loop, which was hard to understand. This PR makes that logic more local, but I opted to still keep the check in eigenda-proxy as well as an assert/safety guarantee. Think we also need it in the Get route... but not super familiar with that code path cc @epociask to make sure logic in this PR still works for that.

## Fixes Issue

Fixes #

### DIFF
--- a/flags/eigendaflags/cli.go
+++ b/flags/eigendaflags/cli.go
@@ -1,6 +1,8 @@
 package eigendaflags
 
 import (
+	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/api/clients"
@@ -21,6 +23,9 @@ var (
 	PutBlobEncodingVersionFlagName       = withFlagPrefix("put-blob-encoding-version")
 	DisablePointVerificationModeFlagName = withFlagPrefix("disable-point-verification-mode")
 	WaitForFinalizationFlagName          = withFlagPrefix("wait-for-finalization")
+	ConfirmationDepthFlagName            = withFlagPrefix("confirmation-depth")
+	EthRPCURLFlagName                    = withFlagPrefix("eth-rpc")
+	SvcManagerAddrFlagName               = withFlagPrefix("svc-manager-addr")
 )
 
 func withFlagPrefix(s string) string {
@@ -101,11 +106,41 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			EnvVars:  []string{withEnvPrefix(envPrefix, "WAIT_FOR_FINALIZATION")},
 			Value:    false,
 			Category: category,
+			Hidden:   true,
+			Action: func(_ *cli.Context, _ bool) error {
+				return fmt.Errorf("flag --%s is deprecated, instead use --%s finalized", WaitForFinalizationFlagName, ConfirmationDepthFlagName)
+			},
+		},
+		&cli.StringFlag{
+			Name: ConfirmationDepthFlagName,
+			Usage: "Number of Ethereum blocks to wait after the blob's batch has been included on-chain, " +
+				"before returning from PutBlob calls. Can either be a number or 'finalized'.",
+			EnvVars:  []string{withEnvPrefix(envPrefix, "CONFIRMATION_DEPTH")},
+			Value:    "0",
+			Category: category,
+			Action: func(ctx *cli.Context, val string) error {
+				return validateConfirmationFlag(val)
+			},
+		},
+		&cli.StringFlag{
+			Name:     EthRPCURLFlagName,
+			Usage:    "URL of the Ethereum RPC endpoint. Needed to confirm blobs landed onchain.",
+			EnvVars:  []string{withEnvPrefix(envPrefix, "ETH_RPC")},
+			Category: category,
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     SvcManagerAddrFlagName,
+			Usage:    "Address of the EigenDAServiceManager contract. Required to confirm blobs landed onchain. See https://github.com/Layr-Labs/eigenlayer-middleware/?tab=readme-ov-file#current-mainnet-deployment",
+			EnvVars:  []string{withEnvPrefix(envPrefix, "SERVICE_MANAGER_ADDR")},
+			Category: category,
+			Required: true,
 		},
 	}
 }
 
 func ReadConfig(ctx *cli.Context) clients.EigenDAClientConfig {
+	waitForFinalization, confirmationDepth := parseConfirmationFlag(ctx.String(ConfirmationDepthFlagName))
 	return clients.EigenDAClientConfig{
 		RPC:                          ctx.String(DisperserRPCFlagName),
 		StatusQueryRetryInterval:     ctx.Duration(StatusQueryRetryIntervalFlagName),
@@ -116,6 +151,40 @@ func ReadConfig(ctx *cli.Context) clients.EigenDAClientConfig {
 		SignerPrivateKeyHex:          ctx.String(SignerPrivateKeyHexFlagName),
 		PutBlobEncodingVersion:       codecs.BlobEncodingVersion(ctx.Uint(PutBlobEncodingVersionFlagName)),
 		DisablePointVerificationMode: ctx.Bool(DisablePointVerificationModeFlagName),
-		WaitForFinalization:          ctx.Bool(WaitForFinalizationFlagName),
+		WaitForFinalization:          waitForFinalization,
+		WaitForConfirmationDepth:     confirmationDepth,
+		EthRpcUrl:                    ctx.String(EthRPCURLFlagName),
+		SvcManagerAddr:               ctx.String(SvcManagerAddrFlagName),
 	}
+}
+
+// Helper to parse the flag value into config fields
+func parseConfirmationFlag(val string) (waitForFinalization bool, confirmationDepth uint64) {
+	if val == "finalized" {
+		return true, 0
+	}
+
+	depth, err := strconv.ParseUint(val, 10, 64)
+	if err != nil {
+		panic("this should never happen, as the flag is validated before this point")
+	}
+
+	return false, depth
+}
+
+func validateConfirmationFlag(val string) error {
+	if val == "finalized" {
+		return nil
+	}
+
+	depth, err := strconv.ParseUint(val, 10, 64)
+	if err != nil {
+		return fmt.Errorf("confirmation-depth must be either 'finalized' or a number, got: %s", val)
+	}
+
+	if depth >= 64 {
+		fmt.Printf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Consider using 'finalized' instead.\n", depth)
+	}
+
+	return nil
 }
--- a/flags/eigendaflags/cli.go
+++ b/flags/eigendaflags/cli.go
@@ -2,6 +2,7 @@ package eigendaflags
 
 import (
 	"fmt"
+	"log"
 	"strconv"
 	"time"
 
@@ -118,7 +119,7 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			EnvVars:  []string{withEnvPrefix(envPrefix, "CONFIRMATION_DEPTH")},
 			Value:    "0",
 			Category: category,
-			Action: func(ctx *cli.Context, val string) error {
+			Action: func(_ *cli.Context, val string) error {
 				return validateConfirmationFlag(val)
 			},
 		},
@@ -158,8 +159,8 @@ func ReadConfig(ctx *cli.Context) clients.EigenDAClientConfig {
 	}
 }
 
-// Helper to parse the flag value into config fields
-func parseConfirmationFlag(val string) (waitForFinalization bool, confirmationDepth uint64) {
+// parse the val (either "finalized" or a number) into waitForFinalization (bool) and confirmationDepth (uint64).
+func parseConfirmationFlag(val string) (bool, uint64) {
 	if val == "finalized" {
 		return true, 0
 	}
@@ -183,7 +184,7 @@ func validateConfirmationFlag(val string) error {
 	}
 
 	if depth >= 64 {
-		fmt.Printf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Consider using 'finalized' instead.\n", depth)
+		log.Printf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Consider using 'finalized' instead.\n", depth)
 	}
 
 	return nil
--- a/go.mod
+++ b/go.mod
@@ -5,10 +5,7 @@ go 1.22
 toolchain go1.22.0
 
 require (
-	// pointing to latest commit of https://github.com/Layr-Labs/eigenda/pull/821
-	// TODO: merge eigenda-client PR first, and then point to the latest master commit
-	//       instead of PR commit (which will get squash and merged)
-	github.com/Layr-Labs/eigenda v0.8.5-0.20241025110857-96ad889d3c17
+	github.com/Layr-Labs/eigenda v0.8.5-0.20241028201743-5fe3e910a22d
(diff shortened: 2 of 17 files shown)


go.mod
```diff
@@ -293,3 +293,7 @@ require (
 
 // We inherit this from OP by forking their Alt-DA server.
 replace github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101408.0-rc.4.0.20240827042333-110c433a2469
+
+// Adding this replace directive to test against local changes from https://github.com/Layr-Labs/eigenda/pull/828
+// TODO: remove this replace directive once that PR has stabilized
+replace github.com/Layr-Labs/eigenda => ../../eigenda
```
samlaf, line 299: TODO

go.mod
```diff
@@ -5,7 +5,10 @@ go 1.22
 toolchain go1.22.0
 
 require (
-	github.com/Layr-Labs/eigenda v0.8.4
+	// pointing to latest commit of https://github.com/Layr-Labs/eigenda/pull/821
+	// TODO: merge eigenda-client PR first, and then point to the latest master commit
+	//       instead of PR commit (which will get squash and merged)
+	github.com/Layr-Labs/eigenda v0.8.5-0.20241025110857-96ad889d3c17
```
samlaf, line 11: TODO: merge https://github.com/Layr-Labs/eigenda/pull/821 before this PR

samlaf: Also notttt....... sure why the ci workflows are all broken. Can't find batchHeader... at lines that don't even exist. lint test and build work fine locally for me so I'm confused. Wondering if it doesn't have to do with this reference somehow..

verify/cert.go
```diff
@@ -49,40 +51,60 @@ func NewCertVerifier(cfg *Config, l log.Logger) (*CertVerifier, error) {
...
+	// 1. verify batch is actually onchain at the batchMetadata's state confirmedBlockNumber
+	// This assert is technically not necessary, but it's a good sanity check.
+	// It could technically happen that a confirmed batch's block gets reorged out,
+	// yet the tx is included in an earlier or later block, making this check fail...
+	confirmationBlockNumber := batchMetadata.GetConfirmationBlockNumber()
+	confirmationBlockNumberBigInt := big.NewInt(0).SetInt64(int64(confirmationBlockNumber))
+	_, err := cv.retrieveBatchMetadataHash(ctx, batchID, confirmationBlockNumberBigInt)
 	if err != nil {
-		return fmt.Errorf("failed to get context block: %w", err)
+		return fmt.Errorf("batch not found onchain at supposedly confirmed block %d: %w", confirmationBlockNumber, err)
```
samlaf, line 67: Is this true? If so I think we should get rid of this check but document exactly why we are getting rid of it. Would have saved me a lot of time if this had been documented in the initial code.

bxue-l2 (dismissed):

verify/cert.go
```diff
@@ -49,40 +51,60 @@ func NewCertVerifier(cfg *Config, l log.Logger) (*CertVerifier, error) {
 	}, nil
 }
 
-// verifies on-chain batch ID for equivalence to certificate batch header fields
-func (cv *CertVerifier) VerifyBatch(
-	header *binding.IEigenDAServiceManagerBatchHeader, id uint32, recordHash [32]byte, confirmationNumber uint32,
+// verifyBatchConfirmedOnchain verifies that batchMetadata (typically part of a received cert)
+// matches the batch metadata hash stored on-chain
+func (cv *CertVerifier) verifyBatchConfirmedOnchain(
```
epociask, line 56: Knit
```suggestion
func (cv *CertVerifier) verifyBatchConfirmedOnChain(
```

verify/cert.go
```diff
@@ -49,40 +51,60 @@ func NewCertVerifier(cfg *Config, l log.Logger) (*CertVerifier, error) {
...
 	if err != nil {
-		return fmt.Errorf("failed to get context block: %w", err)
+		return fmt.Errorf("batch not found onchain at supposedly confirmed block %d: %w", confirmationBlockNumber, err)
 	}
 
-	// 1. ensure that a batch hash can be looked up for a batch ID for a given block number
-	expectedHash, err := cv.manager.BatchIdToBatchMetadataHash(&bind.CallOpts{BlockNumber: blockNumber}, id)
+	// 2. verify that the confirmation status has been reached
+	// We retry 5 times waiting 12 seconds (block time) between each retry,
+	// in case our eth node is behind that of the eigenda_client's node that deemed the batch confirmed
```
epociask, line 72: why would we ever use two nodes? or is the idea to have some resiliency to transient failures where the provider could be unreliable due to poor configuration?

verify/cert.go
```diff
@@ -49,40 +51,60 @@ func NewCertVerifier(cfg *Config, l log.Logger) (*CertVerifier, error) {
...
+	}
+	recordHash := [32]byte(batchMetadata.GetSignatoryRecordHash())
+	computedHash, err := HashBatchMetadata(header, recordHash, confirmationBlockNumber)
 	if err != nil {
 		return fmt.Errorf("failed to hash batch metadata: %w", err)
 	}
 
-	equal := slices.Equal(expectedHash[:], actualHash[:])
+	// 4. ensure that hash generated from local cert matches one stored on-chain
+	equal := slices.Equal(onchainHash[:], computedHash[:])
```
epociask, line 98: knit - there is an equality func in utils: https://github.com/Layr-Labs/eigenda-proxy/blob/main/utils/utils.go#L31-L43

verify/cert.go
```diff
@@ -103,10 +125,35 @@ func (cv *CertVerifier) VerifyMerkleProof(inclusionProof []byte, root []byte,
...
+	}
+	return new(big.Int).SetUint64(blockNumber - cv.ethConfirmationDepth), nil
+}
+
+// retrieveBatchMetadataHash retrieves the batch metadata hash stored on-chain at a specific blockNumber for a given batchID
+// returns an error if some problem calling the contract happens, or the hash is not found
+func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID uint32, blockNumber *big.Int) ([32]byte, error) {
+	onchainHash, err := cv.manager.BatchIdToBatchMetadataHash(&bind.CallOpts{Context: ctx, BlockNumber: blockNumber}, batchID)
+	if err != nil {
+		return [32]byte{}, fmt.Errorf("calling EigenDAServiceManager.BatchIdToBatchMetadataHash: %w", err)
```
epociask, line 153: knit
```suggestion
		return [32]byte{}, fmt.Errorf("error calling EigenDAServiceManager.BatchIdToBatchMetadataHash: %w", err)
```

verify/cert.go
```diff
@@ -103,10 +125,35 @@ func (cv *CertVerifier) VerifyMerkleProof(inclusionProof []byte, root []byte,
 }
 
 // fetches a block number provided a subtraction of a user defined conf depth from latest block
-func (cv *CertVerifier) getConfDeepBlockNumber() (*big.Int, error) {
-	blockNumber, err := cv.ethClient.BlockNumber(context.Background())
+func (cv *CertVerifier) getConfDeepBlockNumber(ctx context.Context) (*big.Int, error) {
+	if cv.waitForFinalization {
+		// https://github.com/ethereum/execution-apis/blob/4140e528360fea53c34a766d86a000c6c039100e/src/eth/block.yaml#L61
+		var header = types.Header{}
+		err := cv.ethClient.Client().CallContext(ctx, &header, "eth_getBlockByNumber", "finalized", false)
```
epociask, line 132: assuming this is encoded to the raw params like [`finalized, `false`] where the boolean indicates that we aren't collecting txs for the block? We may wanna provide more explicit detailing of this value for easier readability long-term

verify/cert.go
samlaf: Interesting. But I'm not even sure, because the case I'm describing is this:
0. (assumption: we wait for 5 block confirmation)
1. batcher confirms batch onchain at block 1000
2. eigenda-client wait for 5 blocks and returns a BlobStatus
3. proxy receives this one block later at block 1006, and the batch has actually been reorged and included in block 1001 instead.
4. In this case the check highlighted here would fail, even though the block is actually confirmed with depth 5 (the next assert below around line 70 would catch if that was not the case)

For this reason I think this check is actually wrong here, and I still think we should remove this check. Does this make sense?

verify/cert.go
samlaf: [b1550e3](https://github.com/Layr-Labs/eigenda-proxy/pull/192/commits/b1550e38518627449974c2cd38ed6c389389bc33)

verify/cert.go
samlaf: Both eigenda-client and proxy dial the endpoint individually, establishing independent connections. If the nodes are behind a load-balancer, then both connections could actually be connected to different eth-nodes, its very frequent that rpc provider nodes are not in sync (bit me very hard once) for some dumb reason (web3 standards...).

Even if we injected the dialed connection from the proxy into the eigenda-client, the underlying tcp connection could drop in between the eigenda-client returning and proxy making its call, which would reconnect to another node, with some effect.

verify/cert.go
samlaf: Didn't know! But checked and that implementation is literally the same as [slices.Equal](https://pkg.go.dev/golang.org/x/exp/slices#Equal) though. I'd prefer sticking with stdlib (even if in exp package) stuff when we can.

verify/cert.go
samlaf: Disagree here. Read https://github.com/uber-go/guide/blob/master/style.md#error-wrapping last week which makes a lot of sense, been trying to use this convention as much as possible.

verify/cert.go
samlaf: [7069e13](https://github.com/Layr-Labs/eigenda-proxy/pull/192/commits/7069e13e252623e6e9b163e7e8b628e7c662e2bb)

flags/eigendaflags/cli.go
```diff
@@ -21,6 +21,9 @@ var (
 	PutBlobEncodingVersionFlagName       = withFlagPrefix("put-blob-encoding-version")
 	DisablePointVerificationModeFlagName = withFlagPrefix("disable-point-verification-mode")
 	WaitForFinalizationFlagName          = withFlagPrefix("wait-for-finalization")
+	WaitForConfirmationDepthFlagName     = withFlagPrefix("wait-for-confirmation-depth")
```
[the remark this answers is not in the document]
samlaf: Thought about this but not sure I like it. Perhaps we could do something like what the json-rpc does (where you can either pass a block number, or a string like "finalized" or "latest")?
```
cli.StringFlag{
    Name:  "confirmation-mode",
    Usage: "Confirmation mode: <number of blocks> or 'finalized'",
    Value: "finalized",
}
```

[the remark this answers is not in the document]
samlaf: Made the change: [fbcf650](https://github.com/Layr-Labs/eigenda-proxy/pull/192/commits/fbcf650275252985cc1bd4c403cea59fbc9cf897)

samlaf:
@epociask I'll let you review my answers before rebasing and fixing conflicts, so that the commit links still work.

flags/eigendaflags/cli.go
```diff
@@ -116,6 +151,41 @@ func ReadConfig(ctx *cli.Context) clients.EigenDAClientConfig {
...
+		WaitForFinalization:          waitForFinalization,
+		WaitForConfirmationDepth:     confirmationDepth,
+		EthRpcUrl:                    ctx.String(EthRPCURLFlagName),
+		SvcManagerAddr:               ctx.String(SvcManagerAddrFlagName),
+	}
+}
+
+// Helper to parse the flag value into config fields
+func parseConfirmationFlag(val string) (waitForFinalization bool, confirmationDepth uint64) {
+	fmt.Println("val: ", val)
```
epociask, line 163: should probably rm this line

flags/eigendaflags/cli.go
```diff
@@ -116,6 +151,41 @@ func ReadConfig(ctx *cli.Context) clients.EigenDAClientConfig {
...
+	if err != nil {
+		return fmt.Errorf("confirmation-depth must be either 'finalized' or a number, got: %s", val)
+	}
+
+	if depth >= 64 {
+		fmt.Printf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Consider using 'finalized' instead.\n", depth)
+	}
+
+	return nil
 }
```
epociask, line 191: Is there any potential value using geth's json rpc unmarshaling directly to get cross compatibility with other commitment levels? (e.g, `latest`, `safe`)  without having to redefine types?

https://github.com/ethereum/go-ethereum/blob/80bdab757dfb0f6d73fb869d834979536fe474e5/rpc/types.go#L73-L112

verify/cert.go
epociask: makes sense and was thinking that too - got burnt in the past using a node cluster with an ALB ingress where subsequent RPC queries could fail due to nodes being out of sync. Ik OP clusters typically use [proxyd](https://github.com/ethereum-optimism/infra/tree/main/proxyd) to mitigate this from happening. from what I remember the connection would be held with the ALB directly with different nodes being queried per each new request on the client --> ALB connection. Feel free to resolve this comment!

verify/cert.go
epociask: understood - lets cut an issue to remove that utils function entirely then

verify/cert.go
epociask: Actually on second thought - isn't this a bit naive when we have access to the `L1ReferenceBlock` number at which the batch should've confirmed? Also we're adding a 12 second cool-off which is aligned with the block production interval on Ethereum but not syncing

samlaf: proxyd doesn't seem to solve this problem, they say: "we consider healthy is... latest block lag ≤ configurable threshold". Way to solve this is to use sticky sessions.

Not sure I understand your last comment. Is https://github.com/Layr-Labs/eigenda-proxy/pull/192#discussion_r1818085482 related to what you are talking about?

flags/eigendaflags/cli.go
samlaf: yep thanks, removed

flags/eigendaflags/cli.go
samlaf: Maybe, but they take a byte as input, and accept many more strings. For now let's keep it like this.

verify/cert.go
bxue-l2: EigenDA batcher detects and updates confirmation number, if it detects reorg, https://github.com/Layr-Labs/eigenda/blob/master/disperser/batcher/finalizer.go#L198

verify/cert.go
samlaf: Talked offline. Let's update to retrying every 2 seconds for the same amount of time (60 seconds).

go.mod
samlaf: merged + updated: [6a25446](https://github.com/Layr-Labs/eigenda-proxy/pull/192/commits/6a25446508b5e9e4a89f2fabd0027d1b7f739e4b)

verify/cert.go
samlaf: [55edfae](https://github.com/Layr-Labs/eigenda-proxy/pull/192/commits/55edfaefb4297be0dbd31e6251db691b951560a4)

verify/cert.go
samlaf: [0c54503](https://github.com/Layr-Labs/eigenda-proxy/pull/192/commits/0c545036bcf8f27cb9dac9882aa8649fd508a659)

epociask (approved):
LGTM

PR #194, opened 2024-10-26
fix: Hide other sensitive cfg values

### LINKED ISSUE #176: Cache credentials and the private key are exposed in the logs
When the proxy starts, it prints the proxy server config as a log message at info level. If S3 or Redis credentials are used, they are leaked to the logs from that server config. Additionally `SignerPrivateKeyHex` is always leaked regardless of the cache or fallback being used.

```
│ t=2024-10-08T17:28:35+0000 lvl=info msg="Initializing EigenDA proxy server with config: {\n  \"EigenDAConfig\": {\n    \"EdaClientConfig\": {\n      \"RPC\": \"disperser-holesky.eigen │
│ da.xyz:443\",\n      \"StatusQueryTimeout\": 1800000000000,\n      \"StatusQueryRetryInterval\": 5000000000,\n      \"ResponseTimeout\": 60000000000,\n      \"CustomQuorumIDs\": [],\n │
│       \"SignerPrivateKeyHex\": \"REDACTED",\n      \"DisableTLS\": false,\n      \"PutBlobEncodingVersion\": 0,\n      \"Disab │
│ lePointVerificationMode\": false,\n      \"WaitForFinalization\": false\n    },\n    \"VerifierConfig\": {\n      \"KzgConfig\": {\n        \"G1Path\": \"resources/g1.point\",\n       │
│   \"G2Path\": \"\",\n        \"G1PowerOf2Path\": \"\",\n        \"G2PowerOf2Path\": \"resources/g2.point.powerOf2\",\n        \"CacheDir\": \"resources/SRSTables/\",\n        \"NumWor │
│ ker\": 8,\n        \"SRSOrder\": 268435456,\n        \"SRSNumberToLoad\": 524288,\n        \"Verbose\": false,\n        \"PreloadEncoder\": false\n      },\n      \"VerifyCerts\": fal │
│ se,\n      \"RPCURL\": \"\",\n      \"SvcManagerAddr\": \"\",\n      \"EthConfirmationDepth\": 0\n    },\n    \"MemstoreEnabled\": false,\n    \"MemstoreConfig\": {\n      \"MaxBlobSi │
│ zeBytes\": 16777216,\n      \"BlobExpiration\": 1500000000000,\n      \"PutLatency\": 0,\n      \"GetLatency\": 0\n    },\n    \"FallbackTargets\": [],\n    \"CacheTargets\": [],\n    │
│  \"RedisConfig\": {\n      \"Endpoint\": \"\",\n      \"Password\": \"\",\n      \"DB\": 0,\n      \"Eviction\": 86400000000000,\n      \"Profile\": false\n    },\n    \"S3Config\": { │
│ \n      \"CredentialType\": \"static\",\n      \"Endpoint\": \"storage.googleapis.com\",\n      \"EnableTLS\": false,\n      \"AccessKeyID\": \"REDACTED\",\n      \"AccessKeySecret\": \"REDACTED\",\n      \"Bucket\": \"REDACTED\",\n      \"Path\": \"\",\n      │
│  \"Backup\": false,\n      \"Timeout\": 5000000000,\n      \"Profiling\": false\n    }\n  },\n  \"MetricsCfg\": {\n    \"Enabled\": true,\n    \"ListenAddr\": \"0.0.0.0\",\n    \"List │
│ enPort\": 7300\n  }\n}" role=eigenda_proxy
```

epociask:
## Fixes Issue

Fixes #176

### DIFF
--- a/cmd/server/entrypoint.go
+++ b/cmd/server/entrypoint.go
@@ -76,13 +76,17 @@ func StartProxySvr(cliCtx *cli.Context) error {
 func prettyPrintConfig(cliCtx *cli.Context, log log.Logger) error {
 	// we read a new config which we modify to hide private info in order to log the rest
 	cfg := server.ReadCLIConfig(cliCtx)
-	cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "*****" // marshaling defined in client config
-	cfg.EigenDAConfig.EdaClientConfig.RPC = "*****"                 // hiding as RPC providers typically use sensitive API keys within
+	if cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex != "" {
+		cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "*****" // marshaling defined in client config
+	}
+	if cfg.EigenDAConfig.EdaClientConfig.EthRpcUrl != "" {
+		cfg.EigenDAConfig.EdaClientConfig.EthRpcUrl = "*****" // hiding as RPC providers typically use sensitive API keys within
+	}
 
 	configJSON, err := json.MarshalIndent(cfg, "", "  ")
 	if err != nil {
 		return fmt.Errorf("failed to marshal config: %w", err)
 	}
-	log.Info(fmt.Sprintf("Initializing EigenDA proxy server with config: %v", string(configJSON)))
+	log.Info(fmt.Sprintf("Initializing EigenDA proxy server with config (\"*****\" fields are hidden): %v", string(configJSON)))
 	return nil
 }
--- a/cmd/server/entrypoint.go
+++ b/cmd/server/entrypoint.go
@@ -76,12 +76,13 @@ func StartProxySvr(cliCtx *cli.Context) error {
 func prettyPrintConfig(cliCtx *cli.Context, log log.Logger) error {
 	// we read a new config which we modify to hide private info in order to log the rest
 	cfg := server.ReadCLIConfig(cliCtx)
+	cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "*****" // marshaling defined in client config
+	cfg.EigenDAConfig.EdaClientConfig.RPC = "*****"                 // hiding as RPC providers typically use sensitive API keys within
+
 	configJSON, err := json.MarshalIndent(cfg, "", "  ")
 	if err != nil {
 		return fmt.Errorf("failed to marshal config: %w", err)
 	}
-	cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "" // marshaling defined in client config
-	cfg.EigenDAConfig.EdaClientConfig.RPC = ""                 // hiding as RPC providers typically use sensitive API keys within
 	log.Info(fmt.Sprintf("Initializing EigenDA proxy server with config: %v", string(configJSON)))
 	return nil
 }
--- a/cmd/server/entrypoint.go
+++ b/cmd/server/entrypoint.go
@@ -73,17 +73,15 @@ func StartProxySvr(cliCtx *cli.Context) error {
 }
 
 // TODO: we should probably just change EdaClientConfig struct definition in eigenda-client
-// to have a `json:"-"` tag on the SignerPrivateKeyHex field, to prevent the privateKey from being marshaled at all
 func prettyPrintConfig(cliCtx *cli.Context, log log.Logger) error {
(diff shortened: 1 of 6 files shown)


cmd/server/entrypoint.go
```diff
@@ -73,17 +73,15 @@ func StartProxySvr(cliCtx *cli.Context) error {
...
 	// we read a new config which we modify to hide private info in order to log the rest
 	cfg := server.ReadCLIConfig(cliCtx)
-	cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "HIDDEN"
-	cfg.EigenDAConfig.VerifierConfig.RPCURL = "HIDDEN"
-
 	configJSON, err := json.MarshalIndent(cfg, "", "  ")
 	if err != nil {
 		return fmt.Errorf("failed to marshal config: %w", err)
 	}
+	cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "" // marshaling defined in client config
```
samlaf, line 83: Why did you move this after the marshaling? Like this it doesn’t do anything

store/precomputed_key/redis/redis.go
```diff
@@ -12,11 +12,10 @@ import (
 
 // Config ... user configurable
 type Config struct {
-	Endpoint string
-	Password string
-	DB       int
-	Eviction time.Duration
-	Profile  bool
+	Endpoint string        `json:"endpoint"`
+	Password string        `json:"-"`
```
samlaf, line 16: Nice! Ty!

samlaf (commented):
Thanks for this! Just one thing to fix

cmd/server/entrypoint.go
epociask: it does actually - the `SignerPrivateKey` is stored as part of the `EigenDAClientConfig` which is defined within layr-labs/eigenda

samlaf: The log line below is
```
log.Info(fmt.Sprintf("Initializing EigenDA proxy server with config: %v", string(configJSON)))
```
which prints the already marshalled config. So changing the fields of cfg here above doesn't do anything.

epociask: 🤦‍♂️

store/precomputed_key/redis/redis.go
samlaf, line 16: Thought about this approach some more and not sure I like it, for a few reasons:
1) if some future developer ever wants to marshal AND unmarshal the config for whatever reason he will run into a bug (can fix this by defining a custom marshalJSONHideSecrets function instead)
2) this will print empty string, so reader won’t know if password field was hidden or just never set (we can fix this by using omitempty json tag)

Could look something like this (Claude generated so beware):
```
package main

import (
	"encoding/json"
	"fmt"
	"time"
)

type Config struct {
	Endpoint string        `json:"endpoint"`
	Password string        `json:"password,omitempty"`
	DB       int           `json:"database"`
	Eviction time.Duration `json:"eviction"`
}

// Custom MarshalJSON function to control what gets included in the JSON output
func (c Config) MarshalJSON() ([]byte, error) {
	type Alias Config // Use an alias to avoid recursion with MarshalJSON
	aux := struct {
		Alias
		Password string `json:"password,omitempty"`
	}{
		Alias:    (Alias)(c),
		Password: "",
	}

	// Conditionally include a masked password if it is set
	if c.Password != "" {
		aux.Password = "****" // Mask the password instead of excluding it
	}
	return json.Marshal(aux)
}

func main() {
	cfg := Config{
		Endpoint: "localhost:6379",
		Password: "supersecret",
		DB:       0,
		Eviction: time.Hour,
	}

	jsonData, _ := json.MarshalIndent(cfg, "", "  ")
	fmt.Println(string(jsonData))
}
```
Maybe also use a separate Marshal function to not change the default marshaling behavior?

cmd/server/entrypoint.go
samlaf: [53421d7](https://github.com/Layr-Labs/eigenda-proxy/pull/194/commits/53421d71c3e4c96761395078089c3e3c0411a401)

store/precomputed_key/redis/redis.go
samlaf: [4a1404b](https://github.com/Layr-Labs/eigenda-proxy/pull/194/commits/4a1404b42979775ef2b90c09cee1ba2a674a154d)

samlaf:
@epociask @bxue-l2 can you review my fixes. Want to get this merged ASAP, especially before 1.6.0 release.

bxue-l2 (approved):

PR #200, opened 2024-11-18
fix(middleware): proper handling of handler status codes in log/metrics middlewares

### LINKED ISSUE #186: middlewares: implement ResponseWriter wrapper to log status code
Currently we cannot log the status code because go's default ResponseWriter interface does not expose it.

We also use `w.Header().Get("status")` in the [metrics middleware](https://github.com/Layr-Labs/eigenda-proxy/blob/df3a6f8ee1d4d6dd7389c7e78cb034a7516e05e2/server/server.go#L70), which doesn't work because the status code is not actually a header.

### Action Item:
- implement a ResponseWriter wrapper that saves the status code: see https://github.com/golang/go/issues/18997
- use it to log the status code in the logging middleware
- fix the metrics middleware to properly read the status code

samlaf:
## Fixes Issue

Fixes #186


server/middleware.go
```diff
@@ -21,24 +45,27 @@ func withMetrics(
...
+			recordDur(strconv.Itoa(scw.status), commitMode, certVersion)
 			return err
 		}
-		// TODO: some routes don't have a version byte, such as /put simple commitments
 		versionByte, err := parseVersionByte(w, r)
 		if err != nil {
-			// we assume that every route will set the status header
-			recordDur(w.Header().Get("status"), string(mode), string(versionByte))
-			return fmt.Errorf("metrics middleware: error parsing version byte: %w", err)
+			// Some routes don't have a version byte, such as /put simple commitments.
```
epociask, line 63: why would /put simple commitments not have a version byte but OP /put would?

samlaf: Good catch.... I was totally wrong here. I updated to: [096003c](https://github.com/Layr-Labs/eigenda-proxy/pull/200/commits/096003c396b2fa32840616452ced021d00109d75)
Realized our version bytes are hardcoded for all the PUT routes.. bit ugly how I did some of the routing now I'm realizing. But let me know what you think of these changes. You'll have to change this when you get the v2 changes in.

samlaf: Oh also for reference it seems like its the keccak commitments that don't have a version byte: https://github.com/Layr-Labs/eigenda-proxy/blob/096003c396b2fa32840616452ced021d00109d75/server/routing.go#L66

epociask (approved):
LGTM

PR #201, opened 2024-11-18
refactor: rename client -> SimpleCommitmentClient and add comments

### LINKED ISSUE #184: Add comments explaining when our client is used instead of the rollup stack's client
We have a client in the repo: https://github.com/Layr-Labs/eigenda-proxy/blob/feat--openapi-spec/client/client.go

Seems it's only used for simple commitments, meaning nitro doesn't have a client the same way op does?

Could we document this in the code, as comments above the client struct? And link to the op client, explaining why we only have a client for the simple commitment mode.

samlaf:
## Fixes Issue

Fixes #184


epociask (approved):
LGTM - should we also update the file name (e.g, `client/simple.go`)?

samlaf:
> 964989d

Yep, done: [964989d](https://github.com/Layr-Labs/eigenda-proxy/pull/201/commits/964989d5c71f4247ba5a8d3f2f5def2494e8aa67)

PR #205, opened 2024-12-03
fix: Disable `SERVICE_MANAGER_ADDR` & `ETH_RPC` as mandatory

epociask:
## Changes proposed
Fixes a config processing bug where cert verification can be disabled but the application will crash since the service manager and eth rpc params are still required even-though they shouldn't be. 

Created https://github.com/Layr-Labs/eigenda-proxy/issues/206 to identify that there are better solutions.

### DIFF
--- a/flags/eigendaflags/cli.go
+++ b/flags/eigendaflags/cli.go
@@ -144,14 +144,14 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			Usage:    "URL of the Ethereum RPC endpoint. Needed to confirm blobs landed onchain.",
 			EnvVars:  []string{withEnvPrefix(envPrefix, "ETH_RPC")},
 			Category: category,
-			Required: true,
+			Required: false,
 		},
 		&cli.StringFlag{
 			Name:     SvcManagerAddrFlagName,
 			Usage:    "Address of the EigenDAServiceManager contract. Required to confirm blobs landed onchain. See https://github.com/Layr-Labs/eigenlayer-middleware/?tab=readme-ov-file#current-mainnet-deployment",
 			EnvVars:  []string{withEnvPrefix(envPrefix, "SERVICE_MANAGER_ADDR")},
 			Category: category,
-			Required: true,
+			Required: false,
 		},
 		// Flags that are proxy specific, and not used by the eigenda-client
 		// TODO: should we move this to a more specific category, like EIGENDA_STORE?
--- a/server/config.go
+++ b/server/config.go
@@ -45,6 +45,13 @@ func (cfg *Config) Check() error {
 		}
 	}
 
+	// provide dummy values to eigenda client config. Since the client won't be called in this
+	// mode it doesn't matter.
+	if cfg.MemstoreEnabled {
+		cfg.EdaClientConfig.SvcManagerAddr = "0x0000000000000000000000000000000000000000"
+		cfg.EdaClientConfig.EthRpcUrl = "http://0.0.0.0:666"
+	}
+
 	// cert verification is enabled
 	// TODO: move this verification logic to verify/cli.go
 	if cfg.VerifierConfig.VerifyCerts {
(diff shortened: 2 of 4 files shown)


flags/eigendaflags/cli.go
```diff
@@ -144,14 +144,14 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			Usage:    "URL of the Ethereum RPC endpoint. Needed to confirm blobs landed onchain.",
 			EnvVars:  []string{withEnvPrefix(envPrefix, "ETH_RPC")},
 			Category: category,
-			Required: true,
+			Required: false,
```
samlaf, line 147: needed here: https://github.com/Layr-Labs/eigenda/blob/a84eff96e545a168e46b66d82b40798188d78706/api/clients/eigenda_client.go#L383

samlaf:
How/where does the proxy crash? If we are using the latest version of eigenda_client (which I think we are?), those fields are actually mandatory, even if we don't do cert checking.

server/config.go
```diff
@@ -45,6 +45,13 @@ func (cfg *Config) Check() error {
 		}
 	}
 
+	// provide dummy values to eigenda client config. Since the client won't be called in this
+	// mode it doesn't matter.
+	if cfg.MemstoreEnabled {
+		cfg.EdaClientConfig.SvcManagerAddr = "0x0000000000000000000000000000000000000000"
+		cfg.EdaClientConfig.EthRpcUrl = "http://0.0.0.0:666"
+	}
```
samlaf, line 53: This works well for when memstore is enabled, but the issue now is that eth_rpc and svrmanager_addr are not mandatory anymore. So if someone starting proxy in prod forgets to initialize those, it will randomly crash at some point in the future when a request is made, which is not good.

Ideally we'd add eager defensive logic in eigenda_client initialization that errors in constructor if the svcmanager or ethrpc are uninitialized.

epociask: doesn't this check already capture this? https://github.com/Layr-Labs/eigenda-proxy/blob/main/server/config.go#L48-L60

samlaf: Ah! But I think it's wrong now because even with cert enabled we still mandatorily need eth_rpc and svc_manager_addr in eigenda_client now right?

epociask: isn't this the same check implicitly since they're both ingested via the same env source?

samlaf: Yes but my point is that the case: memstore off, cert verification off, still needs eth_rpc and svc_manager_addr to be populated. The check you shared above only runs when cert verification is on.

epociask: ahh gotcha - will update the conditional

epociask: [68a6979](https://github.com/Layr-Labs/eigenda-proxy/pull/205/commits/68a6979891597ebf4c470fade4e348aa4798f5e7)

server/config.go
```diff
@@ -47,7 +47,7 @@ func (cfg *Config) Check() error {
 
 	// provide dummy values to eigenda client config. Since the client won't be called in this
 	// mode it doesn't matter.
-	if cfg.MemstoreEnabled {
+	if cfg.MemstoreEnabled || !cfg.VerifierConfig.VerifyCerts {
```
samlaf, line 50: This won't be nice for users, because the eigenda_client constructor right now doesnt do sanitization of its input (it should be it doesnt).

Can you instead change to crashing with an error if svcManagerAddr and EthRpcUrl are not set when memstoreEnabled=false?

samlaf: https://github.com/Layr-Labs/eigenda-proxy/blob/main/server/config.go#L48-L60
Think this should be changed instead, to pull out the checks outside of the if statement, since eth_rpc and svc_manager_addr are no longer related to verifyCerts, they are always needed WHEN memstore is disabled.

epociask: [b3dd92b](https://github.com/Layr-Labs/eigenda-proxy/pull/205/commits/b3dd92b13ac252e7e5681234e481e5c7aa3fc3e3)

samlaf (approved):
Thanks :) :)

PR #208, opened 2024-12-04
feat: separate simple client into its own module

samlaf:
This will make it easier to import it into other projects without importing all of the other dependencies of our root go.mod

TODO steps:
1. merge this PR with a replace directive in the main go.mod
2. release v0.1.0 of the new client module
3. make a new PR to use that version in the main go.mod file
4. release v1.6.1 of main module that uses the new client module

## Fixes Issue

Fixes #


jianoaix (approved):

client/go.mod
jianoaix, line 1: also add go.sum?

epociask (approved):

client/go.mod
samlaf: The client only depends on stdlib so `go mod tidy` doesn't generate a go.sum in this case it seems like.

PR #209, opened 2024-12-04
chore: Add mockable http client to simple client

epociask:
## Fixes Issue

Fixes #

### DIFF
--- a/client/simple.go
+++ b/client/simple.go
@@ -14,7 +14,7 @@ var (
 )
 
 type Config struct {
-	URL string // EigenDA proxy RPC URL
+	URL string // EigenDA proxy REST API URL
 }
 
 type HTTPClient interface {
--- a/client/simple.go
+++ b/client/simple.go
@@ -99,7 +99,7 @@ func (c *SimpleCommitmentClient) GetData(ctx context.Context, comm []byte) ([]by
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("received error response when reading from eigenda, code=%d, msg = %s", resp.StatusCode, string(b))
+		return nil, fmt.Errorf("received error response when reading from eigenda-proxy, code=%d, msg = %s", resp.StatusCode, string(b))
 	}
 
 	return b, nil
@@ -131,7 +131,7 @@ func (c *SimpleCommitmentClient) SetData(ctx context.Context, b []byte) ([]byte,
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("received error response when dispersing to eigenda, code=%d, err = %s", resp.StatusCode, string(b))
+		return nil, fmt.Errorf("received error response when dispersing to eigenda-proxy, code=%d, err = %s", resp.StatusCode, string(b))
 	}
 
 	if len(b) == 0 {
--- a/client/simple.go
+++ b/client/simple.go
@@ -10,12 +10,15 @@ import (
 
 var (
 	// 503 error type informing rollup to failover to other DA location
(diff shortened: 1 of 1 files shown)


client/simple.go
```diff
@@ -10,12 +10,15 @@ import (
...
 var (
 	// 503 error type informing rollup to failover to other DA location
-	ErrServiceUnavailable = fmt.Errorf("eigenda service is unavailable")
+	ErrServiceUnavailable = fmt.Errorf("eigenda service is temporarily unavailable")
 )
 
-// TODO: Add support for custom http client option
 type Config struct {
-	URL string
+	URL string // EigenDA proxy RPC URL
```
samlaf, line 17: nit: it's actually a REST-ish (not quite kek) endpoint, not RPC

client/simple.go
```diff
@@ -10,12 +10,15 @@ import (
...
+}
+
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
```
samlaf, line 21: Is this really needed here? How are you mocking the client? with some mock library?
Can we use https://pkg.go.dev/net/http/httptest#Server.Client instead to simplify the code and not have this interface (don't think its typical to have an interface for an http client in golang, but not 100% sure)

client/simple.go
```diff
@@ -80,7 +99,7 @@ func (c *SimpleCommitmentClient) GetData(ctx context.Context, comm []byte) ([]by
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("received error response, code=%d, msg = %s", resp.StatusCode, string(b))
+		return nil, fmt.Errorf("received error response when reading from eigenda, code=%d, msg = %s", resp.StatusCode, string(b))
```
samlaf, line 102: we don't know that the error comes from eigenda right? it could come from the proxy itself, like some 500 or whatever

samlaf: maybe just change to "from eigenda-proxy"?

client/simple.go
```diff
@@ -106,16 +125,17 @@ func (c *SimpleCommitmentClient) SetData(ctx context.Context, b []byte) ([]byte,
...
 	}
 
+	// failover signal
 	if resp.StatusCode == http.StatusServiceUnavailable {
 		return nil, ErrServiceUnavailable
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("failed to store data: %v, err = %s", resp.StatusCode, string(b))
+		return nil, fmt.Errorf("received error response when dispersing to eigenda, code=%d, err = %s", resp.StatusCode, string(b))
```
samlaf, line 134: same here

client/simple.go
epociask: its easiest for me to just bind it to an interface for mocking - done this across quite a few projects in the past

client/simple.go
epociask: [a2e80ee](https://github.com/Layr-Labs/eigenda-proxy/pull/209/commits/a2e80ee0dd07827280a4383a2753acb8f4e4169e)

client/simple.go
epociask: [a2e80ee](https://github.com/Layr-Labs/eigenda-proxy/pull/209/commits/a2e80ee0dd07827280a4383a2753acb8f4e4169e)

client/simple.go
epociask: [a2e80ee](https://github.com/Layr-Labs/eigenda-proxy/pull/209/commits/a2e80ee0dd07827280a4383a2753acb8f4e4169e)

samlaf (approved):

PR #211, opened 2024-12-09
fix: copy `client/go.mod` to `builder` in Dockerfile

MegaRedHand:
## Fixes Issue

When building the Dockerfile at the project's root, it fails because a `go.mod` file is missing.

## Changes proposed

Copy the missing `go.mod` file (`client/go.mod`) along with the root's `go.mod`. This requires adding the file to the `Dockerfile.dockerignore` as an exception.


epociask (approved):
LGTM - ty for catching this so quickly 🎉

PR #214, opened 2024-12-11
fix(CI): Add docker build and curl check

epociask:
## Changes proposed
Introduces a new CI job which asserts that the server can be built into a docker container. The job also ensures that the health endpoint can be successfully pinged. This is done to avoid regressions where container builds fail on main when publishing to ghcr.


litt3 (approved):
lgtm

scripts/wait-for.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# poll the proxy endpoint until we get a 0 return code or 5mins have passed, in that case exit 1
+timeout_time=$(($(date +%s) + 300))
+
+while (( $(date +%s) <= timeout_time )); do
+  if curl -X GET 'http://localhost:6666/health'; then
+    exit 0
+  else
+    sleep 20
```
samlaf, line 9: sleep 2 is prob better? otherwise the ci takes minimum 20seconds given Id assume the first ping is not going to work because container is still being spun up.

scripts/wait-for.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# poll the proxy endpoint until we get a 0 return code or 5mins have passed, in that case exit 1
+timeout_time=$(($(date +%s) + 300))
```
samlaf, line 3: prob 2 mins is enough? I personally prefer cis to crash fast and I can restart them manually (restarting on a separate machine might also fix a problem if it was github/azure related)

samlaf (approved):
LGTM though I think we should lower the sleep and timeout

Issue #215, opened 2024-12-17
Inclusion proof verification doesn't permit trees consisting of a single node


litt3:
## Description
 
- The `ProcessInclusionProof` method in [merkle.go](https://github.com/Layr-Labs/eigenda-proxy/blob/main/verify/merkle.go) doesn't permit an inclusion proof of 0 length
- The means that a merkle proof with only a single node will not verify, since the root hash of such a tree doesn't include any sibling hashes (i.e., there is nothing in the inclusion proof)

## Fix

- The fix should be as simple as not explicitly failing the verification when the length of the inclusion proof is 0
    - ~~@litt3 is going to dig in to the generation of the merkle tree, to make sure there aren't any gotchas that initially caused this check to be added~~
    - [Confirmed](https://github.com/Layr-Labs/eigenda/blob/6aa7d59e8620674319af1e0fcfc6c9e22fd56462/core/serialization.go#L488) that the merkle proof consists exclusively of sibling hashes. This means that an empty proof is correct, in the case of a merkle tree containing a single node

## Notes

- Since the fix is simple, it will be made in the v1 architecture verification code
- The fix will also be included in the v2 version of verification that is being [upstreamed](https://github.com/Layr-Labs/eigenda/issues/1000) to eigenDA

PR #218, opened 2024-12-18
Fix single node tree verification bug

litt3:
- Remove the incorrect invariant check, which would fail proof with 0 length
- Add unit test to confirm 0 length proofs can succeed verification
- closes https://github.com/Layr-Labs/eigenda-proxy/issues/215


litt3:
Closing, to not use a fork

PR #219, opened 2024-12-18
Fix single node tree verification bug

litt3:
- Remove the incorrect invariant check, which would fail proof with 0 length
- Add unit test to confirm 0 length proofs can succeed verification
- closes https://github.com/Layr-Labs/eigenda-proxy/issues/215

### DIFF
--- a/verify/merkle.go
+++ b/verify/merkle.go
@@ -7,9 +7,14 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-// ProcessInclusionProof processes the Merkle root proof
+// ProcessInclusionProof computes the merkle root hash based on the provided leaf and proof, returning the result.
+// An error is returned if the proof param is malformed.
+//
+// NOTE: this method returning a nil error does NOT indicate that the proof is valid. Rather, it merely indicates that
+// the proof was well-formed. The hash returned by this method must be compared to the claimed root hash, to
+// determine if the proof is valid.
 func ProcessInclusionProof(proof []byte, leaf common.Hash, index uint64) (common.Hash, error) {
-	if len(proof) == 0 || len(proof)%32 != 0 {
+	if len(proof)%32 != 0 {
 		return common.Hash{}, errors.New("proof length should be a multiple of 32 bytes or 256 bits")
 	}
 
--- a/verify/merkle.go
+++ b/verify/merkle.go
@@ -10,6 +10,10 @@ import (
 // ProcessInclusionProof computes the merkle root hash based on the provided leaf and proof, returning the result.
 // An error is returned if the proof param is malformed.
 //
+// index is the index of the leaf in the tree, starting from the bottom left of the tree at 0.
+//
+// If the proof length is 0, then the leaf hash is returned.
+//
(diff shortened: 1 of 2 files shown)


verify/merkle.go
```diff
@@ -7,9 +7,14 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-// ProcessInclusionProof processes the Merkle root proof
+// ProcessInclusionProof computes the merkle root hash based on the provided leaf and proof, returning the result.
+// An error is returned if the proof param is malformed.
+//
+// NOTE: this method returning a nil error does NOT indicate that the proof is valid. Rather, it merely indicates that
+// the proof was well-formed. The hash returned by this method must be compared to the claimed root hash, to
+// determine if the proof is valid.
```
samlaf, line 15: Thanks for writing docstring!!
Is `ProcessInclusionProof` the most accurate name for this method? I dont have full context on where/how its result (the root) is used, but would it make more sense to rename to `ComputeMerkleRoot` or something?

verify/merkle.go
```diff
@@ -7,9 +7,14 @@ import (
...
 func ProcessInclusionProof(proof []byte, leaf common.Hash, index uint64) (common.Hash, error) {
```
samlaf, line 16: Can we add a comment explaining that index is the index of the leaf. Also is it leafIndex? or index in the tree? does it start at 0? etc

verify/merkle.go
samlaf: Ok just checked its used in `verifyMerkleProof`. Feels like verify should be in merkle.go, and ProcessInclusionProof should be private but Verify public?

samlaf: Can you also add a comment like in the onchain code that if proof is empty the leafHash will be returned.

litt3: > Can you also add a comment like in the onchain code that if proof is empty the leafHash will be returned.

Requested clarification added in [26d68ca](https://github.com/Layr-Labs/eigenda-proxy/pull/219/commits/26d68ca1b1c7bfe12c678b5c481b1865fd567bbd)

verify/merkle.go
litt3: Clarification added in [26d68ca](https://github.com/Layr-Labs/eigenda-proxy/pull/219/commits/26d68ca1b1c7bfe12c678b5c481b1865fd567bbd)

verify/merkle.go
samlaf: Yep makes sense! Thanks for the change :)

samlaf (approved):

epociask (approved):

verify/merkle.go
```diff
@@ -7,9 +7,18 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-// ProcessInclusionProof processes the Merkle root proof
+// ProcessInclusionProof computes the merkle root hash based on the provided leaf and proof, returning the result.
```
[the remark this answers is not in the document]
litt3: Done [6357d45](https://github.com/Layr-Labs/eigenda-proxy/pull/219/commits/6357d456c4ffb99d6b3d96b241af5639a79c0f16)

PR #222, opened 2024-12-21
feat: add GET header to return raw blob

bxue-l2:
This is a POC for returning raw codec-ed blob.

This POC is only tested against raw blob and other store getRaw isn't tested. 

This POC relies on http header on the http GET method to indicate if it wants the raw blob.

Recall suppose our blob is called A, the final output is IFFT(codec(A)).

The raw blob return you codec(A). An alternative way to reconstruct the header again inside the client code, which looks more verbose. 

This is only used for internal testing, the final approach is yet to be decided.

### DIFF
--- a/store/generated_key/eigenda/eigenda.go
+++ b/store/generated_key/eigenda/eigenda.go
@@ -67,6 +67,22 @@ func (e Store) Get(ctx context.Context, key []byte) ([]byte, error) {
 	return decodedBlob, nil
 }
 
+// ToDo still not correct impl
+func (e Store) GetRaw(ctx context.Context, key []byte) ([]byte, error) {
+	var cert verify.Certificate
+	err := rlp.DecodeBytes(key, &cert)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


store/generated_key/eigenda/eigenda.go
```diff
@@ -67,6 +67,22 @@ func (e Store) Get(ctx context.Context, key []byte) ([]byte, error) {
 	return decodedBlob, nil
 }
 
+// ToDo still not correct impl
+func (e Store) GetRaw(ctx context.Context, key []byte) ([]byte, error) {
```
samlaf, line 71: I'm confused, what is this supposed to do? It's doing the exact same thing as `Get()` right now.

samlaf (commented):
Don't think we should duplicate code like this. Most of the code in GetRaw everywhere is exactly the same as Get. If all it's doing is not taking the IFFT at the end or something, then right way is prob just to add a bool argument to the Get methods.

bxue-l2:
Close in favor of this PR, https://github.com/Layr-Labs/hokulea/pull/19

PR #225, opened 2025-01-08
docs: fix readme test docs, and add test comments to makefile

litt3:
- These doc updates are being made in advance of the next proxy release

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,8 @@ Unit tests can be ran via invoking `make test`.
 
 End-to-end (E2E) tests can be ran via `make e2e-test`. These tests use the [op-e2e](https://github.com/ethereum-optimism/optimism/tree/develop/op-e2e) framework for asserting correct interaction behaviors with batch submission and state derivation.
 
+These tests also assert E2E client <-> server interactions using simple/op clients.
+
 ### E2E Fuzz
 
 This E2E test will fuzz the proxy client server integration and op client keccak256 with malformed inputs. This is never meant to be fuzzed with EigenDA. Run with `make e2e-fuzz-test`.
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ End-to-end (E2E) tests can be ran via `make e2e-test`. These tests use the [op-e
 
 ### E2E Fuzz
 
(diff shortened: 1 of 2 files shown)


samlaf:
@litt3 guessing this is no longer a draft given that you requested reviews?

README.md
```diff
@@ -213,15 +213,16 @@ The `raw commitment` is an RLP-encoded [EigenDA certificate](https://github.com/
 Unit tests can be ran via invoking `make test`.
 
 ### Integration
-End-to-end (E2E) tests can be ran via `make e2e-test`.
 
-### Holesky
+End-to-end (E2E) tests can be ran via `make e2e-test`. These tests use the [op-e2e](https://github.com/ethereum-optimism/optimism/tree/develop/op-e2e) framework for asserting correct interaction behaviors with batch submission and state derivation.
```
samlaf, line 217: Why did we switch from calling the make target `optimism-test` to `e2e-test`? I'm guessing these e2e test also test some nitro stuff? If not, should we just change the name back to `test-optimism-integration`? cc @epociask

samlaf (approved):
LGTM. unsure about the naming e2e-test though, left a comment. But don't want to block the release for a small thing like this, so approving.

README.md
epociask: we initially had a separate entry-point (i.e, OP env flag) for these tests independent of the other E2E tests. Now `e2e-test` is a superset of `optimism-test`

epociask: > test some nitro stuff 

Nop we currently don't do any E2E nitro tests in proxy directly. Could be worth exploring in the future

epociask: let's also add an additional section specifying that these tests also assert E2E client <-> server interactions using simple/op clients

epociask (commented):
Nice - just one comment 🙏🏻

README.md
litt3: Added verbatim [c5d6f0d](https://github.com/Layr-Labs/eigenda-proxy/pull/225/commits/c5d6f0d4f4fed70d9debba5355d468af22e98613)

epociask (approved):
LGTM

PR #226, opened 2025-01-08
feat: blob staleness check

samlaf:
DEPRECATED: see https://github.com/Layr-Labs/eigenda-proxy/pull/378 instead

We add a check to discard blobs that have been included in the rollup's batcher inbox a long time after it was included in an EigenDA batch by the eigenda disperser.

It is now expected that GET requests to the proxy can include an optional `l1_inclusion_block_number` query argument, which will cause a blob to be rejected if it was included onchain > `RollupBlobInclusionWindow` L1 blocks after the blob's batch's RBN.

The `RollupBlobInclusionWindow` is a flag added to the proxy binary, which is set to 0 by default, which skips the check.

## Fixes Issue

Fixes #

## Note to reviewers

Commits are:
- [8eaef49](https://github.com/Layr-Labs/eigenda-proxy/pull/226/commits/8eaef499692e3d98b1552dfa8202642e08f60f99): contains the main feature logic
- [a3b6ca4](https://github.com/Layr-Labs/eigenda-proxy/pull/226/commits/a3b6ca4fdfe5d53fd17c838c2e688a8b14aabf2c): test the query param parsing logic
- [0cf0e78](https://github.com/Layr-Labs/eigenda-proxy/pull/226/commits/0cf0e781eda3a52aea181dccf486acf2bc2941ba): test the verification logic (also includes refactoring the CertVerifier to an interface to be able to test other verification that doesnt need an eth rpc connection)


common/store.go
```diff
@@ -61,8 +61,17 @@ func StringToBackendType(s string) BackendType {
...
+
+type VerifyArgs struct {
+	// L1 block number at which the rollup batch was submitted to the batcher inbox.
+	// This is optional, and should be set to -1 to mean to not verify the reference block number distance check.
+	//
+	// Used to determine the validity of the eigenDA batch.
+	// The eigenDA batch header contains a reference block number (RBN) which is used to pin the stake of the eigenda operators at that specific blocks.
+	// The rollup batch containing the eigenDA cert is only valid if it was included within a certain number of blocks after the RBN.
+	// validity condition is: RBN < l1_inclusion_block_number < RBN + some_delta
+	RollupL1InclusionBlockNum int64
```
epociask, line 74: shouldn't this be `uint64`?

verify/cert.go
```diff
@@ -19,17 +19,26 @@ import (
...
+	verifyBatchConfirmedOnChain(ctx context.Context, batchID uint32, batchMetadata *disperser.BatchMetadata) error
+	verifyMerkleProof(inclusionProof []byte, root []byte, index uint32, blobHeader BlobHeader) error
+	quorumNumbersRequired(opts *bind.CallOpts) ([]byte, error)
+	quorumAdversaryThresholdPercentages(opts *bind.CallOpts) ([]byte, error)
+}
+
+// CertVerification verifies the DA certificate against on-chain EigenDA contracts
 // to ensure disperser returned fields haven't been tampered with
-type CertVerifier struct {
+type CertVerification struct {
```
epociask, line 31: knit - any  reason to make this public if we're binding to an interface?

verify/verifier.go
```diff
@@ -40,24 +49,38 @@ func (c Config) MarshalJSON() ([]byte, error) {
...
-	verifyCerts bool
-	cv          *CertVerifier
+	// When config.VerifyCerts is false, we use a noop verifier that does nothing
+	cv CertVerifier
+	// Allowed distance (in L1 blocks) between the eigenDA reference block number (RBN) of the batch the blob is included in,
+	// and the L1 block number at which the blob cert was included in the batcher's inbox.
+	// If batch.RBN + rollupBlobInclusionWindow < cert.L1InclusionBlock, the batch is considered stale and verification will fail.
+	// This check is optional and will be skipped when rollupBlobInclusionWindow is set to 0.
+	// Note: if there are more rollup related properties that we need to check in the future, then maybe create a RollupVerifier struct
+	rollupBlobInclusionWindow uint32
```
epociask, line 69: wouldn't this need to be globally set per the rollup? Otherwise couldn't you have multiple verifiers articulating alternative views due to different values?

verify/verifier.go
```diff
@@ -66,31 +89,44 @@ func NewVerifier(cfg *Config, l log.Logger) (*Verifier, error) {
...
+func (v *Verifier) VerifyCert(ctx context.Context, cert *Certificate, args common.VerifyArgs) error {
 	// 1 - verify batch in the cert is confirmed onchain
 	err := v.cv.verifyBatchConfirmedOnChain(ctx, cert.Proof().GetBatchId(), cert.Proof().GetBatchMetadata())
 	if err != nil {
 		return fmt.Errorf("failed to verify batch: %w", err)
 	}
 
-	// 2 - verify merkle inclusion proof
+	// 2 - verify that the blob cert was submitted to the rollup's batch inbox within the allowed RollupBlobInclusionWindow window.
+	// This is to prevent timing attacks where a rollup batcher could try to game the fraud proof window by including an old DA blob that is about to expire on the DA layer
```
epociask, line 108: how will this verification look when one step proving the certificate? ie, will a challenger need to provide the inbox confirmation tx block #, if so then how do they do so in a way that isn't tamper resistant?

mocks/manager.go
```diff
@@ -1,5 +1,10 @@
 // Code generated by MockGen. DO NOT EDIT.
 // Source: github.com/Layr-Labs/eigenda-proxy/store (interfaces: IManager)
+//
+// Generated by this command:
+//
+//	mockgen -package mocks --destination ../mocks/manager.go . IManager
+//
```
epociask, line 7: There's also `make go-gen-mocks`

verify/verifier.go
```diff
@@ -66,31 +89,44 @@ func NewVerifier(cfg *Config, l log.Logger) (*Verifier, error) {
...
+	// and is hence not retrievable.
+	if args.RollupL1InclusionBlockNum >= 0 && v.rollupBlobInclusionWindow > 0 {
+		blockNumEigenDA := int64(cert.BlobVerificationProof.BatchMetadata.BatchHeader.ReferenceBlockNumber)
+		blockNumRollup := args.RollupL1InclusionBlockNum
+		// We need blockNumEigenDA < blockNumRollup < blockNumEigenDA + rollupBlobInclusionWindow
+		if !(blockNumEigenDA < blockNumRollup) {
+			return fmt.Errorf("eigenda batch reference block number (%d) needs to be < rollup inclusion block number (%d)", blockNumEigenDA, blockNumRollup)
+		}
+		if !(blockNumRollup < blockNumEigenDA+int64(v.rollupBlobInclusionWindow)) {
```
litt3, line 117: I believe you have an off-by-one error here

The comment in the config struct above states that the following constitutes a stale batch:

```
batch.RBN + RollupBlobInclusionWindow < cert.L1InclusionBlock
```

but it seems what's actually implemented is this:

```
batch.RBN + RollupBlobInclusionWindow <= cert.L1InclusionBlock
```

It might help to have consistent variable names between the two.

verify/verifier.go
```diff
@@ -66,31 +89,44 @@ func NewVerifier(cfg *Config, l log.Logger) (*Verifier, error) {
...
-	// 2 - verify merkle inclusion proof
+	// 2 - verify that the blob cert was submitted to the rollup's batch inbox within the allowed RollupBlobInclusionWindow window.
+	// This is to prevent timing attacks where a rollup batcher could try to game the fraud proof window by including an old DA blob that is about to expire on the DA layer
+	// and is hence not retrievable.
+	if args.RollupL1InclusionBlockNum >= 0 && v.rollupBlobInclusionWindow > 0 {
+		blockNumEigenDA := int64(cert.BlobVerificationProof.BatchMetadata.BatchHeader.ReferenceBlockNumber)
+		blockNumRollup := args.RollupL1InclusionBlockNum
+		// We need blockNumEigenDA < blockNumRollup < blockNumEigenDA + rollupBlobInclusionWindow
+		if !(blockNumEigenDA < blockNumRollup) {
+			return fmt.Errorf("eigenda batch reference block number (%d) needs to be < rollup inclusion block number (%d)", blockNumEigenDA, blockNumRollup)
```
litt3, line 115: Suggest adding a test case which covers this error, since the conditional is non trivial

common/store.go
samlaf: I'm using -1 to mean that this query param was not passed, and hence to skip the test. Could using a pointer to uint64 instead and use nil to mean this? I don't like either tbh... but go is dumb.

verify/cert.go
samlaf: Not really, I can change. Don't think using an interface matters here though, it's more about whether we think people will want to import and use this struct elsewhere.

verify/verifier.go
samlaf: Yes that's a good point. I wrote the code out but now's the time to discuss how we want to handle this feature. cc @bxue-l2 for discussion. I'll at least add a bit WARNING comment to the flag, saying that the feature is experimental and not really meant to be turned on? But then what's our plan for eventually turning it on?

verify/verifier.go
samlaf: Right precisely the right question to ask. Discussed this at length with Bowen. I'll need to add comments detailing this.

The 2 options for securing this check is:
1. do it in derivation pipeline
2. do it in preimage oracle contract (as you mention, this would require a zk proof that the inclusion block # is correct)

We are opting for 1, to do this check in the derivation pipeline. It's a simple check to make in the rust hokulea client, so we'll add it there. Ideally we'd add the same check to the golang derivation pipeline, but the problem is that we haven't had much success lately getting our PRs upstreamed. Also if ever a problem happens and we need to change the logic in the golang binary then we depend on optimism's team to review and merge our PR, which feels less than ideal (this is different from the rust code b/c in rust we have our own Hokulea library which projects using eigenda can import directly, and which we have full control over).

So the idea is that the rust stack will be a requirement to do the fraud proofs. But we still want normal validators running op-node to be able to follow the chain properly, which is why we are adding this check here in the proxy.

cc @bxue-l2 for any more thoughts.

common/store.go
litt3: Can't you use `0` to indicate "skip the test"?

`RollupL1InclusionBlockNum` must be strictly greater RBN, and the min value of RBN is `0`, therefore `RollupL1InclusionBlockNum == 0` isn't sensical, and is free realestate to have special meaning

verify/cert.go
samlaf: [bc6d810](https://github.com/Layr-Labs/eigenda-proxy/pull/226/commits/bc6d8109fd710c91379a13e331c2898a4524e2b4)

samlaf:
rebased on top of master to get the latest change (was one minor conflict where verifier's Verify function had change the argument name from expectedSomething to certCommitment)

mocks/manager.go
samlaf: I did use `make go-gen-mocks`. Not sure why this repo was setup to use the deprecated go mock library instead of the new uber mock repo. Updated that dependency in go.mod.

Created this issue to make a reproducible dev env: https://github.com/Layr-Labs/eigenda-proxy/issues/236

verify/verifier.go
bxue-l2: yes, this require an upgrade process like what op currently has. not very good

bxue-l2: we can put the staleness gap in the cert, signed by DA network

common/store.go
```diff
@@ -61,8 +61,17 @@ func StringToBackendType(s string) BackendType {
...
-	// Verify verifies the given key-value pair.
-	Verify(ctx context.Context, key []byte, value []byte) error
+}
+
+type VerifyArgs struct {
+	// L1 block number at which the rollup batch was submitted to the batcher inbox.
+	// This is optional, and should be set to -1 to mean to not verify the reference block number distance check.
+	//
+	// Used to determine the validity of the eigenDA batch.
+	// The eigenDA batch header contains a reference block number (RBN) which is used to pin the stake of the eigenda operators at that specific blocks.
```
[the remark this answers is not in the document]
samlaf: [e633096](https://github.com/Layr-Labs/eigenda-proxy/pull/226/commits/e633096cd33d5982b5ded8ff2d47a788d1bdad93)

verify/verifier.go
samlaf: [47ff9ba](https://github.com/Layr-Labs/eigenda-proxy/pull/226/commits/47ff9ba71a9047b80f7b8af7d3f702d0fb0c97f8)

verify/verifier.go
samlaf: I think it was doing < everywhere? But I agree it should be <=, so changed. Also updates comments and made naming consistent: [72451c6](https://github.com/Layr-Labs/eigenda-proxy/pull/226/commits/72451c6f3d649c2da206639b028d5d9737d9abd8)

common/store.go
litt3: > but I think its technically possible, especially on a devnet

I don't see how this would be possible

The eigenDA cert directly contains the RBN, so you can't honestly construct the cert until the reference block exists. And if the RB already exists, how could the eigenDA cert be put on chain any earlier than `RBN+1`?

verify/cert.go
```diff
@@ -166,3 +175,29 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
...
+func (cv *certVerification) quorumNumbersRequired(opts *bind.CallOpts) ([]byte, error) {
+	return cv.manager.QuorumNumbersRequired(opts)
+}
+
+func (cv *certVerification) quorumAdversaryThresholdPercentages(opts *bind.CallOpts) ([]byte, error) {
+	return cv.manager.QuorumAdversaryThresholdPercentages(opts)
+}
+
+// NoopCertVerifier is used in place of CertVerification when certificate verification is disabled
+type NoopCertVerifier struct{}
```
litt3, line 188: Any reason you left this public?

verify/verifier.go
litt3: The discrepancy was in the config doc:

```
// If batch.RBN + RollupBlobInclusionWindow < cert.L1InclusionBlock, the batch is considered stale and verification will fail.
```

Since this doc is defining what constitutes _staleness_, the implied invariant would be the _inverse_

```
cert.L1InclusionBlock <= batch.RBN + RollupBlobInclusionWindow
```

Your most recent changeset has things correct, since you modified the invariant, but left the staleness definition in the `Config` struct unchanged. Perhaps worth considering stating only the invariant, or only the definition of staleness, to decrease the likelihood for such discrepancies.

common/store.go
samlaf: Started new PR to rebase cleanly: https://github.com/Layr-Labs/eigenda-proxy/pull/378

updated in: [2eddc29](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/2eddc29e6489ff1e920f15a8fdc4c48325fe5474)

verify/verifier.go
samlaf: Let's continue the discussion [here](https://github.com/Layr-Labs/eigenda-proxy/pull/378#discussion_r2063787470).

verify/verifier.go
samlaf: Started new PR to rebase cleanly: https://github.com/Layr-Labs/eigenda-proxy/pull/378

Added comment in [a76737d](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/a76737d4b67029904b3fe067fa0dbc9ceada35dd)

verify/verifier.go
samlaf: Don't think there's an issue with that comment.

The invariant to respect is 
```
cert.L1InclusionBlock <= batch.RBN + RollupBlobInclusionWindow
```
but so naturally if that is not true that
```
cert.L1InclusionBlock > batch.RBN + RollupBlobInclusionWindow
```
is true, then the batch is considered stale.

samlaf:
Closing as we have decided to not do this check inside proxy, and instead provide golang and rust libraries to do these checks inside the derivation pipeline directly.

PR #227, opened 2025-01-09
fix(sigmap-EDAP-04): Add IO parsing size constraint on incoming request bodies

epociask:
## Fixes Issue

Fixes #

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -12,6 +12,11 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	// limit requests to only 32 mib to mitigate potential DoS attacks
+	maxRequestBodySize int64 = 1048576 * 32
+)
+
 func (svr *Server) handleHealth(w http.ResponseWriter, _ *http.Request) error {
 	w.WriteHeader(http.StatusOK)
 	return nil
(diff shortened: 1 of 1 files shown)


server/handlers.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	// limit requests to only 32 mib to mitigate potential DoS attacks
+	maxRequestBodySize int64 = 1048576 * 32
```
litt3, line 17: IMO `1024 * 1024 * 32` might be a clearer representation. Not sure `1048576` is common knowledge (I personally had to do the math to check it)

litt3 (approved):

server/handlers.go
epociask: [48c8cdc](https://github.com/Layr-Labs/eigenda-proxy/pull/227/commits/48c8cdc810a2c1dee50811e7f40413302d01422a)

server/handlers.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	// limit requests to only 32 mib to mitigate potential DoS attacks
```
samlaf, line 16:
```suggestion
	// limit requests to only 32 MiB to mitigate potential DoS attacks
```

PR #228, opened 2025-01-09
fix(sigmap-EDAP-04): Miscellaneous General Comments

epociask:
## Fixes Issue

Fixes #

### DIFF
--- a/verify/cli.go
+++ b/verify/cli.go
@@ -20,7 +20,7 @@ var (
 
 // TODO: should this live in the resources pkg?
 // So that if we ever change the SRS files there we can change this value
-const srsOrder = 268435456 // 2 ^ 32
+const srsOrder = 268435456 // 2 ^ 28
 
 var (
 	// cert verification flags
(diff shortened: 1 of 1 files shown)


litt3 (approved):

verify/cli.go
```diff
@@ -20,7 +20,7 @@ var (
 
 // TODO: should this live in the resources pkg?
 // So that if we ever change the SRS files there we can change this value
-const srsOrder = 268435456 // 2 ^ 32
+const srsOrder = 268435456 // 2 ^ 28
```
samlaf, line 23:
```suggestion
const srsOrder = 1 << 28
```

samlaf: Should we also remove `MaxSRSPoints` above and use srsOrder instead (since its hardcoded anyways)?

samlaf (changes requested):

verify/cli.go
epociask: [bc7642e](https://github.com/Layr-Labs/eigenda-proxy/pull/228/commits/bc7642e2944f7f9bdefac29b45c130ba00a68066)

samlaf (approved):

PR #229, opened 2025-01-09
fix(sigmap-EDAP-06): Missing IsOnCurve & IsInSubgroup Checks For Elli…

epociask:
## Fixes Issue

Fixes #

### DIFF
--- a/verify/verifier.go
+++ b/verify/verifier.go
@@ -124,27 +124,27 @@ func (v *Verifier) Commit(blob []byte) (*bn254.G1Affine, error) {
 // Verify regenerates a commitment from the blob and asserts equivalence
 // to the commitment in the certificate
 // TODO: Optimize implementation by opening a point on the commitment instead
-func (v *Verifier) VerifyCommitment(expectedCommit *common.G1Commitment, blob []byte) error {
+func (v *Verifier) VerifyCommitment(certCommitment *common.G1Commitment, blob []byte) error {
 	actualCommit, err := v.Commit(blob)
 	if err != nil {
 		return err
 	}
 
 	expectedX := &fp.Element{}
-	expectedX.Unmarshal(expectedCommit.X)
+	expectedX.Unmarshal(certCommitment.X)
 
 	// map coordinates to G1 and ensure they are on the curve
 	xAffine := bn254.MapToG1(*expectedX)
 	if !xAffine.IsOnCurve() {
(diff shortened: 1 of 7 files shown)


verify/verifier.go
```diff
@@ -132,9 +132,21 @@ func (v *Verifier) VerifyCommitment(expectedCommit *common.G1Commitment, blob []
 
 	expectedX := &fp.Element{}
 	expectedX.Unmarshal(expectedCommit.X)
+
+	// map coordinates to G1 and ensure they are on the curve
+	xAffine := bn254.MapToG1(*expectedX)
+	if !xAffine.IsOnCurve() {
+		return fmt.Errorf("expected x is not on the curve: %x", expectedX.Marshal())
```
samlaf, line 139: "expected x" feels like a very testy variable name. Maybe change to `claimedCommitment` everywhere instead?

samlaf (commented):
Change LGTM but I'd prefer to change the variable name.

Even the documentation comment mentions "commitment in the certificate". Why are we using something abstract like `expectedCommit`? Can we just change to `certificateCommitment`?

verify/verifier.go
epociask: but commitment is a constitution of (x,y) field elements.... trying to think of better name here

verify/verifier.go
```diff
@@ -124,16 +124,28 @@ func (v *Verifier) Commit(blob []byte) (*bn254.G1Affine, error) {
...
-func (v *Verifier) VerifyCommitment(expectedCommit *common.G1Commitment, blob []byte) error {
+func (v *Verifier) VerifyCommitment(certCommitment *common.G1Commitment, blob []byte) error {
 	actualCommit, err := v.Commit(blob)
 	if err != nil {
 		return err
 	}
 
 	expectedX := &fp.Element{}
-	expectedX.Unmarshal(expectedCommit.X)
+	expectedX.Unmarshal(certCommitment.X)
```
samlaf, line 134: can we change all expected -> certCommitment

samlaf (approved):
Approving but would still like to get the last variable name change in. Night ;)

verify/verifier.go
epociask: [dc39a19](https://github.com/Layr-Labs/eigenda-proxy/pull/229/commits/dc39a19adc7105241a6240cb9b5a0ec61785b250)

PR #231, opened 2025-01-09
fix(sigmap-EDAP-01): Missing nil Checks On Parameters Of Incoming Requests

epociask:
## Fixes Issue

Fixes #


litt3 (approved):

samlaf (approved):
LGTM. check for nils to return error early, but still use the protobuf getter methods in case we forgot to check some nil statement or the code ever gets refactored etc.

PR #237, opened 2025-01-10
chore: remove routing duplicate route

samlaf:
not sure what exactly this does. Guessing it didn't affect the semantics, but still better to clean up.

## Fixes Issue

Fixes #


samlaf:
Nvm non issue, these are not duplicates. It's because we want to handle both `put` and `put/`. Closing.

PR #241, opened 2025-01-14
chore: modify verifier to not require eth archive node

### LINKED ISSUE #230: Remove archive node requirements
Right now the verifier logic requires an [rpc url](https://github.com/Layr-Labs/eigenda-proxy/blob/0fb5ae801140b2518c305d572fc82b4a3772f68b/verify/verifier.go#L26) to an **archive** node.

This is because it needs to retrieve information from the [eigenDAServiceManager contract](https://github.com/Layr-Labs/eigenda-proxy/blob/0fb5ae801140b2518c305d572fc82b4a3772f68b/verify/cert.go#L28) at old blocks, for eg:
1. at the batch's [confirmation block number](https://github.com/Layr-Labs/eigenda-proxy/blob/0fb5ae801140b2518c305d572fc82b4a3772f68b/verify/cert.go#L68)
2. at the batch's [reference block number](https://github.com/Layr-Labs/eigenda-proxy/blob/0fb5ae801140b2518c305d572fc82b4a3772f68b/verify/verifier.go#L180)

Creating this issue to discuss ways to remove this dependency, and allow proxy to be run with a normal (non-archive) node rpc. Some obvious suggestion is to emit events in the contract and run some kind of indexer, basically moving the dependency from an archive node to a graph node. It isn't clear to me this is much better...

EigenDA v2 will remove some needs for the archive node dependency, but I'm not sure it removes **all** need atm. For eg. we'll still need to reference operator state at a previous block number (RBN), but this can be done at the current block because we store all state updates in the contract state itself.

### Action Items
1. Add README documentation for what/when archive node is required
2. Design V2 such that we don't need archive node (move requiredQuorums history into the contract state?)

samlaf:
## Fixes Issue

Fixes #230 

## Changes proposed

Contains 2 main changes needed to remove archive node requirements:
1. [4d85d3b](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/4d85d3b01e4f67a4610f0a1cc9cde1ec224b70e7): force ethConfirmationDepth to be <64 (so we read at most 64 blocks in the past, while normal nodes store the latest 128 states): not strictly necessary as this was a flag which could have been set by teh validators anyways, but its a nice to have to make sure that the code works even without archival node dependency.
2. [c4e6c90](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/c4e6c90059216228d8340031f005861e7189b7b0): read the quorum parameters (required quorums and adversary thresholds) from the EigenDAServiceManager only once, at startup.

Also updated the README with this information in [cc3538f](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/cc3538f39c9e4242a6b4ad6c025681c8b311173c)

Copying the comment on the `getQuorumParametersAtLatestBlock` function here to highlight the solution proposed in this PR:
> // Note: this strategy (fetching once and caching) only works because these parameters are immutable.
// They might be different in different environments (for eg on a devnet or testnet), but they are fixed on a given network.
// We used to allow these parameters to change (via a setter function on the contract), but that then forced us here in the proxy
// to query for these parameters on every request, at the batch's reference block number (RBN).
// This in turn required rollup validators running this proxy to have an archive node, in case the RBN was >128 blocks in the past,
// which was not ideal. So we decided to make these parameters immutable, and cache them here.


README.md
```diff
@@ -121,6 +121,9 @@ The following specs are recommended for running on a single production server:
 * 4 GB RAM
 * 1-2 cores CPU
 
+### Ethereum Node Requirements
+The cert verification logic inside proxy used to require an archive node to fetch quorum information at reference block numbers in the past. We have removed this requirement by making the quorum parameters immutable in the EigenDAServiceManager contract. This means that a normal Ethereum node can now be used to run the proxy. See https://github.com/Layr-Labs/eigenda-proxy/issues/230 for more details.
```
epociask, line 125: This summary feels like something better encapsulated in release notes and within our README

verify/cert.go
```diff
@@ -155,7 +176,10 @@ func (cv *CertVerifier) getConfDeepBlockNumber(ctx context.Context) (*big.Int, e
 }
 
 // retrieveBatchMetadataHash retrieves the batch metadata hash stored on-chain at a specific blockNumber for a given batchID
-// returns an error if some problem calling the contract happens, or the hash is not found
+// returns an error if some problem calling the contract happens, or the hash is not found.
+// We make an eth_call to the EigenDAServiceManager at the given blockNumber to retrieve the hash.
+// Therefore, make sure that blockNumber is <128 blocks behind the latest block, to avoid requiring an archive node.
+// This is currently enforced by having EthConfirmationDepth be <64.
```
epociask, line 182: Should we poll the latest block number and ensure it's greater than the confirmation block number?

verify/cert.go
```diff
@@ -22,14 +23,27 @@ import (
 // CertVerifier verifies the DA certificate against on-chain EigenDA contracts
 // to ensure disperser returned fields haven't been tampered with
 type CertVerifier struct {
-	l                    log.Logger
+	l log.Logger
+	// ethConfirmationDepth is using to verify that a blob's batch has been bridged to the EigenDAServiceManager contract at least
+	// this many blocks in the past. To do so we make an eth_call to the contract at the current block_number - ethConfirmationDepth.
+	// Hence in order to not require an archive node, this value should be kept low. We force it to be < 64.
+	// waitForFinalization should be used instead of ethConfirmationDepth if the user wants to wait for finality (typically 64 blocks in happy case).
```
epociask, line 30:
```suggestion
	// ethConfirmationDepth is used to verify that a blob's batch commitment has been bridged to the EigenDAServiceManager contract at least
	// this many blocks in the past. To do so we make an eth_call to the contract at the current block_number - ethConfirmationDepth.
	// Hence in order to not require an archive node, this value should be kept low. We force it to be < 64.
	// waitForFinalization should be used instead of ethConfirmationDepth if the user wants to wait for finality (typically 64 blocks in happy case).
```

verify/cert.go
epociask: i.e, do we care about the case where the RPC node is far behind?

README.md
samlaf: changed in [dc0b205](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/dc0b205bdfabc99be9fddfd1b4b86d1f9da4e076)

verify/cert.go
samlaf: Good question. I personally don't think so. For 3 reasons:
1. what would we do anyways? sleep until that block catches up? Don't think this is a valid strategy. Think its better to return an error and let the outer service (batcher/node) retry. This will also raise alerts and get the human in the loop to intervene
2. sequencer should have an up to date eth node as a requirement
3. validators will typically be many many blocks ahead, so their RPC would have to be suuuuuper stale anyways for this case to happen

verify/cert.go
```diff
@@ -22,14 +23,27 @@ import (
...
 	ethClient            *ethclient.Client
+	// The two fields below are fetched from the EigenDAServiceManager contract in the constructor.
+	// They are used to verify the quorums in the received certificates.
+	// See getQuorumParametersAtLatestBlock for more details.
+	quorumsRequired           []uint8
+	quorumAdversaryThresholds map[uint8]uint8
 }
 
 func NewCertVerifier(cfg *Config, l log.Logger) (*CertVerifier, error) {
+	if cfg.EthConfirmationDepth >= 64 {
```
litt3, line 43: consider creating a constant, for easy visibility of this value, and reuse

verify/cert.go
```diff
@@ -155,7 +176,10 @@ func (cv *CertVerifier) getConfDeepBlockNumber(ctx context.Context) (*big.Int, e
 }
 
 // retrieveBatchMetadataHash retrieves the batch metadata hash stored on-chain at a specific blockNumber for a given batchID
-// returns an error if some problem calling the contract happens, or the hash is not found
+// returns an error if some problem calling the contract happens, or the hash is not found.
+// We make an eth_call to the EigenDAServiceManager at the given blockNumber to retrieve the hash.
+// Therefore, make sure that blockNumber is <128 blocks behind the latest block, to avoid requiring an archive node.
```
litt3, line 181: I'm not understanding where 128 is coming from. Is there some other value == 64 that gets added to `EthConfirmationDepth` to produce 128? Or is the comment just stale?

verify/cert.go
```diff
@@ -166,3 +190,47 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
 	}
 	return onchainHash, nil
 }
+
+// getQuorumParametersAtLatestBlock fetches the required quorums and quorum adversary thresholds
+// from the EigenDAServiceManager contract at the latest block.
+// We then cache these parameters and use them in the Verifier to verify the certificates.
+//
+// Note: this strategy (fetching once and caching) only works because these parameters are immutable.
+// They might be different in different environments (for eg on a devnet or testnet), but they are fixed on a given network.
```
litt3, line 199: nit: "for eg" is redundant, "e.g." expands to "for example"

verify/cert.go
```diff
@@ -166,3 +190,47 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
...
+// We used to allow these parameters to change (via a setter function on the contract), but that then forced us here in the proxy
+// to query for these parameters on every request, at the batch's reference block number (RBN).
+// This in turn required rollup validators running this proxy to have an archive node, in case the RBN was >128 blocks in the past,
```
litt3, line 202: Here's another 128 (in case this is a stale comment, see above)

verify/cert.go
```diff
@@ -166,3 +190,47 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
...
+// which was not ideal. So we decided to make these parameters immutable, and cache them here.
+func getQuorumParametersAtLatestBlock(
+	manager *binding.ContractEigenDAServiceManagerCaller,
+) ([]uint8, map[uint8]uint8, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
litt3, line 207: How did you settle on 5 seconds? Would it be worth making configurable? (maybe not considering imminent v1 deprecation)

verify/cert.go
```diff
@@ -166,3 +190,47 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
...
+	}
+	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	thresholds, err := manager.QuorumAdversaryThresholdPercentages(&bind.CallOpts{Context: ctx})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to fetch QuorumAdversaryThresholdPercentages from EigenDAServiceManager: %w", err)
+	}
+	var quorumAdversaryThresholds = make(map[uint8]uint8)
+	for quorumNum, threshold := range thresholds {
+		if quorumNum > math.MaxInt8 {
```
litt3, line 221: It seems odd to use `range`, and then check on each iteration that the index isn't too large, or negative. 

Recommend simply checking that length is <= MaxInt8 before the loop

verify/cert.go
```diff
@@ -22,14 +23,27 @@ import (
...
+	// The two fields below are fetched from the EigenDAServiceManager contract in the constructor.
+	// They are used to verify the quorums in the received certificates.
+	// See getQuorumParametersAtLatestBlock for more details.
+	quorumsRequired           []uint8
+	quorumAdversaryThresholds map[uint8]uint8
 }
 
 func NewCertVerifier(cfg *Config, l log.Logger) (*CertVerifier, error) {
+	if cfg.EthConfirmationDepth >= 64 {
+		// We keep this low (<128) to avoid requiring an archive node.
```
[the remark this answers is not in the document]
samlaf: Ah, sorry for the confusion. 128 is the magic number for eth rpcs because normal (non-archive) nodes only store the latest 128 states as the chain advances. The number 64 is because finalization in happy-path takes exactly 64 blocks (2 epochs). So we could use 128 instead of 64, but 64 makes sense given that we have another flag "WaitForFinalization" or something which we recommend people use instead of EthConfirmationDepth if they are waiting a lot (>2 epochs) worth of blocks, given that at that point the chain has in most cases already finalized anyways.

verify/cert.go
samlaf: See https://github.com/Layr-Labs/eigenda-proxy/pull/241#discussion_r1927216570

verify/cert.go
samlaf: See https://github.com/Layr-Labs/eigenda-proxy/pull/241#discussion_r1927216570

verify/cert.go
samlaf: Arbitrary number that I typically use everywhere for calls (as its ample time to request an eth_call even from a node that halfway across the globe). Indeed don't think its worth creating a flag for this given v1 deprecation.

Overall this is another thing that I would to put in the style guide. Not sure if you have an opinion about this, some people don't like for constructors to make network calls. because what I should be doing is adding a context to the constructor to hint at this, and let call sites pass in the context that they want. But looking quickly that just bubbled up like 4 levels of constructors... so not sure what the right strategy is here, and don't really want to spend too much time on this atm, unless you have a good idea.

verify/cert.go
samlaf: [ec8d910](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/ec8d910a2a3cc7777b37024fa6f8acd52dd4dca3)

verify/cert.go
samlaf: I was scared of disabling the stupid G115 linter (see for eg https://dev.to/ccoveille/about-the-gosec-g115-drama-or-how-i-faced-back-integer-conversion-overflow-in-go-1302)
But you're right that this was just dumb: [4a86753](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/4a86753f0ee2c8080a73e76174a22e9b4bdfe106)

verify/cert.go
samlaf: [275acd6](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/275acd6be9116c340c3595baec1280c0d073a83b)

verify/cert.go
litt3: I don't have a great idea for this - no complaints leaving it as-is

verify/cert.go
litt3: Annoying that the linter isn't smart enough to handle this...

litt3 (approved):

verify/cert.go
EthenNotEthan: at a minimum we could fail loudly within proxy indicating that the node has a stale state. agree though its a performance degradation (i.e, additional rpc query) in-favor of better devx.

common/consts/consts.go
```diff
@@ -0,0 +1,8 @@
+package consts
+
+// EthHappyPathFinalizationDepth is the number of blocks that must be included on top of a block for it to be considered "final",
+// under happy-path aka normal network conditions.
+//
+// See https://docs.optimism.io/stack/transactions/transaction-finality for a quick TLDR explanation,
```
EthenNotEthan, line 6: feel like this is a counter-intuitive explanation since it refers to the lifecycle of a rollup tx. Prefer we use more L1 agnostic terminology like [commitment levels](https://www.alchemy.com/overviews/ethereum-commitment-levels)

flags/eigendaflags/cli.go
```diff
@@ -209,8 +210,13 @@ func validateConfirmationFlag(val string) error {
...
 	}
 
-	if depth >= 64 {
-		log.Printf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Consider using 'finalized' instead.\n", depth)
+	if depth >= uint64(consts.EthHappyPathFinalizationDepthBlocks) {
+		// We keep this low (<128) to avoid requiring an archive node (see how this is used in CertVerifier).
+		// Note: assuming here that no sane person would ever need to set this to a number to something >64.
+		// But perhaps someone testing crazy reorg scenarios where finalization takes >2 epochs might want to set this to a higher number...?
+		// Let's deal with that case if and when it comes up (ideally never). Do keep in mind if you ever change this
+		// that it might affect a LOT of validators on your rollup who would now need an archival node.
```
EthenNotEthan, line 218: knit - feel like the first two comments adequately describe the methodology. The latter 3 increase cognitive complexity imo

verify/cert.go
```diff
@@ -166,3 +191,44 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
 	}
 	return onchainHash, nil
 }
+
+// getQuorumParametersAtLatestBlock fetches the required quorums and quorum adversary thresholds
+// from the EigenDAServiceManager contract at the latest block.
+// We then cache these parameters and use them in the Verifier to verify the certificates.
+//
+// Note: this strategy (fetching once and caching) only works because these parameters are immutable.
```
EthenNotEthan, line 199: immutable per the implementation contract right? IIUC we could technically still upgrade the source proxy contract with a new implementation with alternative bytecode defining different immutable values

verify/cert.go
```diff
@@ -166,3 +191,44 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
...
+		return nil, nil, fmt.Errorf("failed to fetch QuorumNumbersRequired from EigenDAServiceManager: %w", err)
+	}
+	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	thresholds, err := manager.QuorumAdversaryThresholdPercentages(&bind.CallOpts{Context: ctx})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to fetch QuorumAdversaryThresholdPercentages from EigenDAServiceManager: %w", err)
+	}
+	if len(thresholds) > math.MaxUint8 {
+		return nil, nil, fmt.Errorf("thresholds received from ServiceManager contains %d > 256 quorums, which isn't possible", len(thresholds))
```
EthenNotEthan, line 221:
```suggestion
		return nil, nil, fmt.Errorf("thresholds received from EigenDAServiceManager contains %d > 256 quorums, which isn't possible", len(thresholds))
```

verify/cert.go
```diff
@@ -166,3 +191,44 @@ func (cv *CertVerifier) retrieveBatchMetadataHash(ctx context.Context, batchID u
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to fetch QuorumNumbersRequired from EigenDAServiceManager: %w", err)
+	}
+	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	thresholds, err := manager.QuorumAdversaryThresholdPercentages(&bind.CallOpts{Context: ctx})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to fetch QuorumAdversaryThresholdPercentages from EigenDAServiceManager: %w", err)
+	}
+	if len(thresholds) > math.MaxUint8 {
```
EthenNotEthan, line 220: what about the opposite case where `len(thresholds) == 0`?

verify/cert.go
samlaf: Don't think we should fail loudly. It might just be a temporary glitch. That's why retries exist. Failure should happen after some number of retries from the outer loop service.

flags/eigendaflags/cli.go
```diff
@@ -118,6 +118,7 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			Category: category,
 		},
 		&cli.BoolFlag{
+			// This flag is DEPRECATED. Use ConfirmationDepthFlagName, which accept "finalization" or a number <64.
```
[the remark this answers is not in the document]
samlaf: Yes but I prefer to keep it there. An action typically is not meant for deprecation, so on a first reading without the comment one would read through this entire flag and it would take some brain processing to figure out its deprecated. Typically some redundancy is always good in programming. Also helps catch errors.

verify/cert.go
samlaf: Yes, but you're the one asking me to reduce comment cognitive overload :P
Do you think I should add that explanation?

verify/cert.go
samlaf: That's fine I think. It'll also be caught by the `Sanity check` at the bottom if those missing quorums are required. If they are not... then whatever the contract should be the source of truth, not the proxy.

flags/eigendaflags/cli.go
```diff
@@ -209,8 +210,13 @@ func validateConfirmationFlag(val string) error {
...
 
-	if depth >= 64 {
-		log.Printf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Consider using 'finalized' instead.\n", depth)
+	if depth >= uint64(consts.EthHappyPathFinalizationDepthBlocks) {
+		// We keep this low (<128) to avoid requiring an archive node (see how this is used in CertVerifier).
+		// Note: assuming here that no sane person would ever need to set this to a number to something >64.
+		// But perhaps someone testing crazy reorg scenarios where finalization takes >2 epochs might want to set this to a higher number...?
+		// Let's deal with that case if and when it comes up (ideally never). Do keep in mind if you ever change this
+		// that it might affect a LOT of validators on your rollup who would now need an archival node.
+		panic(fmt.Sprintf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Use 'finalized' instead.\n", depth))
```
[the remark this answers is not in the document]
samlaf: [0739f3b](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/0739f3b4a126d7c7c327e7e46715e95b213f9313)

common/consts/consts.go
samlaf: nice link. updated: [f255bc8](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/f255bc8679b056cec4d9086cf47601dfe9a40abd)

flags/eigendaflags/cli.go
samlaf: Overall disagree. Don't think we should be looking to reduce cognitive complexity of comments on flags that are read like one per year, at most. If anything, the person who will want to change that flag value will definitely appreciate having as much context as possible.
Did cleanup the comment though, let me know what you think: [9c43884](https://github.com/Layr-Labs/eigenda-proxy/pull/241/commits/9c438846c3f784b68a640ec1500e4dc919495c20)

verify/cert.go
```diff
@@ -22,14 +22,22 @@ import (
 // CertVerifier verifies the DA certificate against on-chain EigenDA contracts
 // to ensure disperser returned fields haven't been tampered with
 type CertVerifier struct {
-	l                    log.Logger
+	l log.Logger
+	// ethConfirmationDepth is using to verify that a blob's batch has been bridged to the EigenDAServiceManager contract at least
+	// this many blocks in the past. To do so we make an eth_call to the contract at the current block_number - ethConfirmationDepth.
+	// Hence in order to not require an archive node, this value should be kept low. We force it to be < 64.
+	// waitForFinalization should be used instead of ethConfirmationDepth if the user wants to wait for finality (typically 64 blocks in happy case).
 	ethConfirmationDepth uint64
```
bxue-l2, line 32: Is it correct to state this variable state the upper bound for the punctuality gap. 

Tangential question, how to ensure all proxy set the same parameters. Although not part of our job.

verify/cert.go
EthenNotEthan: nah not necessary for an explanation - more just calling out a potential risk. i.e if the service manager were to be upgraded arbitrarily to use a different quorum then proxy would have no conception? maybe lets callout in a small issue

verify/cert.go
bxue-l2, line 32: Looks to me this variable is the upper bound for the punctuality gap.

In v2 we don't have to worry about it.

Tangential question, how to ensure all proxy set the same parameters, and upgrades.

EthenNotEthan (approved):
LGTM

flags/eigendaflags/cli.go
```diff
@@ -210,7 +210,12 @@ func validateConfirmationFlag(val string) error {
 	}
 
 	if depth >= 64 {
-		log.Printf("Warning: confirmation depth set to %d, which is > 2 epochs (64). Consider using 'finalized' instead.\n", depth)
+		// We keep this low (<128) to avoid requiring an archive node (see how this is used in CertVerifier).
+		// Note: assuming here that no sane person would ever need to set this to a number to something >64.
```
bxue-l2: why this is insane to set > 64. Is it reasonable assumption

verify/cert.go
samlaf: Yep agree. Thanks for raising this. Will go ahead and merge as is, but let's keep in mind the entire space of interaction with the smart contracts as we go forward, especially for V2 as we try to make the SecurityThresholds upgradeable again.

PR #253, opened 2025-01-22
Required quorums glitch

### LINKED ISSUE #252: bug: requiredQuorums glitch on holesky
We made a smart contract upgrade mistake on holesky. Between blocks 2950538 and 2958297, required quorum was changed from 1 (which requires both quorum 0 and 1) to quorum 0 (only requiring quorum 0).
```
❯ cast call -r https://rpc.ankr.com/eth_holesky -b 2950537 0xD4A7E1Bd8015057293f0D0A557088c286942e84b "quorumNumbersRequired() (bytes)"
0x0001
❯ cast call -r https://rpc.ankr.com/eth_holesky -b 2950538 0xD4A7E1Bd8015057293f0D0A557088c286942e84b "quorumNumbersRequired() (bytes)"
0x00
❯ cast call -r https://rpc.ankr.com/eth_holesky -b 2958296 0xD4A7E1Bd8015057293f0D0A557088c286942e84b "quorumNumbersRequired() (bytes)"
0x00
❯ cast call -r https://rpc.ankr.com/eth_holesky -b 2958297 0xD4A7E1Bd8015057293f0D0A557088c286942e84b "quorumNumbersRequired() (bytes)"
0x0001
```

Right now any rollup that has saved the certs on some long-term storage like S3, when trying to archival sync, would hit these certs which would fail verification (because proxy will read contract on latest block after this [PR](https://github.com/Layr-Labs/eigenda-proxy/pull/241) is merged).

Easy(ish) fix is to hardcode on holesky (behind a holesky flag probably), that any batches with RBN between 2950538 and 2958297 should use requiredQuorum=0.

One extra related bug is that our v1 disperser is actually not looking up the requiredQuorums at the RBN block number, but is instead querying the chain's current block at the moment it is creating the batch. So unfortunately there might be some batches outside of the above range (2950538 and 2958297 ) that have been created with only quorum 0. 

## Proposed Fix
I would propose instead just being conservative and only verifying quorum0 for all blob whose batch's RBN is in some conservation interval encompassing the [2950538, 2958297] range. Perhaps something like [2950000, 2960000]

gastonponti:
## Fixes Issue

Closes #252 

Depends on #241 (merge only after the 241 is part of main)

## Changes proposed

As is stated in the issue, it avoids the quorum 1 check for the range of blocks [2950000, 2960000) for Holesky

## Note to reviewers

Instead of using a flag passed through the cli, I'm checking the svc address, which will change for the v2, but I'm assuming that this issue will be only for the v1 users.
This check avoids cascading a flag in a lot of places just because of this particular bug. With this approach the custom changes live only in the verifier.


verify/verifier.go
```diff
@@ -17,6 +18,10 @@ import (
 	"github.com/Layr-Labs/eigenda/encoding/rs"
 )
 
+var (
+	HOLESKY_SVG_V1_ADDRESS = "0xD4A7E1Bd8015057293f0D0A557088c286942e84b"
```
samlaf, line 22:
```suggestion
	HOLESKY_SVM_V1_ADDRESS = "0xD4A7E1Bd8015057293f0D0A557088c286942e84b"
```

samlaf: What is SVG? SVM = ServiceManager

verify/verifier.go
```diff
@@ -17,6 +18,10 @@ import (
...
+)
```
EthenNotEthan, line 23: knit
```suggestion
const (
	HOLESKY_SVC_MANAGER_V1_ADDRESS = "0xD4A7E1Bd8015057293f0D0A557088c286942e84b"
)
```

verify/verifier.go
```diff
@@ -183,11 +191,26 @@ func (v *Verifier) verifySecurityParams(blobHeader BlobHeader, batchHeader *disp
 	}
 
 	// ensure that required quorums are present in the confirmed ones
-	for _, quorum := range v.cv.quorumsRequired {
+	for _, quorum := range requiredQuorum(int64(batchHeader.ReferenceBlockNumber), v) {
```
EthenNotEthan, line 194: Why cast every-time vs taking a `uint32` as input to `requiredQuorum`?

verify/verifier.go
```diff
@@ -183,11 +191,26 @@ func (v *Verifier) verifySecurityParams(blobHeader BlobHeader, batchHeader *disp
...
 		}
 	}
 
 	return nil
 }
+
+func requiredQuorum(referenceBlockNumber int64, v *Verifier) []uint8 {
+	// This check is required due to a bug we had when we updated the EigenDAServiceManager in Holesky. For a brief period of time, the quorum 1 was not
+	// required for the commitment to be confirmed, which will not validate archive blobs required for full syncs (archive nodes).
+	// This check is only for testnet and for a specific block range.
```
samlaf, line 206:
```suggestion
	// required for the commitment to be confirmed, so the disperser created batches with only quorum 0 signatures. 
	// Archive nodes trying to sync from these stored batches would thus fail validation here since
	// quorumsRequired is read from the latestBlock, where the bug has been fixed and both quorums are required.
	// This check is only for testnet and for a specific block range.
```

samlaf (commented):
This approach is fine with me. @gastonponti did you want to try an archive sync using this branch just to double check that it works on your end before we merge it?

verify/verifier.go
gastonponti: Yes, my bad

verify/verifier.go
gastonponti: I think I left it after rebasing this branch from the one removing the archive nodes necessity.
I agree, is unnecessary. I'll change it

EthenNotEthan (approved):
LGTM

gastonponti:
I'm not able to reopen it, but I can create a new one

gastonponti:
Created it here: https://github.com/Layr-Labs/eigenda-proxy/pull/255

PR #255, opened 2025-01-23
Required quorums glitch

### LINKED ISSUE #252: bug: requiredQuorums glitch on holesky

gastonponti:
## Fixes Issue

Closes #252

## Changes proposed

As is stated in the issue, it avoids the quorum 1 check for the range of blocks [2950000, 2960000) for Holesky

## Note to reviewers

Instead of using a flag passed through the cli, I'm checking the svc address, which will change for the v2, but I'm assuming that this issue will be only for the v1 users.
This check avoids cascading a flag in a lot of places just because of this particular bug. With this approach the custom changes live only in the verifier.

This is a new PR required because the original one (#253 ) was pointing to a different branch that after being merged it was closed


bxue-l2 (approved):

samlaf (approved):
LGTM after you fix the lint issues. Also let's make sure not to merge before celo has actually succesfully tested this (to make sure the range is wide enough mainly).

samlaf:
@gastonponti confirmed that the archive sync worked smoothly with this change. Good to merge

PR #264, opened 2025-01-29
chore: Update logging to use `layr-labs/eigensdk-go`

ethenotethan:
## Fixes Issue

Fixes #

## Changes proposed
Updates all logging references to use `eigengo-sdk/logging`. This ensures better compatibility with the mono repo. Also refactored our metrics to no longer be an OP dependency and instead a local one 😄 

More specifically:
- Use same logging software as monorepo (mostly). There are some conflicts with urfave/cli that we currently mitigate by duplicating code. 
- Updated to latest eigenda commit - this required updating binding references as well as the encoder config. 
- Deprecate flags no longer being used associated with op-service logging
- Use custom metrics and not CLI provided by OP

### Key Consideration
The slogger leveraged here dumps out json format logs. This could potentially bork existing loki / observability integrations that our customers use for viewing proxy output. We will need to articulate this very clearly in our release notes to ensure our customers have appropriate visibility.  

## Blocked by
- https://github.com/Layr-Labs/eigenda/pull/1183


README.md
```diff
@@ -269,10 +269,9 @@ To quickly set up monitoring dashboard, add eigenda-proxy metrics endpoint to a
...
-| `--log.level`                               | `INFO`                          | `$EIGENDA_PROXY_LOG_LEVEL`                               | The lowest log level that will be output.                                                                                                                                                                     |
-| `--log.pid`                                 | `false`                         | `$EIGENDA_PROXY_LOG_PID`                                 | Show pid in the log.                                                                                                                                                                                          |
+| `--log.format`                              | `text`                          | `$EIGENDA_PROXY_LOG_FORMAT`                              | The format of the log file. Accepted options are 'json' and 'text' (default: "json")                                                                                                               |
```
samlaf, line 272: I personally prefer text (its both human and machine parsable, unlike json which is only machine readable). Why are you choosing json?

README.md
```diff
@@ -269,10 +269,9 @@ To quickly set up monitoring dashboard, add eigenda-proxy metrics endpoint to a
...
+| `--log.level`                               | `INFO`                          | `$EIGENDA_PROXY_LOG_LEVEL`                               | The lowest log level that will be output. The lowest log level that will be output. Accepted options are "debug", "info", "warn" "error"                                                                                                                                                                     |
+| `log.path`                                  | `""`                              | `$EIGENDA_PROXY_LOG_PATH`                                | Path to file where logs will be written |
```
samlaf, line 274: Feel like we shouldn't have this. logging to file can be done externally, either by piping output, or using tea, or if you're running in a container then just use docker log mechanism. Unix's modular philosophy is the way.

common/logging_cli.go
```diff
@@ -0,0 +1,127 @@
...
+			Action: func(_ *cli.Context, _ bool) error {
+				return fmt.Errorf("flag --%s is deprecated", ColorFlagName)
+			},
+		},
+	}
+}
+
+func ReadLoggerCLIConfig(ctx *cli.Context, flagPrefix string) (*common.LoggerConfig, error) {
+	cfg := common.DefaultLoggerConfig()
+	format := ctx.String(common.PrefixFlag(flagPrefix, FormatFlagName))
```
samlaf, line 101: easier to read if you define flag names like we do for all other CLI flags.

common/logging_cli.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+
+		&cli.BoolFlag{
+			Name:     common.PrefixFlag(flagPrefix, ColorFlagName),
+			Category: category,
+			Usage:    "Color the log output if in terminal mode",
+			EnvVars:  []string{common.PrefixEnvVar(envPrefix, "LOG_COLOR")},
+			Hidden:   true,
+			Action: func(_ *cli.Context, _ bool) error {
+				return fmt.Errorf("flag --%s is deprecated", ColorFlagName)
```
samlaf, line 93: we didn't have this flag before did we? Why mark it deprecated now if we didn't have it before?

common/logging_cli.go
```diff
@@ -0,0 +1,127 @@
...
+			Value:    "info",
+			EnvVars:  []string{common.PrefixEnvVar(envPrefix, "LOG_LEVEL")},
+		},
+		&cli.StringFlag{
+			Name:     common.PrefixFlag(flagPrefix, PathFlagName),
+			Category: category,
+			Usage:    "Path to file where logs will be written",
+			Value:    "",
+			EnvVars:  []string{common.PrefixEnvVar(envPrefix, "LOG_PATH")},
+		},
```
samlaf, line 64: I would get rid of this, as explained in my other comment

common/logging_cli.go
```diff
@@ -0,0 +1,127 @@
...
+		&cli.StringFlag{
+			Name:     common.PrefixFlag(flagPrefix, FormatFlagName),
+			Category: category,
+			Usage:    "The format of the log file. Accepted options are 'json' and 'text'",
+			Value:    "json",
```
samlaf, line 69: text?

cmd/server/entrypoint.go
```diff
@@ -5,26 +5,37 @@ import (
...
 func StartProxySvr(cliCtx *cli.Context) error {
-	log := oplog.NewLogger(oplog.AppOut(cliCtx), oplog.ReadCLIConfig(cliCtx)).New("role", "eigenda_proxy")
-	oplog.SetGlobalLogHandler(log.Handler())
+
+	logCfg, err := common.ReadLoggerCLIConfig(cliCtx, "log")
+	if err != nil {
+		return err
+	}
+
+	log, err := eigenda_common.NewLogger(*logCfg)
```
samlaf, line 27: how much code is this function? Feels like starting a logger is pretty simple code, how about we just duplicate it here and then we don't need to deal with your urfave version issue?

common/logging_cli.go
samlaf, line 1: can we put logging in its own package? I think overall we should try to minimize package names like "common", "utils", etc. which are just catch all names. They can be part of the package path (common/logging), but in my mind we shouldnt have anything in those packages themselves.

metrics/cli.go
```diff
@@ -0,0 +1,75 @@
...
+	return []cli.Flag{
+		&cli.BoolFlag{
+			Name:    EnabledFlagName,
+			Usage:   "Enable the metrics server",
+			EnvVars: common.PrefixEnvVar("METRICS_ENABLED"),
+		},
+		&cli.StringFlag{
+			Name:    ListenAddrFlagName,
+			Usage:   "Metrics listening address",
+			Value:   defaultListenAddr, // TODO: Switch to 127.0.0.1
```
samlaf, line 39: why this todo?

verify/cert.go
```diff
@@ -10,28 +10,29 @@ import (
 
 	"github.com/Layr-Labs/eigenda-proxy/common/consts"
 	"github.com/Layr-Labs/eigenda/api/grpc/disperser"
-	binding "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDAServiceManager"
+	binding "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDACertVerifier"
+	svc_binding "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDAServiceManager"
```
samlaf, line 14:
```suggestion
	edsm_binding "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDAServiceManager"
```

samlaf: do you mind using this convention? its more precise and its what we use everywhere in the middleware contracts repo

README.md
ethenotethan: this is in strict alignment with the eigenda logger CLI which we'll eventually migrate to

README.md
ethenotethan: Again just in alignment with upstream

common/logging_cli.go
ethenotethan: true but this is ported from upstream so we'll end up nuking this at some point anyway

common/logging_cli.go
ethenotethan: we had it before:
```
   --log.color           Color the log output if in terminal mode (default: false) [$EIGENDA_PROXY_LOG_COLOR]
   --log.format value    Format the log output. Supported formats: 'text', 'terminal', 'logfmt', 'json', 'json-pretty', (default: text) [$EIGENDA_PROXY_LOG_FORMAT]
   --log.level value     The lowest log level that will be output (default: INFO) [$EIGENDA_PROXY_LOG_LEVEL]
   --log.pid             Show pid in the log (default: false) [$EIGENDA_PROXY_LOG_PID]
```

common/logging_cli.go
ethenotethan: would prefer if we kept it since this entire artifact is copy pasta from upstream

common/logging_cli.go
ethenotethan: can update - but this is the default for eigenda

litt3 (approved):
lgtm

verify/cert.go
ethenotethan: [20208b9](https://github.com/Layr-Labs/eigenda-proxy/pull/264/commits/20208b9ad29957cbefb39af4d314765ffde423c7)

common/logging_cli.go
samlaf: Does it matter? Very unclear when we'll actually merge the two. And in the meantime we'll have to read these logs.

samlaf: Also this is a breaking change for users of this right? They might have had infra setup to ingest our previous log formats. Were those text?

samlaf (commented):
Just realized this is potentially a breaking change for customers right? See https://github.com/Layr-Labs/eigenda-proxy/pull/264#discussion_r1938330373

common/logging_cli.go
ethenotethan: lemme evaluate - if both `op-log` and `eigensdk-go/logger` use the same slogger under the hood then it should be ok. I flagged this as a risk in the initial PR description.

ethenotethan: yeah output formats change: https://www.diffchecker.com/hFA97XKJ/ 

Let's just make sure to be very loud in our comms about this potential observability risk?

common/logging_cli.go
ethenotethan: [8645e79](https://github.com/Layr-Labs/eigenda-proxy/pull/264/commits/8645e79550d48c8c99d67ea0460a4d1f729018e3)

cmd/server/entrypoint.go
ethenotethan: [8645e79](https://github.com/Layr-Labs/eigenda-proxy/pull/264/commits/8645e79550d48c8c99d67ea0460a4d1f729018e3)

metrics/cli.go
ethenotethan: [34aa9e9](https://github.com/Layr-Labs/eigenda-proxy/pull/264/commits/34aa9e987203d18765c6a91d6c61e412fd1cdb12)

common/logging_cli.go
ethenotethan: on second thought it'd be best to adhere this to the existing abstractions within proxy: [0bb600b](https://github.com/Layr-Labs/eigenda-proxy/pull/264/commits/0bb600b45f55ab7dec75d1caf4daceca1585a955)

common/logging_cli.go
samlaf: Yeah its not that bad. still a golang format key-value pair output, so should be fine.
image

Something looks off though. Some lines use different logging format than others.

samlaf (approved):
LGTM, except for the fact that we seem to be using different loggers in diffeent places. Is that because some of the library code is using a hardcoded eigenda logger?

PR #270, opened 2025-02-01
chore: Update compose with secondary store setup && minor refactors

ethenotethan:
## Fixes Issue

Fixes #

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -46,6 +46,7 @@ services:
       - EIGENDA_PROXY_LOG_LEVEL=debug
       - EIGENDA_PROXY_ADDR=0.0.0.0
       - EIGENDA_PROXY_PORT=4242
+      ## Turn this off to talk to actual eigenda network
       - EIGENDA_PROXY_MEMSTORE_ENABLED=true
       - EIGENDA_PROXY_MEMSTORE_EXPIRATION=45m
       - EIGENDA_PROXY_EIGENDA_CERT_VERIFICATION_DISABLED=true
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,22 +1,81 @@
+## The following is a proxy instance 
+## pointed to redis for storage caching and S3
+## for storage failovers
+
 services:
+  ## Used as secondary read failover target
+  minio:
+    image: minio/minio:latest
+    container_name: minio
+    environment:
+      - MINIO_ROOT_USER=minioadmin
+      - MINIO_ROOT_PASSWORD=minioadmin
+    ports:
+      - "9000:9000"
+      - "9001:9001"
+    command: server /data
+    volumes:
+      - minio_data:/data
+
+  minio-init:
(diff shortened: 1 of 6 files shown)


scripts/create-test-s3-bucket.sh
samlaf, line 1: Don't need to do this in this PR but could we convert our Makefile into a Justfile?
Justfile commands allow embedding shell scripts directly, so we don't need to maintain a whole bunch of separate scripts.
Personally find it much cleaner setup.

scripts/create-test-s3-bucket.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Wait 2 seconds to ensure minio is finished bootstrapping
+# TODO: Update this to do event based polling on minio server directly vs semi-arbitrary timeout
+sleep 2s
+
+# Configure MinIO client (mc)
+echo "Configuring MinIO client..."
+mc alias set local http://minio:9000 minioadmin minioadmin
```
samlaf, line 9: What is mc? I don't have it on my system. Is this an external dependency we're adding? Prob want to document it somewhere. Eventually we'll prob want to have a mise file like op (or go all out and use nix)

docker-compose.yaml
samlaf, line 1: What's the goal for this docker-compose file? Is it just for manual testing?

It's not mentioned anywhere in the README, its not in CI, and don't think we use it for anything explicit right now. Might be good to document it and add it to ci if we want to keep it around?

samlaf (approved):

scripts/create-test-s3-bucket.sh
ethenotethan: oo probably worth noting that this script is only runnable in the `minio-init` service since it references the docker network from within

docker-compose.yaml
ethenotethan: some ideas around this; we coulda add feature testing to ensure lightweight dispersal/retrieval in a more real environment than the one we codify for E2E testing. The idea is you have one script that does a E2E set/get assertion which takes in specific feature flags denoting a specific runtime behavior of proxy which wrap around it; ie:
- `fallback_dest`: [`S3`, `redis`, `nil`] w/ 3 variations
- `cache_dest`: [`S3`, `redis`, `nil`] w/ 3 variations
- `cert_verification`: [`true`, `false`] w 2 variations

```
3 * 3 * 2 = 18 permutations; but less since `fallback_dest` and `cache_dest` can only be sampled with replacement for `nil` so it'd be - ( 2 * 2) or:

18 - 4 = 14 test case permutations
```

Think this would be good to have since it also provides a more infra level overview of what wiring proxy with dependencies look like. This would definitely help devx when people have to spin-up additional infra and wire it with proxy.

scripts/create-test-s3-bucket.sh
ethenotethan: yeah I'm down to use it - lets file an issue

PR #279, opened 2025-02-06
Create codeql-scanning.yml

anupsv:
## Fixes Issue
Adding codeql config in yaml file than github UI enabled.

### DIFF
--- a/.github/workflows/codeql-scanning.yml
+++ b/.github/workflows/codeql-scanning.yml
@@ -24,20 +24,20 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-scanning.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      security-events: write
+      pull-requests: read
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2
```
samlaf, line 27: think if we want to use commits like this instead of versions we should at least comment the version being used. otherwise it's hard to find and figure out which commit to update to next if we ever want to upgrade.

samlaf (approved):
Very sexy, also not sure what's happening with that config supposedly missing
image

.github/workflows/codeql-scanning.yml
anupsv: sure. will tag

PR #285, opened 2025-02-10
chore: bump default confirmation-depth 0->8 blocks

samlaf:
This is a very important change. Reorgs of 1 block are quite common and can cause a da cert to be submitted to the rollup inbox with a wrong confirmationBlock! It's important to wait a few blocks for the confirmBatch to have landed onchain before submitting the dacert to the batcher inbox.

Also added a troubleshooting guide and added info as to how to troubleshoot the `batch-hash-mismatch` error.

Should we set default to finalized instead of 8? That might greatly impact the throughput of rollups, and iirc the longest reorg that ever happened on pos was on goerli and was something like 7 or 8 blocks long.

## Fixes Issue

Fixes #


docs/troubleshooting_v1.md
```diff
@@ -0,0 +1,51 @@
...
+RLP_ENCODED_CERT=${OP_ALTDA_COMMITMENT:6}
+cast --from-rlp $RLP_ENCODED_CERT | jq
+```
+Unfortunately `cast --from-rlp` doesn't currently support schemas, so you will have to compare the arrays-of-hex-strings output to the [BlobInfo](https://github.com/Layr-Labs/eigenda/blob/f305e046ae3e611e19c15e134571cc2ec83062b4/api/proto/disperser/disperser.proto#L178) schema (note that [CertV1 === BlobInfo](https://github.com/Layr-Labs/eigenda-proxy/blob/97b71bda6e2136a6c0d71be2f7e9642dc840fefc/verify/certificate.go#L31)).
+
+And here's a full script to compare both (replace `REQUEST_ID`, `OP_ALTDA_COMMITMENT`, and `DISPERSER_ENDPOINT` with your values):
+```bash
+DISPERSER_ENDPOINT=disperser-holesky.eigenda.xyz:443
+# Get the request_id from your batcher's proxy logs
+REQUEST_ID=05cb90531098964cc73f0e6b782f52f3b4387e3aed3ef6b89cd9a3c118dd781e-313733393032373134323137353330363738362f302f33332f312f33332fe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```
bxue-l2, line 33: Do we actually log it?

docs/troubleshooting_v1.md
```diff
@@ -0,0 +1,51 @@
...
+RLP_ENCODED_CERT=${OP_ALTDA_COMMITMENT:6}
+cast --from-rlp $RLP_ENCODED_CERT | jq
+```
+Unfortunately `cast --from-rlp` doesn't currently support schemas, so you will have to compare the arrays-of-hex-strings output to the [BlobInfo](https://github.com/Layr-Labs/eigenda/blob/f305e046ae3e611e19c15e134571cc2ec83062b4/api/proto/disperser/disperser.proto#L178) schema (note that [CertV1 === BlobInfo](https://github.com/Layr-Labs/eigenda-proxy/blob/97b71bda6e2136a6c0d71be2f7e9642dc840fefc/verify/certificate.go#L31)).
```
bxue-l2, line 27: shall we create a tool for parsing the Cert?

bxue-l2 (approved):

docs/troubleshooting_v1.md
bxue-l2: I remember Emiliano has the tool, but it isn't difficult to write

docs/troubleshooting_v1.md
samlaf: probably... how else would the rollups get it?

bxue-l2: They might have done some local change

PR #286, opened 2025-02-11
ci: add test to make sure env vars in .env.example.holesky are valid

samlaf:
## Fixes Issue

Fixes #

## Changes proposed

Mantle has run into issues because one of the flags in our .env.holesky.example file was incorrect.
> In the configuration file example and the readme file, the variable name has always been EIGENDA_PROXY_EIGENDA_ETH_CONFIRMATION_DEPTH.
> https://github.com/Layr-Labs/eigenda-proxy/blob/97b71bda6e2136a6c0d71be2f7e9642dc840fefc/.env.example.holesky
> https://github.com/Layr-Labs/eigenda-proxy/blob/main/README.md

This PR removes the flags in the README which are just very prone to be outdated and induce errors like this, and also adds a ci check to make sure the flags defined in .env.example.holesky are valid. This test is simple and will only catch if that file still contains deprecated flags (which cause the proxy to crash on startup); it won't be able to catch the subtle case above where the flag is not an old deprecated flag but instead just an unused env var.


scripts/test-proxy-startup-with-env-vars.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e  # Exit on any error
+
+##### This script is meant to be run in ci #####
+# It tests that the env vars defined in .env.example.holesky are correct.
+# It starts the eigenda-proxy with those env vars, waits 5 seconds, and then kills the proxy.
+# If any deprecated flags are still being used in .env.example.holesky, the script will fail.
```
ethenotethan, line 7: shouldn't we also test set/get logic as well? or at a minimum ping the health endpoint to at least ensure the server instance successfully finished bootstrapping?

.github/workflows/build.yml
```diff
@@ -17,8 +17,23 @@ jobs:
       with:
         go-version: 1.22
 
-    - name: Build App
+    - name: Check that build succeeds
       run: make eigenda-proxy
+  
+  flags:
```
ethenotethan, line 23: knit: `holesky-flags`

bxue-l2 (dismissed):

scripts/test-proxy-startup-with-env-vars.sh
samlaf: [9204b0d](https://github.com/Layr-Labs/eigenda-proxy/pull/286/commits/9204b0d5d7dc2ae8e62d8cbb4b73c083423ed3c2)

ethenotethan (approved):
LGTM

PR #289, opened 2025-02-13
feat: add memstore config to return failover errors on puts

samlaf:
This is useful for testing failover in batcher.

Right now a new proxy would need to be spun up with this config turned on. We should consider adding an admin rpc rpi to turn these on/off dynamically.

## Fixes Issue

Fixes #


store/generated_key/memstore/cli.go
```diff
@@ -67,18 +76,25 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 		},
 		&cli.DurationFlag{
 			Name:     PutLatencyFlagName,
-			Usage:    "Artificial latency added for memstore backend to mimic EigenDA's dispersal latency.",
+			Usage:    "(FOR TESTING) Artificial latency added for memstore backend to mimic EigenDA's dispersal latency.",
```
ethenotethan, line 79: nit - isn't this entire construction used for testing? maybe less redundant to put disclosure in category description instead

ethenotethan (approved):
LGTM - just one nit

store/generated_key/memstore/cli.go
samlaf: [55b4ec1](https://github.com/Layr-Labs/eigenda-proxy/pull/289/commits/55b4ec1364979030da269e0ebd9f086e1b88e8b7)

litt3 (approved):
LGTM

PR #295, opened 2025-02-18
ci: cleanup into a single per-pr.yml gha

samlaf:
Much needed cleanup of CI pipelines. We now have 2 main pipelines:
- one that runs on every PR: runs tests, lint, builds
- one that runs on every release: pushes image to GHCR

Only downside of this cleanup is that we can't have separate README badges for the different jobs (GHA only allows one badge for the entire workflow). Personally I prefer having a cleaner CI than extra precision of badges on the repo.

Even considering switching our CI to use https://docs.dagger.io/ at some point.

## Fixes Issue

Fixes #

## Note to reviewers

We will have to update the required workflows in the github repo settings. The old workflows that have been deleted (but are still required) will never pass. We should prob update them right before merging this PR. I can do that once this PR is approved.


litt3 (dismissed):
LGTM

ethenotethan (dismissed):
LGTM

ethenotethan (approved):
LGTM

PR #296, opened 2025-02-18
chore(V2): Refactor dependency injection && config mgmt

ethenotethan:
## Changes proposed
- Move V2 flags into a separate package vs overloading into `eigendaflags` and managing a janky intersection
- Refactor config mgmt. The highest order or "Application Config" lived in /server despite server being one of two server instances that live on the application. E.g the app could be thought of the (proxy_server, metrics_server). Created a `config` package to store this higher order object which forced some refactors given abstraction leaks. 
- Move `load_store.go` which was a stateless set of dependency injection functions into a `StoreLoader` construction within the storage package. 
- Remove deprecated flag categories to reduce cognitive burden of flag overhead.


config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	clients_v2 "github.com/Layr-Labs/eigenda/api/clients/v2"
+)
+
+// AppConfig ... Highest order config. Stores
+// all relevant fields necessary for running both proxy
+// & metrics servers.
+type AppConfig struct {
+	EigenDAConfig ProxyConfig
+	MetricsCfg    metrics.Config
+}
```
samlaf, line 26: wait so we actually run 2 servers on 2 separate ports?
Think we should get rid of this, and just load the metrics handler on the same port that the normal REST routes are. Don't think we need the extra complexity.

config/config.go
```diff
@@ -0,0 +1,145 @@
...
+
+	EdaV2ClientConfig common.V2ClientConfig
+
+	MemstoreConfig memstore.Config
+	StorageConfig  store.Config
+	PutRetries     uint
+
+	MemstoreEnabled bool
+
+	EigenDAV2Enabled bool
```
samlaf, line 58: add comments saying that eigendav2 can only be enabled when memstore is not enabled.

config/config.go
```diff
@@ -0,0 +1,145 @@
...
+		cfg.EdaV1ClientConfig.EthRpcUrl = "http://0.0.0.0:666"
+	}
+
+	if !cfg.MemstoreEnabled {
+		if cfg.EdaV1ClientConfig.SvcManagerAddr == "" {
+			return fmt.Errorf("service manager address is required for communication with EigenDA")
+		}
+		if cfg.EdaV1ClientConfig.EthRpcUrl == "" {
+			return fmt.Errorf("eth prc url is required for communication with EigenDA")
+		}
```
samlaf, line 111: move these above, you already have another `if !cfg.MemstoreEnabled` cobnditional

config/eigendaflags/v2/cli.go
```diff
@@ -0,0 +1,238 @@
...
+	// dependencies migrate to using EigenDA V2 network
+	V2EnabledFlagName       = withFlagPrefix("enabled")
+	DisperserFlagName       = withFlagPrefix("disperser-rpc")
+	DisableTLSFlagName      = withFlagPrefix("disable-tls")
+	CustomQuorumIDsFlagName = withFlagPrefix("custom-quorum-ids")
+	BlobStatusPollInterval  = withFlagPrefix("blob-status-poll-interval")
+	// TODO: Determine whether we should change this to something like PaymentPrivateKeyHex
+	PutBlobEncodingVersionFlagName = withFlagPrefix("put-blob-encoding-version")
+	// TODO: Consider renaming this to FFT mode or something pseudo-similar
+	DisablePointVerificationModeFlagName = withFlagPrefix("disable-point-verification-mode")
```
samlaf, line 24: yes we should use the same name that @litt3 's client is using.

config/eigendaflags/v2/cli.go
```diff
@@ -0,0 +1,238 @@
...
+
+func readPayloadClientConfig(ctx *cli.Context) v2_clients.PayloadClientConfig {
+	noPolynomial := ctx.Bool(DisablePointVerificationModeFlagName)
+	polyMode := codecs.PolynomialFormCoeff
+
+	// if point verification mode is disabled then blob is treated as evaluations and
+	// not FFT'd before dispersal
+	if noPolynomial {
+		polyMode = codecs.PolynomialFormEval
+	}
```
samlaf, line 193: `polyMode` is not very precise here. Let's stick to `PayloadPolynomialForm`

store/loader.go
```diff
@@ -0,0 +1,224 @@
...
+		MaxGRPCMessageSize: d.v2ClientCfg.PutRetries,
+		Sockets:            relayURLs,
+	}
+
+	retriever, err := clients_v2.BuildRelayPayloadRetriever(d.log, d.v2ClientCfg.RetrievalConfig, &relayCfg, g1Points)
+	if err != nil {
+		return nil, err
+	}
+
+	disperser, err := clients_v2.BuildPayloadDisperser(d.log, d.v2ClientCfg.PayloadClientCfg, &d.v2ClientCfg.DisperserClientCfg, &gethCfg, nil, nil)
```
samlaf, line 113: @litt3 I thought we had embedded the disperser config into the payloadDisperserConfig?
Doesn't it feel a bit weird to have to pass 2 configs here? Feels like the abstraction is leaking

store/loader.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/Layr-Labs/eigenda/common/geth"
+	eigenda_eth "github.com/Layr-Labs/eigenda/core/eth"
+	"github.com/Layr-Labs/eigenda/encoding/kzg"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	geth_common "github.com/ethereum/go-ethereum/common"
+)
+
+// StoreLoader centralizes dependency initialization
+// It ensures proper typing and avoids redundant logic scattered across functions.
+type StoreLoader struct {
```
samlaf, line 30: does this actually "load" anything? it more so feels like a builder to me, but maybe I'm misunderstanding how its used.

config/config.go
ethenotethan: it's pretty common to run metrics on another port. imo it makes more sense to do this since you can control ingress per port which allows an operator to delineate networking policies.

config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	MemstoreEnabled bool
+
+	EigenDAV2Enabled bool
+}
+
+type V2ClientConfig struct {
+	DisperserClientCfg clients_v2.DisperserClientConfig
+	PayloadClientCfg   clients_v2.PayloadDisperserConfig
+	RetrievalConfig    clients_v2.RelayPayloadRetrieverConfig
+}
```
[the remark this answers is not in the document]
ethenotethan: agree that we should merge into one

config/config.go
ethenotethan: technically not true with the current implementation - as of now memstore can only be enabled for V1

config/config.go
```diff
@@ -0,0 +1,145 @@
...
+
+func ReadCLIConfig(ctx *cli.Context) AppConfig {
+	return AppConfig{
+		EigenDAConfig: ReadProxyConfig(ctx),
+		MetricsCfg:    metrics.ReadConfig(ctx),
+	}
+}
+
+// ProxyConfig ... Higher order config which bundles all configs for instrumenting
```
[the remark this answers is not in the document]
ethenotethan: hmm maybe orchestrating is better fit

config/config.go
ethenotethan: [20944e3](https://github.com/Layr-Labs/eigenda-proxy/pull/296/commits/20944e32884209950be9ff3c4ef017945ca3fc94)

config/config.go
ethenotethan: Added comment to capture: [20944e3](https://github.com/Layr-Labs/eigenda-proxy/pull/296/commits/20944e32884209950be9ff3c4ef017945ca3fc94)

config/config.go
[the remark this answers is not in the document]
ethenotethan: [20944e3](https://github.com/Layr-Labs/eigenda-proxy/pull/296/commits/20944e32884209950be9ff3c4ef017945ca3fc94)

config/config.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// V2 dispersal/retrieval enabled
+	if cfg.EigenDAV2Enabled {
+
+		if cfg.EdaV1ClientConfig.SvcManagerAddr == "" {
+			return fmt.Errorf("service manager address is required for interacting with EigenDA V2")
+		}
+
+		if cfg.EdaV1ClientConfig.EthRpcUrl == "" {
+			return fmt.Errorf("eth rpc is required for interacting with EigenDA V2")
```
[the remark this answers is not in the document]
ethenotethan: good catch: [20944e3](https://github.com/Layr-Labs/eigenda-proxy/pull/296/commits/20944e32884209950be9ff3c4ef017945ca3fc94)

config/eigendaflags/v2/cli.go
```diff
@@ -0,0 +1,238 @@
+package eigendaflags
```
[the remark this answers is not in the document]
ethenotethan: this wouldn't adhere with how we're doing so in monorepo: https://github.com/Layr-Labs/eigenda/blob/master/api/clients/v2/accountant.go#L1

config/eigendaflags/v2/cli.go
ethenotethan: [20944e3](https://github.com/Layr-Labs/eigenda-proxy/pull/296/commits/20944e32884209950be9ff3c4ef017945ca3fc94)

store/loader.go
```diff
@@ -0,0 +1,224 @@
...
+				StatusQueryTimeout:   d.v1EdaClientCfg.StatusQueryTimeout,
+				PutRetries:           putRetries,
+			},
+		)
+	}
+
+	return eigenDA, nil
+}
+
+func (d *StoreLoader) LoadManager(ctx context.Context, putRetries uint) (IManager, error) {
```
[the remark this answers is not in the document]
ethenotethan: [df44d5f](https://github.com/Layr-Labs/eigenda-proxy/pull/296/commits/df44d5f3bdcbdd76a5eaf2d5014cd4134bf88573)

store/loader.go
ethenotethan: [df44d5f](https://github.com/Layr-Labs/eigenda-proxy/pull/296/commits/df44d5f3bdcbdd76a5eaf2d5014cd4134bf88573)

ethenotethan: agree

config/config.go
samlaf: very true. makes sense

config/config.go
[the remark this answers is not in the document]
samlaf: also see https://github.com/Layr-Labs/eigenda-proxy/pull/296#discussion_r1960089685

samlaf (approved):

config/config.go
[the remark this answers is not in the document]
litt3: > Feels a bit weird that we ned 2 separate configs for dispersal, but only a single one for retrieval.

In all cases, we need two configs: one for the "payload" level component, and one for the lower-level "client":

- Dispersal requires `PayloadDisperserConfig`, and `DisperserClientConfig`
- Relay retrieval requires `RelayPayloadRetrieverConfig`, and `RelayClientConfig`
- Validator node retrieval requires `ValidatorPayloadRetrieverConfig`, and a bunch of stray config values (which really ought to be collected into a config file)

It's not clear to me why the `RelayClientConfig` is missing from the `V2ClientConfig` here?

> agree that we should merge into one

Regarding this, the correct solution isn't immediately obvious to me (I actually lean towards *not* combining).

I think it would only make sense to combine the configs *IF* we also combine the clients themselves, i.e. combine `RelayClient` with `RelayPayloadRetriever`, `RetrievalClient` with `ValidatorPayloadRetriever`, and `DisperserClient` with `PayloadDisperser`. But should we actually combine these? The current separation of concern kinda seems correct to me, where there is a user-facing object which deals in the realm of payloads, and that user facing object wraps around an underlying client, which knows how to deal with blobs and interact with eigenDA components.

Regardless of the decision we reach, one thing that could certainly be improved is naming consistency for the underlying clients. The payload level components are already consistently named, but I think better names for the underlying clients would be something like `DispersingClient`, `RelayRetrieverClient`, `ValidatorRetrieverClient`. Or alternatively, these could be named using the term `blob`, in contrast to the term `payload` in the user-facing objects.

Issue #294, opened 2025-02-16
Create a golang client for the memconfig routes


samlaf:
https://github.com/Layr-Labs/eigenda-proxy/pull/293 introduced HTTP routes to modify the memstore config.
My initial use case was to call those in bash tests, but we should prob create golang clients in case anyone wants to modify them in golang tests.

PR #301, opened 2025-02-19
feat: Memconfig client

ethenotethan:
## Fixes Issue

Fixes https://github.com/Layr-Labs/eigenda-proxy/issues/294

## Changes proposed
- Add a structured memconfig client for interacting with memconfig handlers
- Introduce a new integration test to ensure the client can update the server as expected


e2e/resources/kzg/SRSTables/dimE512.coset1
samlaf, line 1: what is this? Does it get auto generated by something?

client/memconfig.go
samlaf, line 1: Can we create separate packages? Like
`clients/memconfig/client.go`
`clients/standard/clieng.go`

I'm always afraid when we bunch things that don't share anything in the same namespace.

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// MemConfig ... contains properties that are used to configure the MemStore's behavior.
+// this is copied directly from /store/generated_key/memstore/memconfig. 
+// importing the struct isn't possible since it'd create cyclic dependency loop 
+// with core proxy's go.mod
+type MemConfig struct {
+	MaxBlobSizeBytes uint64
+	BlobExpiration   time.Duration
+	PutLatency time.Duration
```
samlaf, line 23: dang this is really unfortunate. What we could do is have core import from here instead? Kind of ugly though...

At some point we might need to refactor our stuff and create a `shared` workspace that has these kinds of structs that are shared between proxy and clients.

litt3 (approved):

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+	BlobExpiration          string
+	PutLatency              string
+	GetLatency              string
+	PutReturnsFailoverError bool
+}
+
+// IntoMemConfig ... converts an intermediary config into a memconfig 
+// with structured type definitions
+func (cfg *intermediaryCfg) IntoMemConfig() (*MemConfig, error) {
+	getLatency, err := time.ParseDuration(cfg.GetLatency)
```
samlaf, line 55: Can't we just implement UnmarshalJSON directly on MemConfig the same way we implemented the custom MarshalJSON?

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+		httpClient: http.DefaultClient,
+	}
+
+	return scc
+}
+
+// decodeResponseToMemCfg ... converts http response to structured MemConfig
+func decodeResponseToMemCfg(resp *http.Response) (*MemConfig, error) {
+	var cfg intermediaryCfg
+	if err := json.NewDecoder(resp.Body).Decode(&cfg); err != nil {
```
samlaf, line 104: if we implement the custom UnmarshalJSON on memconfig directly we wouldn't need this afaiu

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, err
+	}
+	return cfg.IntoMemConfig()
+}
+// GetConfig retrieves the current configuration.
+func (c *MemCfgClient) GetConfig(ctx context.Context) (*MemConfig, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.cfg.URL, &bytes.Buffer{})
+	if err != nil {
+		return nil, err
```
samlaf, line 113: wrap with some context

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+// GetConfig retrieves the current configuration.
+func (c *MemCfgClient) GetConfig(ctx context.Context) (*MemConfig, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.cfg.URL, &bytes.Buffer{})
+	if err != nil {
+		return nil, err 
+	}
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, err
```
samlaf, line 118: wrap

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+	}, nil
+}
+
+
+// MemCfgClient implements a standard client for the eigenda-proxy
+// that can be used for updating a memstore configuration in real-time
+// this is useful for API driven tests in protocol forks that leverage
+// custom integrations with EigenDA
+type MemCfgClient struct {
+	cfg        *Config
```
samlaf, line 85: Prefer not to share Configs between different clients, because they might change independently to require different things. Another reason for putting the two clients in separate packages.

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPatch, c.cfg.URL, bytes.NewBuffer(body))
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, err
```
samlaf, line 144: wrap

client/memconfig.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, fmt.Errorf("failed to read config. expected status code 200, got: %d", resp.StatusCode)
+	}
+
+	return decodeResponseToMemCfg(resp)
+}
+
+// UpdateConfig updates the configuration using the new MemConfig instance
+// by sending a PATCH request to the proxy server
+func (c *MemCfgClient) UpdateConfig(ctx context.Context, update *MemConfig) (*MemConfig, error) {
+	body, err := update.MarshalJSON()
```
samlaf, line 132: This is a bit dangerous the way its implemented, because it doesn't really allow you to PATCH. You should prob use `omitempty` tag in the struct. Otherwise any field that is not initialized will be given zero value and happily overwrite the server's config with those values. We want to make sure zero fields are not included in the marshalled json (except for boolean values...!)

e2e/resources/kzg/SRSTables/dimE512.coset1
ethenotethan: yeah its a cache table used by the kzg provers. Removing it since it shouldn't be committed

client/memconfig.go
ethenotethan: we'd still need to cast into `time.Duration` right?

client/memconfig.go
ethenotethan: hmmm how would fields be omitted if we're passing in a structured `MemConfig` to update? This would make more sense if we providing an arbitrary key/value interface which we're not. This client functionality should work fine for the use case of Arbitrum E2E testing and shouldn't result in fields being discarded.

client/memconfig.go
ethenotethan: [b9825c4](https://github.com/Layr-Labs/eigenda-proxy/pull/301/commits/b9825c4e2bd7b72fc53b57584abd921d2940a36a)

client/memconfig.go
ethenotethan: [b9825c4](https://github.com/Layr-Labs/eigenda-proxy/pull/301/commits/b9825c4e2bd7b72fc53b57584abd921d2940a36a)

client/memconfig.go
ethenotethan: [b9825c4](https://github.com/Layr-Labs/eigenda-proxy/pull/301/commits/b9825c4e2bd7b72fc53b57584abd921d2940a36a)

client/memconfig.go
ethenotethan: [b9825c4](https://github.com/Layr-Labs/eigenda-proxy/pull/301/commits/b9825c4e2bd7b72fc53b57584abd921d2940a36a)

client/memconfig.go
ethenotethan: [b9825c4](https://github.com/Layr-Labs/eigenda-proxy/pull/301/commits/b9825c4e2bd7b72fc53b57584abd921d2940a36a)

client/memconfig.go
ethenotethan: [b9825c4](https://github.com/Layr-Labs/eigenda-proxy/pull/301/commits/b9825c4e2bd7b72fc53b57584abd921d2940a36a)

go.mod
```diff
@@ -6,7 +6,8 @@ toolchain go1.22.7
 
 require (
 	github.com/Layr-Labs/eigenda v0.8.6-rc.0.0.20250130173836-0d293cc03198
-	github.com/Layr-Labs/eigenda-proxy/client v0.0.0-00010101000000-000000000000
+	github.com/Layr-Labs/eigenda-proxy/client v1.0.0
```
samlaf, line 9: do we still need this?

client/memconfig.go
samlaf: yes but the point is that this intermediaryCfg would only be defined inside the UnmarshalJSON, and hidden as an implementation detail. Whereas the way its currently structured it feels like it's something important, but it really doesn't outside of unmarshalling.

client/memconfig.go
samlaf: The problem is that because it's a PATCH, someone could pass in
```
{
	PutReturnsFailoverError: true
}
```
thinking they'll just update this single value, but actually what this will do is set all the other fields to their zero value.

This is why I'm suggesting instead of this method taking in a *MemConfig, it instead takes a [ConfigUpdate](https://github.com/Layr-Labs/eigenda-proxy/blob/d7b1642956de39bfc62b1bfe3a4467fe4e67bc9a/store/generated_key/memstore/memconfig/http_handlers.go#L14) like I used in the proxy side.

samlaf: There are 2 other options that you might prefer:
1. we forget that I implemented this as a PATCH, and we add a big disclaimer in the comment to this method saying that this method is actually implemented as a PUT, where the EXACT MemConfig passed as input will be set on the proxy. We then suggest that a client gets the current MemConfig first, and then make modifications to that before calling UpdateConfig.
2. We define individual setters for each field which will create a patch with only that specific field. This might be cleaner actually.. with the downside that if you need to update multiple fields at once you'll need to make N calls.

go.mod
ethenotethan: nope

client/memconfig.go
ethenotethan: would prefer to just keep it as is. currently we only plan to use this in the arbitrum integration for testing failover. reading the entire config to update a specific field should be ok for the existing use case.

client/memconfig.go
ethenotethan: eh I don't see any great benefits with this approach over the existing implementation

go.mod
ethenotethan: [acfe191](https://github.com/Layr-Labs/eigenda-proxy/pull/301/commits/acfe191eaec70b2207edb99e34b13e628f8cbb8d)

client/memconfig.go
samlaf: Fine with me, although I PROMISE you that this will cause a bug at some point.
Can you at least add the disclaimer comment I suggested in option 1 above then?

samlaf (approved):
LGTM, but please just add the comment disclaimer I suggested in option 1 [here](https://github.com/Layr-Labs/eigenda-proxy/pull/301#discussion_r1962351333) before merging.

clients/memconfig_client/client.go
```diff
@@ -0,0 +1,159 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to read config. expected status code 200, got: %d", resp.StatusCode)
+	}
+
+	return decodeResponseToMemCfg(resp)
+}
+
+// UpdateConfig updates the configuration using the new MemConfig instance
+// Despite the API using a PATH method, this function treats the "update" config
```
samlaf, line 134:
```suggestion
// Despite the API using a PATCH method, this function treats the "update" config
```

samlaf (approved):

PR #308, opened 2025-02-23
chore(memstore): hash entire cert + remove cert verification

ethenotethan:
## Fixes Issue

Fixes #


samlaf (commented):
Good cleanup thanks :)
2 other small refactors that would help readability of Put route:
1. refactor all the randomCertGeneration stuff into a function (will make Put route super small and easy to read)
2. move the mu.Lock down to where store is used (I don't think its needed while generating the random cert right?)

Issue #236, opened 2025-01-10
Create reproducible dev env


samlaf:
## Issue
We don't have a reproducible dev env right now. Some of our make commands implicitly depend on some binaries being installed on the host environment. `make go-gen-mocks` is one example dependency that caused some issues in a PR, where the repo was using https://github.com/golang/mock as a dependency but I had mockgen `go install`ed from https://github.com/uber-go/mock. Hence running `make go-gen-mocks` created types that were incompatible with the tests. This is one example, we probably have more.

See https://github.com/Layr-Labs/eigenda-proxy/pull/226#discussion_r1908279043 for context.

## Solution

We can use nix, or mise, or whatever. Have a preference against devcontainers because I don't like working with them personally.

WIP list of dependencies that we need to track:
- mockgen (`make go-gen-mocks` is not reproducible right now, see issue section at the top)
- cast in the README command
- golangci-lint (we use one from github action in ci)

samlaf:
https://www.youtube.com/watch?v=HcNXlC1bjvI
This talk shows how to build an awesome nix environment for go development with some cool bells and whistles (direnv, pre-commit-hooks, etc).
One thing to note is that it uses flake-utils which I think (?) is not the favored approach anymore. Instead people seem to use the simpler https://github.com/nix-systems/default

PR #314, opened 2025-02-26
chore: fix readme with correct cast cmd

shrimalmadhur:
## Fixes Issue

Fixes #

## Changes proposed

- It fixes the cast command to create wallet `-j` is not a short for `--json`. so explicitly specifying `--json` since we expect it to be `json` in subsequent commands 

### Screenshots (Optional)
Screenshot 2025-02-26 at 11 04 04 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ $ make
 # github.com/Layr-Labs/eigenda-proxy/cmd/server
 
 ## Setup new keypair for EigenDA authentication
-$ cast wallet new -j > keypair.json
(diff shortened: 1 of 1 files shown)


samlaf (approved):
Looks fine to me. not sure if `-j` was an old flag, but checked with latest foundry version and `--json` is def correct. We should pin our versions though.... see https://github.com/Layr-Labs/eigenda-proxy/issues/236

Note: CI is broken because holesky is down...

PR #260, opened 2025-01-24
feat: EigenDA V2 Support

ethenotethan:
## Changes proposed
Introduces a working MVP for integration with EigenDA V2. 

## Known Issues
- https://github.com/Layr-Labs/eigenda-proxy/issues/278
- https://github.com/Layr-Labs/eigenda-proxy/issues/274
- https://github.com/Layr-Labs/eigenda-proxy/issues/273 
- https://github.com/Layr-Labs/eigenda-proxy/issues/272

PR #321, opened 2025-03-03
feat!: add eigenda v2 support

litt3:
- This PR is for basic V2 support
- This new PR was created to get a clean slate, old PR was [here](https://github.com/Layr-Labs/eigenda-proxy/pull/260)

TODO in followup PRs:
- config/package refactor/improvements (see https://github.com/Layr-Labs/eigenda-proxy/pull/321#discussion_r1979874780)

## Release Considerations

- This PR removes the `wait-for-finalization` flag
- This constitutes a [breaking change](https://github.com/Layr-Labs/eigenda-proxy/pull/321#discussion_r1979991249): therefore, a new major version release will be required after this PR merges


common/secret_config.go
samlaf, line 1: Wdyt about moving all configs to an `/etc` top-level dir? Like in https://github.com/sagikazarmark/modern-go-application/tree/main/etc
I've seen zksync does the same, and I kind of like it. Generally against directories like `common`, `utils`, `core`, etc that just become a miscellaneous throw-away-stuff-i-dont-want-to-care-to-properly-name place.

common/secret_config.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"fmt"
+)
+
+// SecretConfigV2 contains sensitive config data that must be protected from leakage
+type SecretConfigV2 struct {
+	// SignerPaymentKey is the hex representation of the private payment key, that pays for payload dispersal
+	SignerPaymentKey string
+}
```
samlaf, line 11: What about adding URLs in here? We also don't want to leak those as they often contain private api keys

common/secret_config.go
samlaf: Wait just realized we have a top level `config/` dir. Why is this file and client_config in `common/`?

config/eigendaflags/v2/cli.go
```diff
@@ -0,0 +1,239 @@
...
+		ServiceManagerAddress:           ctx.String(SvcManagerAddrFlagName),
+		DisperserClientCfg:              *disperserConfig,
+		PayloadDisperserCfg:             readPayloadDisperserCfg(ctx),
+		RelayPayloadRetrieverCfg:        readRetrievalConfig(ctx),
+		PutRetries:                      ctx.Uint(PutRetriesFlagName),
+		BlockNumberPollIntervalDuration: ctx.Duration(BlockNumberPollIntervalFlagName),
+	}, nil
+}
+
+func ReadSecretConfigV2(ctx *cli.Context) common.SecretConfigV2 {
```
samlaf, line 186: perhaps time to rename `config/eigendaflags` to just `config/eigenda` and move the configs to there?

cmd/server/entrypoint.go
```diff
@@ -40,17 +48,45 @@ func StartProxySvr(cliCtx *cli.Context) error {
...
+		cfg.EigenDAConfig.EdaClientConfigV2,
+		secretConfig,
+		memConfig,
+		log,
+		m,
+	).BuildManager(
+		ctx,
+		cfg.EigenDAConfig.EdaClientConfigV2.PutRetries,
+		cfg.EigenDAConfig.MaxBlobSizeBytes,
+		cfg.EigenDAConfig.EigenDACertVerifierAddress)
```
samlaf, line 84: does the storeBuilder only have a single `BuildManager` function? maybe rename to just `Build`? Or the more explicit `BuildStoreManager` if there are other things. Otherwise `store.NewBuilder().BuildManager` feels like I'm not sure what exactly this is building.

Also why are we passing extra args to `BuildManager`? Shouldnt those be passed to the builder instead?

cmd/server/entrypoint.go
```diff
@@ -91,19 +127,31 @@ func StartProxySvr(cliCtx *cli.Context) error {
...
-	if cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex != "" {
-		cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "*****" // marshaling defined in client config
+	cfg, err := config.ReadCLIConfig(cliCtx)
+	if err != nil {
+		return fmt.Errorf("read cli config: %w", err)
 	}
-	if cfg.EigenDAConfig.EdaClientConfig.EthRpcUrl != "" {
-		cfg.EigenDAConfig.EdaClientConfig.EthRpcUrl = "*****" // hiding as RPC providers typically use sensitive API keys within
+	if cfg.EigenDAConfig.EdaClientConfigV1.SignerPrivateKeyHex != "" {
+		cfg.EigenDAConfig.EdaClientConfigV1.SignerPrivateKeyHex = redacted // marshaling defined in client config
```
samlaf, line 138: guessgin you want to get rid of this at some point by splitting all configs into config/secret?

cmd/server/main.go
```diff
@@ -26,7 +26,7 @@ func main() {
 	oplog.SetupDefaults()
 
 	app := cli.NewApp()
-	app.Flags = cliapp.ProtectFlags(flags.Flags)
+	app.Flags = cliapp.ProtectFlags(config.Flags)
```
samlaf, line 29: why this change? I typicall prefer to keep config and flags in separate packages

commitments/eigenda.go
```diff
@@ -1,39 +1,40 @@
 package commitments
 
-type CertEncodingCommitment byte
+type EigenDACommitmentType byte
```
samlaf, line 3:
```suggestion
type EigenDACertVersion byte
```

commitments/eigenda.go
```diff
@@ -1,39 +1,40 @@
...
 }
 
-type CertCommitmentV0 []byte
-
-// NewV0CertCommitment creates a new commitment from the given input.
-func NewV0CertCommitment(input []byte) CertCommitmentV0 {
-	return CertCommitmentV0(input)
+type EigenDACommitment struct {
+	prefix EigenDACommitmentType
+	b      []byte
```
samlaf, line 21: Think we wanted to call this `EigenDAVersionedCert` in https://app.excalidraw.com/s/1XPZRMVbRNH/26HgEWKo0sT ?

commitments/mode.go
```diff
@@ -31,19 +31,23 @@ func StringToCommitmentMode(s string) (CommitmentMode, error) {
...
 	case OptimismGeneric:
-		certCommit := NewV0CertCommitment(b).Encode()
+		certCommit := NewEigenDACommitment(bytes, commitmentType).Encode()
 		svcCommit := EigenDASvcCommitment(certCommit).Encode()
 		altDACommit := NewGenericCommitment(svcCommit).Encode()
 		return altDACommit, nil
 
-	case Standard:
-		return NewV0CertCommitment(b).Encode(), nil
+	case Standard: // (i.e, Arbitrum)
```
samlaf, line 49: Think we'll need to stop being arbitrum specific for the standard commitment mode, especially now that AVSs want to also start using that route. I'm guessing other rollup stacks would as well. So Standard is basically everyone but OP.

common/client_config.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	clients_v2 "github.com/Layr-Labs/eigenda/api/clients/v2"
+)
+
+// ClientConfigV2 contains all non-sensitive configuration to construct V2 clients
+type ClientConfigV2 struct {
```
samlaf, line 11:
```suggestion
type ClientsV2Config struct {
```

samlaf: does this make more sense?

samlaf: Also didn't we want to separate retriever clients from disperser clients so that we can start proxy in read-only mode?

config/app_config.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda-proxy/metrics"
+	"github.com/urfave/cli/v2"
+)
+
+// AppConfig ... Highest order config. Stores all relevant fields necessary for running both proxy & metrics servers.
+type AppConfig struct {
+	EigenDAConfig ProxyConfig
+	MetricsCfg    metrics.Config
```
samlaf, line 13:
```suggestion
	EigenDAConfig ProxyConfig
	MetricsConfig    metrics.Config
```

samlaf: consistency between both fields

samlaf: Wait also EigenDAConfig feels wrong. Its a ProxyConfig which contains configs for all internal services, not only eigenda backend.

config/eigendaflags/cli.go
```diff
@@ -24,7 +24,6 @@ var (
 	SignerPrivateKeyHexFlagName          = withFlagPrefix("signer-private-key-hex")
 	PutBlobEncodingVersionFlagName       = withFlagPrefix("put-blob-encoding-version")
 	DisablePointVerificationModeFlagName = withFlagPrefix("disable-point-verification-mode")
-	WaitForFinalizationFlagName          = withFlagPrefix("wait-for-finalization")
```
samlaf, line 27: removing this flag is a breaking change. Can you add note to PR description that we'll need to release a new major version of proxy after we merging this PR ?

config/flags.go
```diff
@@ -1,32 +1,31 @@
...
-	StorageDeprecatedCategory  = "DEPRECATED STORAGE FLAGS -- THESE WILL BE REMOVED IN V2.0.0"
-	RedisCategory              = "Redis Cache/Fallback"
-	S3Category                 = "S3 Cache/Fallback"
-	VerifierCategory           = "KZG and Cert Verifier"
-	VerifierDeprecatedCategory = "DEPRECATED VERIFIER FLAGS -- THESE WILL BE REMOVED IN V2.0.0"
+	EigenDAClientCategory   = "EigenDA Client"
+	EigenDAV2ClientCategory = "EigenDA V2 Client"
+	LoggingFlagsCategory    = "Logging"
+	MetricsFlagCategory     = "Metrics"
+	MemstoreFlagsCategory   = "Memstore (for testing purposes - replaces EigenDA backend)"
```
samlaf, line 24: add a note that this category contains flags for both v1 and v2 memstore (I thin it does?)

config/app_config.go
samlaf: Also personally find it a bit weird that our highest-level app config is "proxy + metrics". Feel like it puts too much emphasis on the servers, whereas the proxy is a bunch of services + servers. I feel like MetrisConfig should just be moved to the ProxyConfig and ProxyConfig should be the highest level "app" config.

config/proxy_config.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Enabling will turn on memstore for both V1 && V2
+	//
+	// Currently the memstore is not persisted to disk, and is lost when proxy is stopped. Testing migrations is
+	// currently not supported because it requires restarting the proxy with new flags, which will cause the
+	// memstore db to be wiped.
+	//
+	MemstoreEnabled bool
+
+	EigenDAV2Enabled bool
```
samlaf, line 39: add documentation for semantic of this flag. What does it do? Does it make puts be sent to v2? Or does it enable v2 reading only?

config/proxy_config.go
```diff
@@ -0,0 +1,120 @@
...
+
+	PutRetries       uint
+	MaxBlobSizeBytes uint
+
+	EigenDACertVerifierAddress string
```
samlaf, line 44: why are these here and not part of other lower-level configs? Is it because they are shared between a bunch of different services?

e2e/benchmark_test.go
```diff
@@ -12,8 +12,9 @@ import (
 
 // BenchmarkPutsWithSecondary  ... Takes in an async worker count and profiles blob insertions using
 // constant blob sizes in parallel
+// TODO: test variations with V2
 func BenchmarkPutsWithSecondary(b *testing.B) {
-	testCfg := TestConfig(true)
+	testCfg := TestConfig(true, false)
```
samlaf, line 17: no need to refactor in this PR but going forard the functional options pattern makes things a lot more readable. here we just see (true, false) and it's impossible to know syntactically (without an IDE to click-through) what these mean.

e2e/main_test.go
```diff
@@ -16,27 +17,41 @@ import (
...
+	runIntegrationTestsV2 = flagActivated("INTEGRATION_V2")
+	runIntegrationTests = flagActivated("INTEGRATION")
+	runTestnetIntegrationTests = flagActivated("TESTNET")
 
 	if runIntegrationTests && runTestnetIntegrationTests {
 		panic("only one of INTEGRATION=true or TESTNET=true env var can be set")
 	}
+
+	if runIntegrationTests && runIntegrationTestsV2 {
+		panic("only one of INTEGRATION=true or INTEGRATION_V2=true env var can be set")
```
samlaf, line 49: why is this the case? why can't I run both v1 and v2? Is it because there are race conditions between the two? should be documented if so

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.22.0
 toolchain go1.22.7
 
 require (
-	github.com/Layr-Labs/eigenda v0.8.6-rc.0.0.20250130173836-0d293cc03198
+	github.com/Layr-Labs/eigenda v0.9.0-rc.0.0.20250228180652-c7f0c3d62bbf
```
samlaf, line 8: TODO: make an eigenda release and update this before we merge

server/errors.go
```diff
@@ -21,7 +21,7 @@ func (me MetaError) Error() string {
 	return fmt.Sprintf("Error: %s (Mode: %s, CertVersion: %b)",
 		me.Err.Error(),
 		me.Meta.Mode,
-		me.Meta.CertVersion)
+		me.Meta.Version)
```
samlaf, line 24: why change the name? Its unclear what Version refers to here. I still very much hate `meta`. Can we refactor that? Perhaps in a subsequent PR?

server/handlers.go
```diff
@@ -98,10 +98,15 @@ func (svr *Server) handleGetOPGenericCommitment(w http.ResponseWriter, r *http.R
 	return svr.handleGetShared(r.Context(), w, commitment, commitmentMeta)
 }
 
-func (svr *Server) handleGetShared(ctx context.Context, w http.ResponseWriter, comm []byte, meta commitments.CommitmentMeta) error {
+func (svr *Server) handleGetShared(
+	ctx context.Context,
+	w http.ResponseWriter,
+	comm []byte,
+	meta commitments.CommitmentMeta,
+) error {
```
samlaf, line 106: did the linter config change or something?

samlaf: Seems like there's a lot of formatting changes bloating this PR :'(

store/builder.go
```diff
@@ -0,0 +1,381 @@
...
+			MaxBlobSizeBytes:     uint64(maxBlobSize),
+			EthConfirmationDepth: d.v1VerifierCfg.EthConfirmationDepth,
+			StatusQueryTimeout:   d.v1EdaClientCfg.StatusQueryTimeout,
+			PutRetries:           putRetries,
+		},
+	)
+}
+
+// BuildManager ... Builds storage manager object
+func (d *Builder) BuildManager(
```
samlaf, line 193: Move Public methods to top of file, right below constructor. Otherwise I have to search for this function, which is really the only important one

samlaf (commented):
First pass. Still have 10 or so files to read, but tired

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
...
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
+            "env": {
+                "ENV_PATH": "../../.env",
+                "EIGENDA_PROXY_EIGENDA_CERT_VERIFICATION_DISABLED": "true",
+                "EIGENDA_PROXY_EIGENDA_TARGET_KZG_G1_PATH": "../../resources/g1.point",
+                "EIGENDA_PROXY_EIGENDA_TARGET_KZG_G2_POWER_OF_2_PATH": "../../resources/g2.point.powerOf2"
+            },
+            "program": "cmd/server"
```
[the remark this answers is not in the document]
litt3: > have we tested that this works?

I have not. I don't use vscode, I simply left this as-is from the previous PR. 

cc @ethenotethan

common/secret_config.go
litt3: By URLs, you mean eth rpc urls?

That isn't here (or anywhere) in v2 configs, since we only have a single eth client that's shared between v1 and v2 proxy code. I had originally placed the url here, but removed it because it wasn't in use, and it wasn't clear how to handle potential discrepancies between v1 and v2 configuration. 

There are a couple of potential options here:
1. Configure an eth RPC here in the secret config for v2, and instantiate 2 eth clients: don't share eth clients between v1 and v2 proxy
2. Configure an eth RPC here, and throw an error if v1 and v2 are both enabled, and the configured values don't precisely match
3. Leave as-is, and rely on v1 configuration for eth rpc. That configuration would need to be moved to v2 when v1 is terminated
    - upon further consideration, I don't like this idea. I think we should choose option 1 or 2

cmd/server/entrypoint.go
litt3: Right, that's the idea.

This additional check is just to cover a vulnerability detected by copilot for v1 configuration (which appeared to be correctly identified, at first glance)

cmd/server/main.go
litt3: I'm not 100% why this was moved (I believe it was part of the changeset when I picked up this task). But my guess is that this was to resolve circular imports, which got nasty when I tried a bit of reorganizing.

commitments/mode.go
litt3: Removed this comment [0f7551f](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/0f7551f013acee9065018405807b502d4cd07f2b)

common/client_config.go
litt3: > does this make more sense?

I don't feel strongly about this- my tendency would be to leave it as-is to avoid refactor work. If you feel strongly, I can make the change, LMK

litt3: > Also didn't we want to separate retriever clients from disperser clients so that we can start proxy in read-only mode?

I think this might be a change to consider in the future. The problem with making this change is that I think that train of logic would lead us to abandon the aggregate `ProxyConfig` struct entirely: 

e.g. "what if the user wants a proxy which only reads from relays, and not validator nodes? Perhaps we shouldn't group those configs together either?"

I think this might actually be the right decision in the future, but it would be a large change.

config/eigendaflags/cli.go
litt3: This flag wasn't in use, so I thought its removal *wouldn't* represent a breaking change. I thought that unused flags were simply ignored?

config/flags.go
litt3: I believe that most of these categories contain flags for both v1 and v2, so I'm not sure if that would be a helpful comment here

cmd/server/entrypoint.go
litt3: Agreed, it doesn't make much sense like that.

I did some substantial renaming / reordering cleanup of the `Builder` in [bcd2379](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/bcd237941f223348dab947e3aa6ca6820d300232), and also some minor changes in the code that uses the builder

config/app_config.go
litt3: I agree that there are a substantial number of changes that can be made here. Are you ok with addressing these comments in a [followup PR](https://linear.app/eigenlabs/issue/DAINT-327/cleanup-v2-proxy-config-code), to avoid  additional scope creep of this PR?

config/eigendaflags/v2/cli.go
litt3: The instant I start moving things around, nasty circular dependencies start forming, which have proven difficult to iron out. If it's ok with you, I'd prefer to defer the question of package organization to the follow up "config improvement" PR

common/secret_config.go
litt3: See my comment [here](https://github.com/Layr-Labs/eigenda-proxy/pull/321#discussion_r1979874780)

store/builder.go
litt3: I did this when I cleaned up the builder in [bcd2379](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/bcd237941f223348dab947e3aa6ca6820d300232)

server/handlers.go
litt3: My IDE is configured to automatically reformat lines that grow beyond a reasonable limit, it's not part of the linter.

Personally I think that too-long-lines are substantial impediment to readability, which is why I have tended to commit such automatic changes. I can stop doing so, though, if others don't agree that long lines are a problem

server/errors.go
litt3: This name change was part of the change set as I picked it up.

I will fix it as part of [this issue](https://linear.app/eigenlabs/issue/DAINT-325/standardize-version-related-language-in-proxy)

cmd/server/main.go
litt3: See [this comment](https://github.com/Layr-Labs/eigenda-proxy/pull/321#discussion_r1979874780)

config/proxy_config.go
litt3: Fair point. In [9f73ebe](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/9f73ebe25de836b7dc7aeb3e4ef2d0eaec2b682e) I moved `EigenDACertVerifierAddress ` into a config struct (that's where I should have put it in the first place).

As for `PutRetries` and `MaxBlobSyzeBytes`... I think it makes sense to just leave them where they are for now (this is how they were configured prior to this PR). It's not clear to me upon initial inspection where these would best be moved to.

common/secret_config.go
samlaf: I prefer duplicating everything for v2 yes. Had talked about this with ethen.
Or at least things that are shared should be in a shared config which makes it obvious that they are shared.

config/eigendaflags/v2/cli.go
samlaf: Yep SGTM. Added TODO to PR description.

store/storage_manager_builder.go
```diff
@@ -32,48 +32,123 @@ import (
...
 		v2ClientCfg,
 		v2SecretCfg,
+		putRetries,
+		maxBlobSize,
+		eigenDaCertVerifierAddress,
+	}
+}
+
+// Build builds the storage manager object
+func (smb *StorageManagerBuilder) Build(ctx context.Context) (IManager, error) {
```
samlaf, line 83: Do we have to return an interface here?

commitments/eigenda.go
samlaf: yes

config/proxy_config.go
litt3: As part of [this issue](https://linear.app/eigenlabs/issue/DAINT-324/implement-v1-v2-canary-testing), I want to entirely rework the interaction between v1 and v2, which will include this flag.

e2e/main_test.go
litt3: I don't know why it is like this, @ethenotethan can you chime in here?

config/eigendaflags/cli.go
samlaf: well its not in use but afair its flag actions makes the program crash if its still used.
So image the scenario where someone is on a super old version of proxy and were still using this flag. Now they upgrade to this new version where this flag no longer exists, but we removed the crash at startup. They will never know that this flag is no longer being used (if they use the ENV_VAR instead of flag).

config/flags.go
samlaf: Sorry this is github review UI being weird. Was only highlighting line 24. Because memstore is a backend for which we have 2 versions. we have eigendav1 and eigendav2 categories, so why not for memstore?

server/handlers.go
samlaf: I'm fine with that, I just think it should be part of the linter config, otherwise someone else might come in and then refactor to the old style (if they have a longer max line length) and the CI will just let this back-and-forth go through

samlaf (approved):
Overall LGTM. Let's wait for ethen's review before merging though.

Also adding the list of future refactors/improvements that we've discussed:
- config
   - cleanup https://linear.app/eigenlabs/issue/DAINT-327/cleanup-v2-proxy-config-code
   - package organization
   - testing https://linear.app/eigenlabs/issue/DAINT-326/create-test-proxy-startup-with-env-varssh-for-v2
- canary testing (which will involve changing how v1 / v2 are enabled / disabled) -https://linear.app/eigenlabs/issue/DAINT-324/implement-v1-v2-canary-testing
- proxy version language standardization

.vscode/launch.json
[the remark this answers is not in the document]
anupsv: Is it common to even commit .vscode's configs ?

.golangci.yml
```diff
@@ -134,7 +133,6 @@ linters:
     - errorlint # finds code that will cause problems with the error wrapping scheme introduced in Go 1.13
     - exhaustive # checks exhaustiveness of enum switch statements
     - copyloopvar # checks for pointers to enclosing loop variables
-    - forbidigo # forbids identifiers
```
anupsv, line 137: Instead of removing, do we want to explicitly ignore in certain areas of where this is used ?

.vscode/launch.json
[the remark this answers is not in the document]
samlaf: yes I think its actually nice to have. Good to share vscode config that is specific to each repo's use case.

.golangci.yml
samlaf: My guess is that this one is deprecated. I've had to remove and change 2 of them last time I tried. Not sure about this specific one though.

server/handlers.go
ethenotethan: +100 to having this as a linter enforced std

.golangci.yml
litt3: The problem is that `forbidigo` complains about each an every print statement, of which this repo has lots. I couldn't figure out how to make `forbidigo` stop complaining about prints

store/generated_key/memstore/v2/memstore.go
```diff
@@ -0,0 +1,213 @@
...
+						X: [2]*big.Int{unsafeRandInt(1000), unsafeRandInt(1000)},
+						Y: [2]*big.Int{unsafeRandInt(1000), unsafeRandInt(1000)},
+					},
+					LengthProof: cert_verifier_binding.BN254G2Point{
+						X: [2]*big.Int{unsafeRandInt(1), unsafeRandInt(1)},
+						Y: [2]*big.Int{unsafeRandInt(1), unsafeRandInt(1)},
+					},
+					Commitment: g1CommitPoint,
+					// #nosec G115 - can never overflow on 16MiB blobs
+					Length: uint32(len(blobContents)),
```
ethenotethan, line 115: Let's change this to be the # of field elements instead

common/secret_config.go
litt3: Sounds good, I will plan on going with option 1, on the followup config PR

store/storage_manager_builder.go
```diff
@@ -0,0 +1,373 @@
...
+	metrics metrics.Metricer
+
+	managerCfg     Config
+	memConfig      *memconfig.SafeConfig
+	v1VerifierCfg  verify.Config
+	v1EdaClientCfg clients.EigenDAClientConfig
+	v2ClientCfg    common.ClientConfigV2
+	v2SecretCfg    common.SecretConfigV2
+
+	putRetries       uint
```
ethenotethan, line 49: Can we add a comment that these should eventually be embedded into their own config vs passing via isolated values to build calls?

config/flags.go
litt3: > Was only highlighting line 24

Yep, understood. My point is just that most of these other categories also have v1 and v2 things mixed together, so it's not a unique issue with the memstore category. This is an argument for cleaning up the flag config system, for sure

> so why not for memstore?

I'm not sure the reason that memstore is combined to a single category.

store/storage_manager_builder.go
ethenotethan: @samlaf probably not but I don't see any benefits to the latter

server/handlers.go
litt3: >  I just think it should be part of the linter config,

Definitely agree. It's a big step to add it to the linter unfortunately, since it involves refactoring the whole repo for line length... In the future I'll be more cautious about my formatting preferences blowing up the size of a PR (particularly ones that are already large)

store/storage_manager_builder.go
```diff
@@ -0,0 +1,373 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("new payload disperser: %w", err)
+	}
+
+	return payloadDisperser, nil
+}
+
+// buildLocalSigner attempts to check the pending balance of the created signer account. If the check fails, or if the
+// balance is determined to be 0, the user is warned with a log. This method doesn't return an error based on this check:
+// it's possible that a user could want to set up a signer before it's actually ready to be used
```
ethenotethan, line 349: Can we add a comment explaining that this is a heuristic with a todo describing how we could extend this in the future to do checks against the payment vault or verify payment states through some accessor on the disperser's accountant

cmd/server/entrypoint.go
```diff
@@ -83,27 +117,39 @@ func StartProxySvr(cliCtx *cli.Context) error {
...
 
 // TODO: we should probably just change EdaClientConfig struct definition in eigenda-client
 func prettyPrintConfig(cliCtx *cli.Context, log logging.Logger) error {
+	redacted := "******"
+
 	// we read a new config which we modify to hide private info in order to log the rest
-	cfg := server.ReadCLIConfig(cliCtx)
-	if cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex != "" {
-		cfg.EigenDAConfig.EdaClientConfig.SignerPrivateKeyHex = "*****" // marshaling defined in client config
+	cfg, err := config.ReadCLIConfig(cliCtx)
```
ethenotethan, line 131: Nit - can change function name to 'ReadFromCLI' since referenced from a config package

common/client_config.go
```diff
@@ -0,0 +1,40 @@
...
+	EigenDACertVerifierAddress      string
+}
+
+// Check checks config invariants, and returns an error if there is a problem with the config struct
+func (cfg *ClientConfigV2) Check() error {
+	if cfg.ServiceManagerAddress == "" {
+		return fmt.Errorf("service manager address is required for using EigenDA V2 backend")
+	}
+
+	if cfg.DisperserClientCfg.Hostname == "" {
```
ethenotethan, line 31: Should we also check port too? IIRC it's also a str

common/client_config.go
```diff
@@ -0,0 +1,40 @@
...
+	// Enabled is true if using eigenDA v2, or false if using eigenDA v1
+	Enabled                  bool
+	DisperserClientCfg       clients_v2.DisperserClientConfig
+	PayloadDisperserCfg      clients_v2.PayloadDisperserConfig
+	RelayPayloadRetrieverCfg clients_v2.RelayPayloadRetrieverConfig
+	// hex address of the EigenDAServiceManager contract
+	ServiceManagerAddress string
+	PutRetries            uint
+	// interval at which the ethereum node is polled, while waiting to reach a given block number
+	BlockNumberPollIntervalDuration time.Duration
```
ethenotethan, line 21: What if a user passes 0 here by mistake?

common/secret_config.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Check checks config invariants, and returns an error if there is a problem with the config struct
+func (s *SecretConfigV2) Check() error {
+	if s.SignerPaymentKey == "" {
```
ethenotethan, line 15: We can also check the length to verify it's 32 bytes

config/eigendaflags/v2/cli.go
```diff
@@ -0,0 +1,240 @@
...
+		Hostname:          hostStr,
+		Port:              portStr,
+		UseSecureGrpcFlag: !ctx.Bool(DisableTLSFlagName),
+	}, nil
+}
+
+func readRetrievalConfig(ctx *cli.Context) clients_v2.RelayPayloadRetrieverConfig {
+	return clients_v2.RelayPayloadRetrieverConfig{
+		PayloadClientConfig: readPayloadClientConfig(ctx),
+		RelayTimeout:        ctx.Duration(RelayTimeoutFlagName),
```
ethenotethan, line 238: Should we add a cfg invariant to ensure this isn't 0 set?

e2e/main_test.go
ethenotethan: Did this ensure testmodes were uniquely set. Otherwise atm someone could provide true for both and it'd just use V2 so making them mutually exclusive seemed most practical. We'll probably need to revisit this once we start testing migrations though

.vscode/launch.json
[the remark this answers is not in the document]
litt3: Elsewhere, we use `make` commands to launch things in proxy, it seems kinda weird to have this `.vscode` config file to launch the server. Additionally, this file has some contents I'm not sure make sense:

- it references `.env`, which is in included in the `.gitignore`
- it launches in `debug` mode, so doesn't seem to be targeting users

e2e/main_test.go
litt3: Yeah, this will change once we implement the changes necessary for canary testing: there will no longer be an "either-or" choice between v1 and v2

go.mod
litt3: Update: eigenda release isn't going to happen in the immediate term. This PR will merge with this direct commit reference, and we will just need to coordinate the next proxy release to wait at least until the next eigenda release is complete

store/generated_key/memstore/v2/memstore.go
litt3: [9c714bd](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/9c714bd88aa56f25e3fc64a02ee69eb2b3b6a856)

store/storage_manager_builder.go
litt3: [62d1e1b](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/62d1e1bdb1322e6906c122a2b9af30fda02219d1)

store/storage_manager_builder.go
litt3: [7384ac1](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/7384ac13b72ea64626f34dc04d71e3ff526fe790)

cmd/server/entrypoint.go
litt3: It's possible that packages will shift when [this issue](https://linear.app/eigenlabs/issue/DAINT-327/cleanup-v2-proxy-config-code) is worked, let's leave as-is for now, and address during the cleanup

common/client_config.go
litt3: [f1e4a8c](https://github.com/Layr-Labs/eigenda-proxy/pull/321/commits/f1e4a8ce604178ff6a0dc8e2962ff3c2cd8cd6c1)

common/client_config.go
litt3: Cert verifier [interprets](https://github.com/Layr-Labs/eigenda/blob/e36cbb4637efbde1f889664dd67c2b3628ef5571/api/clients/v2/verification/cert_verifier.go#L79) interval <= 0 to mean "don't wait for block to advance", and any call made with a behind eth node will fail immediately

common/secret_config.go
litt3: Since the key comes in as a string, which may or may not include `0x` prefix, this would entail a rather awkward check `length must equal 64 or 66`

I think it's reasonable to just let the initialization of the signer fail if the string representing the private key is invalid.

.golangci.yml
samlaf: Yeah it wants us to use log.Print instead of fmt.Print. I’ve never delved into why. Fine with removing

anupsv (approved):

samlaf (approved):

PR #325, opened 2025-03-04
ci: separate holesky-test into its own test

samlaf:
Also lower its timeout to 30mins
Holesky is broken so this test is flaky. Makes sense to sepraate it into its own job so that we can rerun only this test when it fails.

## Fixes Issue

Fixes #

### DIFF
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -7,7 +7,7 @@ on:
     branches: ["main"]
 
 env:
-  GO_VERSION: '1.22'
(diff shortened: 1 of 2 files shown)


litt3 (approved):

samlaf:
This seems completely broken and I'm not exactly sure why (seems like an eigenda issue? given that this worked this afternoon at some point and completed in ~20 minutes). I suggest we merge this nonetheless, and fix the holesky tests in another PR.

litt3 (approved):

PR #326, opened 2025-03-05
chore: add flake.nix to create reproducible dev env

### LINKED ISSUE #236: Create reproducible dev env

samlaf:
## Fixes Issue

Fixes #236 

TODO:
1. write GHA: https://nix.dev/guides/recipes/continuous-integration-github-actions
2. nixify docker builds: https://nix.dev/tutorials/nixos/building-and-running-docker-images.html
   2.1 maybe also look into https://github.com/reproducible-containers/buildkit-nix
3. switch to gomod2nix: https://jameswillia.ms/posts/go-nix-containers.html
4. treefmt + others: https://bmcgee.ie/posts/2023/11/nix-my-workflow/


samlaf:
Closing this since decided to go for the simpler mise: #380

PR #330, opened 2025-03-06
docs: cleanup readme by separating users from contributors

### LINKED ISSUE #310: Add API documentation
We don't have a basic integration guide to show how to submit and retrieve blobs using proxy right now. We should add:
1. api spec (can be as simple as [op's](https://specs.optimism.io/experimental/alt-da.html#da-server), or a full openapi spec)
2. show example curl requests
3. document that the standard_commitment routes should be used by most customers (outside of op rollups basically, right now we only mention that standard_commitment is for nitro rollups)
4. ~example code to use the standard client (which is the client avss should be using)~ moved to #332

samlaf:
First time contributors were lost and didn't understand how to setup proxy and interact with it.
Added a user guide to the readme which gives a quick start example of spinning up proxy in memstore mode, dispersing and retrieving a payload.
Also added API route documentation.
Also refactored a bunch of the sections and TOC to make the whole readme easier to parse.

## Fixes Issue

Fixes #310 (missing examples for standard client usage)


README.md
```diff
@@ -25,119 +25,135 @@ Features:
...
 
-* 4 GB RAM
-* 1-2 cores CPU
+```bash
+# Start the proxy with memstore backend enabled
+$ docker run --rm -p 3100:3100 ghcr.io/layr-labs/eigenda-proxy:latest --memstore.enabled --port 3100
 
-### Ethereum Node Requirements
+# In another terminal... submit a payload save the returned cert in hex format
+$ CERT_HEX=$(curl -X POST -d my-eigenda-payload "http://127.0.0.1:3100/put?commitment_mode=standard" | xxd -p | tr -d ' \n')
```
gpsanant, line 67: recommend a golang example in addition that does the exact same thing

README.md
```diff
@@ -1,8 +1,8 @@
-# EigenDA Proxy
+# EigenDA Proxy <!-- omit from toc -->
 
-A basic REST server for Rollups with the following endpoints:
-- POST routes: used by sequencers to encode their payloads (batches of transactions or state transitions) into EigenDA blobs, submit them to the EigenDA disperser to make them available, and receive a DA certificate to be submitted to the rollup's batcher inbox.
-- GET routes: used by validators to retrieve the blob from the EigenDA network given a DA Certificate, decode the blob back into the original payload, and return it to the validator, to be used in its derivation pipeline.
+A basic REST proxy server to interact with the EigenDA network:
+- POST routes: submit payloads (rollup txs, state-diffs, or anything really) that will be encoded into EigenDA blobs and submitted to the EigenDA disperser to make available for 2 weeks. A DA certificate of availability will be returned, which can be used to validate the availability and query the payload back.
```
litt3, line 4: nit: these super long lines make it hard to read the markdown in a text editor or IDE without autowrapping. IMO it makes sense to manually insert line breaks at reasonable intervals. But as we've discussed before, this should probably be handled by the linter

README.md
```diff
@@ -1,8 +1,8 @@
...
+- GET routes: submit back the received DA Certificate to retrieve the blob from the EigenDA network, which will be decoded, validated, and returned as a response.
```
litt3, line 5: "submit back the received DA Certificate"

This phrase is awkward, consider simplifying to "submit the DA Certificate"

litt3 (approved):

ethenotethan (approved):
LGTM but a bit worried about this README. would be good to sitdown and devise what should go into our existing front facing [docs](https://docs.eigenda.xyz/integrations-guides/dispersal/clients/eigenda-proxy) vs overloading here

README.md
samlaf: yep will do in a later PR because I don't think golang stuff should live in README. Added https://github.com/Layr-Labs/eigenda-proxy/issues/332

samlaf:
> LGTM but a bit worried about this README. would be good to sitdown and devise what should go into our existing front facing [docs](https://docs.eigenda.xyz/integrations-guides/dispersal/clients/eigenda-proxy) vs overloading here

Are you talking about the `Blob Lifecycle` section? If so I agree, plan on moving that to spec and/or docs in a subsequent PR. If you were talking about some other section, which one?

README.md
samlaf: [fdacb98](https://github.com/Layr-Labs/eigenda-proxy/pull/330/commits/fdacb98f16fbf54bf3bdb2048a6f8a446ea404c1)

PR #331, opened 2025-03-06
chore: add linter to enforce max line lengths of 100

samlaf:
This is enforced by the lll linter that we add to the golangci-lint config.
Also use golines as part of `make format` to reformat long lines.
The only problem is that long strings are not formatted by golines, and there doesn't seem to be a way to stop lll from complaining about them. So we would need to change those by hand; there's prob 10-15 of them so not that bad.

Currently using line lengths of 100 max. If you guys agree, then I'll just go ahead and fix the linter by running `make format` and fixing the long strings manually.

## Fixes Issue

No issue created for this, but having this came up as part of this [PR review](https://github.com/Layr-Labs/eigenda-proxy/pull/321#discussion_r1978285364)


.golangci.yml
```diff
@@ -61,12 +61,15 @@ linters-settings:
         # Whether to skip (*x).method() calls where x is a pointer receiver.
         # Default: true
         skipRecvDeref: false
+  lll:
+    # Make sure `golines` formatter command in the Makefile matches this.
+    line-length: 100
```
litt3, line 66: I tend to find 120 characters substantially less disruptive than 100 characters (and it would likely result in a vastly smaller diff on the format commit)

Anecdotally, my monitor / font size setup is able to display two files side-by-side with maxlen 120, without needing to scroll (and I use a pretty large font size)

samlaf: [c064fa8](https://github.com/Layr-Labs/eigenda-proxy/pull/331/commits/c064fa82496b3988f55d4b835df7d531385e1da2)

samlaf:
Made max line length 120, and fixed all of the lint issues.

litt3 (approved):

PR #337, opened 2025-03-10
chore: Do test cleanup

litt3:
## Fixes Issue

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,8 @@ test-unit:
 
 # Local V1/V2 E2E tests, leveraging op-e2e framework. Also tests the standard client against the proxy.
 test-e2e-local:
-	INTEGRATION=true go test -timeout 1m ./e2e -parallel 4
-	INTEGRATION_V2=true go test -timeout 1m ./e2e -parallel 4
+	INTEGRATION=true ENABLE_V2=true  go test -timeout 1m ./e2e -parallel 4 -count=1
+	INTEGRATION=true ENABLE_V2=false go test -timeout 1m ./e2e -parallel 4 -count=1
 
 # E2E tests against holesky testnet
 # Holesky is currently broken after recent pectra hardfork.
@@ -43,7 +43,8 @@ test-e2e-local:
 test-e2e-holesky:
 	# Add the -v flag to be able to observe logs as the run is happening on CI
 	# given that this test takes >20 mins to run. Good to have early feedback when needed.
-	TESTNET=true go test -v -timeout 30m ./e2e  -parallel 4
+	TESTNET=true ENABLE_V2=true  go test -v -timeout 30m ./e2e  -parallel 4 -count=1
+	TESTNET=true ENABLE_V2=false go test -v -timeout 30m ./e2e  -parallel 4 -count=1
 
 # E2E test which fuzzes the proxy client server integration and op client keccak256 with malformed inputs
 test-e2e-fuzz:
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -17,47 +17,96 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// Integration tests are run against memstore whereas
-// testnet integration tests are run against eigenda backend talking to testnet disperser.
-// Some of the assertions in the tests are different based on the backend as well.
-// e.g, in TestProxyServerCaching we only assert to read metrics with EigenDA
-// when referencing memstore since we don't profile the eigenDAClient interactions
-var (
-	runTestnetIntegrationTests bool // holesky tests
-	runIntegrationTests        bool // memstore V1 tests
-	runIntegrationTestsV2      bool // memstore V2 tests
-	runFuzzTests               bool // fuzz tests
+// TestType is an enum describing various types of tests. This TestType is used to determine whether
+// a given test should run in the configured test environment.
+type TestType int
+
+const (
+	// StandardIntegration is a test that should run in any integration environment (local, or testnet)
+	StandardIntegration TestType = iota // 0
+	// LocalOnlyIntegration is a test that should run only in a local integration environment, NOT on testnet
+	LocalOnlyIntegration
+	// Fuzz is a fuzz test
+	Fuzz
 )
 
+var (
+	config testFlagConfig
+)
+
+// testFlagConfig contains the boolean flags used to configure test execution
+type testFlagConfig struct {
+	runTestnetIntegrationTests bool // holesky tests
+	runIntegrationTests        bool // memstore tests
+	runFuzzTests               bool // fuzz tests
+	enableV2                   bool
+}
+
+// validate checks that the values in testFlagConfig are sensical, and prints the configured values for convenience
+func (tfc *testFlagConfig) validate() {
+	if tfc.runIntegrationTests && tfc.runTestnetIntegrationTests {
+		panic("only one of INTEGRATION=true or TESTNET=true env var can be set")
+	}
+
+	fmt.Print(
+		"runFuzzTests", tfc.runFuzzTests,
+		"runIntegrationTests", tfc.runIntegrationTests,
+		"testnet_integration_tests", tfc.runTestnetIntegrationTests,
+		"enableV2", tfc.enableV2,
+	)
+}
+
+// flagActivated returns true if a given environment variable is active, otherwise false
 func flagActivated(envVar string) bool {
 	return os.Getenv(envVar) == "true" || os.Getenv(envVar) == "1"
+}
+
+// shouldRunTest accepts a parameter testType, which represents the type of test that is calling shouldRunTest. Based
+// on the testType, this method returns whether the test in question should run based on the configured values.
+func shouldRunTest(testType TestType) bool {
+	switch testType {
+	case StandardIntegration:
+		return config.runIntegrationTests || config.runTestnetIntegrationTests
+	case LocalOnlyIntegration:
+		return config.runIntegrationTests
+	case Fuzz:
+		return config.runFuzzTests
+	default:
+		panic("unknown test type")
+	}
+}
+
+// useMemory returns true if the test should use a memstore backend for local testing
+//
+// Local integration tests are run against memstore whereas testnet integration tests are run against eigenda backend,
+// talking to testnet disperser.
+func useMemory() bool {
+	return !config.runTestnetIntegrationTests
+}
+
+// v2Enabled returns whether v2 should be enabled in a test
+func v2Enabled() bool {
+	return config.enableV2
 }
 
 // ParseEnv ... reads testing cfg fields. Go test flags don't work for this library due to the dependency on Optimism's
 // E2E framework
 // which initializes test flags per init function which is called before an init in this package.
 func ParseEnv() {
-	runFuzzTests = flagActivated("FUZZ")
-	runIntegrationTestsV2 = flagActivated("INTEGRATION_V2")
-	runIntegrationTests = flagActivated("INTEGRATION")
-	runTestnetIntegrationTests = flagActivated("TESTNET")
-
-	if runIntegrationTests && runTestnetIntegrationTests {
-		panic("only one of INTEGRATION=true or TESTNET=true env var can be set")
+	config = testFlagConfig{
+		runTestnetIntegrationTests: flagActivated("TESTNET"),
+		runIntegrationTests:        flagActivated("INTEGRATION"),
+		enableV2:                   flagActivated("ENABLE_V2"),
+		runFuzzTests:               flagActivated("FUZZ"),
 	}
 
-	if runIntegrationTests && runIntegrationTestsV2 {
-		panic("only one of INTEGRATION=true or INTEGRATION_V2=true env var can be set")
-	}
-
-	fmt.Print("fuzz_tests", runFuzzTests, "integration_tests_v1", runIntegrationTests,
-		"integration_tests_v2", runIntegrationTestsV2, "testnet_integration_tests", runTestnetIntegrationTests,
-	)
+	config.validate()
 }
 
 // TestMain ... run main controller
 func TestMain(m *testing.M) {
 	ParseEnv()
+
 	code := m.Run()
 	os.Exit(code)
 }
--- a/e2e/optimism_test.go
+++ b/e2e/optimism_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/Layr-Labs/eigenda-proxy/commitments"
 	"github.com/Layr-Labs/eigenda-proxy/e2e"
 	altda "github.com/ethereum-optimism/optimism/op-alt-da"
-	"github.com/ethereum-optimism/optimism/op-e2e/config"
+	e2econfig "github.com/ethereum-optimism/optimism/op-e2e/config"
 	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils"
 	"github.com/ethereum-optimism/optimism/op-service/sources"
 	"github.com/ethereum-optimism/optimism/op-service/testlog"
@@ -48,7 +48,7 @@ func NewL2AltDA(t actions.Testing, daHost string, altDA bool) *L2AltDA {
 		ChannelTimeout:      120,
 		L1BlockTime:         12,
 		UseAltDA:            true,
-		AllocType:           config.AllocTypeAltDA,
+		AllocType:           e2econfig.AllocTypeAltDA,
 	}
 
 	log := testlog.Logger(t, log.LvlDebug)
@@ -131,11 +131,11 @@ func (a *L2AltDA) ActL1Finalized(t actions.Testing) {
 }
 
 func TestOptimismKeccak256Commitment(gt *testing.T) {
-	if !runIntegrationTests && !runTestnetIntegrationTests && !runIntegrationTestsV2 {
-		gt.Skip("Skipping test as INTEGRATION or TESTNET env var not set")
+	if !shouldRunTest(StandardIntegration) {
+		gt.Skip()
 	}
 
-	testCfg := e2e.TestConfig(useMemory(), runIntegrationTestsV2)
+	testCfg := e2e.TestConfig(useMemory(), v2Enabled())
 	testCfg.UseKeccak256ModeS3 = true
 
 	tsConfig := e2e.TestSuiteConfig(testCfg)
@@ -185,11 +185,11 @@ func TestOptimismKeccak256Commitment(gt *testing.T) {
 }
 
 func TestOptimismGenericCommitment(gt *testing.T) {
-	if !runIntegrationTests && !runTestnetIntegrationTests && !runIntegrationTestsV2 {
-		gt.Skip("Skipping test as INTEGRATION or TESTNET env var not set")
+	if !shouldRunTest(StandardIntegration) {
+		gt.Skip()
 	}
 
-	testConfig := e2e.TestConfig(useMemory(), runIntegrationTestsV2)
+	testConfig := e2e.TestConfig(useMemory(), v2Enabled())
 
 	tsConfig := e2e.TestSuiteConfig(testConfig)
 	tsSecretConfig := e2e.TestSuiteSecretConfig(testConfig)
--- a/e2e/safety_checks_test.go
+++ b/e2e/safety_checks_test.go
@@ -22,12 +22,12 @@ func isNilPtrDerefPanic(err string) bool {
 // TestOpClientKeccak256MalformedInputs tests the NewDAClient from altda by setting and getting against []byte("")
 // preimage. It sets the precompute option to false on the NewDAClient.
 func TestOpClientKeccak256MalformedInputs(t *testing.T) {
-	if !runIntegrationTests || runTestnetIntegrationTests {
-		t.Skip("Skipping test as TESTNET env set or INTEGRATION var not set")
+	if !shouldRunTest(StandardIntegration) {
+		t.Skip()
 	}
 
 	t.Parallel()
-	testCfg := e2e.TestConfig(useMemory(), runIntegrationTestsV2)
+	testCfg := e2e.TestConfig(useMemory(), v2Enabled())
 	testCfg.UseKeccak256ModeS3 = true
 	tsConfig := e2e.TestSuiteConfig(testCfg)
 	tsSecretConfig := e2e.TestSuiteSecretConfig(testCfg)
@@ -72,13 +72,13 @@ func TestOpClientKeccak256MalformedInputs(t *testing.T) {
 // many unicode characters, single unicode character and an empty preimage. It then tries to get the data from the
 // proxy server with empty byte, single byte and random string.
 func TestProxyClientMalformedInputCases(t *testing.T) {
-	if !runIntegrationTests && !runTestnetIntegrationTests && !runIntegrationTestsV2 {
-		t.Skip("Skipping test as INTEGRATION or TESTNET env var not set")
+	if !shouldRunTest(StandardIntegration) {
+		t.Skip()
 	}
 
 	t.Parallel()
 
-	testConfig := e2e.TestConfig(useMemory(), runIntegrationTestsV2)
+	testConfig := e2e.TestConfig(useMemory(), v2Enabled())
 
 	tsConfig := e2e.TestSuiteConfig(testConfig)
 	tsSecretConfig := e2e.TestSuiteSecretConfig(testConfig)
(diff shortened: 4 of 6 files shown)


e2e/optimism_test.go
```diff
@@ -131,11 +131,11 @@ func (a *L2AltDA) ActL1Finalized(t actions.Testing) {
 }
 
 func TestOptimismKeccak256Commitment(gt *testing.T) {
```
samlaf, line 133:
```suggestion
func TestOptimismKeccak256Commitment(t *testing.T) {
```

e2e/main_test.go
```diff
@@ -17,47 +17,94 @@ import (
...
 func flagActivated(envVar string) bool {
 	return os.Getenv(envVar) == "true" || os.Getenv(envVar) == "1"
 }
 
+// shouldRunTest accepts a parameter testType, which represents the type of test that is calling shouldRunTest. Based
+// on the testType, this method returns whether the test in question should run based on the configured values.
+func shouldRunTest(testType TestType) bool {
+	switch testType {
+	case StandardIntegration:
+		return config.runLocalIntegrationTests || config.runTestnetIntegrationTests
```
samlaf, line 67: feels to me like this is basically saying "skip this test when running fuzz tests"?

Can we instead just move fuzz tests to their own standalone package. I think (?) all the fuzz tests are just meant to be independent anyways? Would reduce the complexity of understanding when a test is skipped and why.

e2e/safety_checks_test.go
```diff
@@ -22,12 +22,12 @@ func isNilPtrDerefPanic(err string) bool {
 // TestOpClientKeccak256MalformedInputs tests the NewDAClient from altda by setting and getting against []byte("")
 // preimage. It sets the precompute option to false on the NewDAClient.
 func TestOpClientKeccak256MalformedInputs(t *testing.T) {
-	if !runIntegrationTests || runTestnetIntegrationTests {
-		t.Skip("Skipping test as TESTNET env set or INTEGRATION var not set")
+	if !shouldRunTest(LocalOnlyIntegration) {
+		t.Skip()
```
samlaf, line 26: why is this local and not standard?

Makefile
```diff
@@ -29,9 +29,10 @@ test-unit:
 	go test ./... -parallel 4
 
 # Local V1/V2 E2E tests, leveraging op-e2e framework. Also tests the standard client against the proxy.
+# The -count=1 flag forces these tests to ignore cached test runs, and always execute.
 test-e2e-local:
-	INTEGRATION=true go test -timeout 1m ./e2e -parallel 4
-	INTEGRATION_V2=true go test -timeout 1m ./e2e -parallel 4
+	LOCAL=true ENABLE_V2=true  go test -timeout 1m ./e2e -parallel 4 -count=1
+	LOCAL=true ENABLE_V2=false go test -timeout 1m ./e2e -parallel 4 -count=1
```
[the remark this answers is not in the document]
litt3: I don't have a great understanding of how this cache is invalidated. Based on what I was seeing, the second line here was *never* being executed, and it just relied on the cached result of the first run.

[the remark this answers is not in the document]
samlaf: Here's what I think would make the most sense (generated with Claude's help):
image
image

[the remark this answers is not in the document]
litt3: The problem of this approach is that there are multiple test dimensions we need to test: both v1 vs v2, and local vs holesky.

I've created a simple test matrix framework instead, as we discussed

e2e/main_test.go
```diff
@@ -17,47 +17,94 @@ import (
...
-// Some of the assertions in the tests are different based on the backend as well.
-// e.g, in TestProxyServerCaching we only assert to read metrics with EigenDA
-// when referencing memstore since we don't profile the eigenDAClient interactions
+// TestType is an enum describing various types of tests. This TestType is used to determine whether
+// a given test should run in the configured test environment.
+type TestType int
+
+const (
+	// StandardIntegration is a test that should run in any integration environment (local, or testnet)
+	StandardIntegration TestType = iota
```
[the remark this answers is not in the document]
litt3: This is no longer relevant, with my recent larger set of changes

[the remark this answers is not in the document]
litt3: Enum doesn't exist anymore

e2e/main_test.go
```diff
@@ -17,47 +17,94 @@ import (
...
 	runFuzzTests               bool // fuzz tests
-)
+	enableV2                   bool
+}
 
+// validate checks that the values in testFlagConfig are sensical, and prints the configured values for convenience
+func (tfc *testFlagConfig) validate() {
+	if tfc.runLocalIntegrationTests && tfc.runTestnetIntegrationTests {
+		panic("only one of LOCAL=true or TESTNET=true env var can be set")
+	}
```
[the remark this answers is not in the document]
litt3: With my most recent changes, the mechanism here is different (uses the test matrix instead)

e2e/main_test.go
```diff
@@ -17,47 +17,94 @@ import (
...
+
+// testFlagConfig contains the boolean flags used to configure test execution
+type testFlagConfig struct {
 	runTestnetIntegrationTests bool // holesky tests
-	runIntegrationTests        bool // memstore V1 tests
-	runIntegrationTestsV2      bool // memstore V2 tests
+	runLocalIntegrationTests   bool // memstore tests
 	runFuzzTests               bool // fuzz tests
-)
+	enableV2                   bool
```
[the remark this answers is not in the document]
litt3: V2 test are always run, with my most recent changeset

e2e/main_test.go
```diff
@@ -17,47 +17,94 @@ import (
...
+	fmt.Printf(
+		"runFuzzTests: %v, runLocalIntegrationTests: %v, runTestnetIntegrationTests: %v, enableV2: %v\n",
+		tfc.runFuzzTests, tfc.runLocalIntegrationTests, tfc.runTestnetIntegrationTests, tfc.enableV2,
+	)
+}
+
+// flagActivated returns true if a given environment variable is active, otherwise false
 func flagActivated(envVar string) bool {
 	return os.Getenv(envVar) == "true" || os.Getenv(envVar) == "1"
 }
```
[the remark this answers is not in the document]
litt3: I've changed the tests to not be configured with env vars anymore

e2e/main_test.go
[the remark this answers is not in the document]
litt3: Global config is gone now

e2e/optimism_test.go
litt3: Fixed in [this huge commit](https://github.com/Layr-Labs/eigenda-proxy/pull/337/commits/78fc10bdfc142b46489506ee39e13af3e94d0c20)

e2e/main_test.go
litt3: Fuzz tests moved to a different package

e2e/optimism_test.go
```diff
@@ -185,11 +185,11 @@ func TestOptimismKeccak256Commitment(gt *testing.T) {
 }
 
 func TestOptimismGenericCommitment(gt *testing.T) {
-	if !runIntegrationTests && !runTestnetIntegrationTests && !runIntegrationTestsV2 {
-		gt.Skip("Skipping test as INTEGRATION or TESTNET env var not set")
+	if !shouldRunTest(StandardIntegration) {
+		gt.Skip()
```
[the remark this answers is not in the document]
litt3: Fuzz tests moved to a separate package

e2e/safety_checks_test.go
litt3: Unknown- I changed it to also run on holesky, and the test is passing. I'm going to leave it like that

.github/workflows/per-pr.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Run flag test
         run: ${{ github.workspace }}/scripts/test-proxy-startup-with-env-vars.sh
 
-  test-local:
+  unit-tests:
     outputs:
       COVERAGE: ${{ steps.unit.outputs.coverage }}
```
samlaf, line 42: this can be removed

samlaf: it doesnt actually do anything, plus we don't even have a unit step anymore.

Makefile
```diff
@@ -25,29 +25,20 @@ disperse-test-blob:
...
-# The test failing in CI is currently expected however, so expect to have to re-run it.
-# See https://dora.holesky.ethpandaops.io/epochs for block production status.
-test-e2e-holesky:
+# E2E tests, leveraging op-e2e framework. Also tests the standard client against the proxy.
+# If holesky tests are failing, consider checking https://dora.holesky.ethpandaops.io/epochs for block production status.
+test-e2e:
 	# Add the -v flag to be able to observe logs as the run is happening on CI
 	# given that this test takes >20 mins to run. Good to have early feedback when needed.
-	TESTNET=true go test -v -timeout 30m ./e2e  -parallel 4
+	go test -v -timeout 30m ./e2e -parallel 4
```
samlaf, line 37: think you said we can reduce the timeout to 10 mins now?

samlaf: or at least update the comment saying that these tests run in ~5 mins

Makefile
```diff
@@ -25,29 +25,20 @@ disperse-test-blob:
 clean:
 	rm bin/eigenda-proxy
 
+# Runs all tests, excluding e2e and fuzz tests
 test-unit:
-	go test ./... -parallel 4
+	go test `go list ./... | grep -v ./e2e | grep -v ./fuzz` -parallel 4
```
samlaf, line 30: this works. Think there are cleaner ways to do this but let's roll with this for now.
I think its fine to keep fuzz tests in though? Those should run really fast with only the seed examples when the `-fuzz` test is not passed. They are basically unit test when not in fuzz mode.

Makefile
```diff
@@ -25,29 +25,20 @@ disperse-test-blob:
...
 	# Add the -v flag to be able to observe logs as the run is happening on CI
 	# given that this test takes >20 mins to run. Good to have early feedback when needed.
-	TESTNET=true go test -v -timeout 30m ./e2e  -parallel 4
+	go test -v -timeout 30m ./e2e -parallel 4
 
 # E2E test which fuzzes the proxy client server integration and op client keccak256 with malformed inputs
-test-e2e-fuzz:
-	FUZZ=true go test ./e2e -fuzz -v -fuzztime=5m
+test-fuzz:
+	go test ./fuzz -fuzz -v -fuzztime=5m
```
samlaf, line 41: what happens if we remove `./fuzz`? Wouldnt the `-fuzz` flag only run fuzz tests?
Feels a bit weird on second thought that fuzz tests are in a separate package of their own. Feel like we should be allowed/encouraged to add fuzz tests in separate packages - and they should live close to the actual code. See my comment above for fuzz tests being unit tests when not passed the `-fuzz` flag.

README.md
```diff
@@ -280,18 +280,16 @@ Browse our [Makefile](./Makefile) for a list of available commands such as `make
...
 #### Unit
 
-Unit tests can be ran via invoking `make test-unit`.
+Unit tests can be run with `make test-unit`.
 
-#### Integration / E2E against local deployment
+#### Integration / E2E
 
-Integration tests against op framework can be ran via `make test-e2e-local`. These tests use the [op-e2e](https://github.com/ethereum-optimism/optimism/tree/develop/op-e2e) framework for asserting correct interaction behaviors with batch submission and state derivation.
+Integration tests against op framework can be run with `make test-e2e`. These tests use the [op-e2e](https://github.com/ethereum-optimism/optimism/tree/develop/op-e2e) framework for asserting correct interaction behaviors with batch
```
samlaf, line 287: so `test-e2e` runs both the memstore backend as well as holesky backend?

README.md
```diff
@@ -280,18 +280,16 @@ Browse our [Makefile](./Makefile) for a list of available commands such as `make
...
 
-#### E2E against Holesky
+#### Fuzz
 
-A holesky integration test can be ran using `make test-e2e-holesky` to assert proper dispersal/retrieval against a public network. Please **note** that EigenDA Holesky network which is subject to rate-limiting and slow confirmation times *(i.e, >10 minutes per blob confirmation)*. Please advise EigenDA's [inabox](https://github.com/Layr-Labs/eigenda/tree/master/inabox#readme) if you'd like to spin-up a local DA network for faster iteration testing.
-
-#### E2E Fuzz
-
-This E2E test will fuzz the proxy client server integration and op client keccak256 with malformed inputs. This is never meant to be fuzzed with EigenDA. Run with `make test-e2e-fuzz`.
+Fuzz tests exercise the proxy client server integration and op client keccak256 with malformed inputs. This is
```
samlaf, line 294: what does "never meant to be fuzzed with EigenDA" mean? this only fuzzes the memstore backend??

fuzz/server_fuzz_test.go
```diff
@@ -1,25 +1,29 @@
...
 	"testing"
 	"unicode"
 
 	"github.com/Layr-Labs/eigenda-proxy/clients/standard_client"
-	"github.com/Layr-Labs/eigenda-proxy/e2e"
 )
 
-// FuzzProxyClientServerIntegrationAndOpClientKeccak256MalformedInputs will fuzz the proxy client server integration
+// FuzzProxyClientServerV1 will fuzz the proxy client server integration
 // and op client keccak256 with malformed inputs. This is never meant to be fuzzed with EigenDA.
```
samlaf, line 14:
```suggestion
// and op client keccak256 with malformed inputs. This is only meant to be fuzzed against the memstore Backend.
```

samlaf: is this what this means? Maybe we should add explanation as to why eigenda backend should be used.
I guess because its too slow..? But then its on us to have a proper inabox setup that could be used to fuzz the actual useful part of the proxy, not just memstore.

e2e/optimism_test.go
```diff
@@ -130,109 +131,139 @@ func (a *L2AltDA) ActL1Finalized(t actions.Testing) {
...
+		t.Run(
+			testConfiguration.ToString(), func(t *testing.T) {
+				v2Enabled, ok := testConfiguration.GetValue(e2e.V2Enabled).(bool)
+				require.True(t, ok)
+
+				environment, ok := testConfiguration.GetValue(e2e.Environment).(e2e.TestEnvironment)
+				require.True(t, ok)
+
+				testCfg := e2e.NewTestConfig(e2e.UseMemstore(environment), v2Enabled)
+				testCfg.UseKeccak256ModeS3 = true
```
samlaf, line 150: Def like the look of this more, but unsure about having a separate matrix testConfigurations, and then another layer testCfg on top.

I guess that's the price to pay for having your untyped dynamic matrix framework. Are we sure that we need its flexibility? Would personally prefer if we started small with a typed hardcoded matrix. But maybe you see otherwise. Just feels like a lot of boilerplate for every test we need to setup...

Perhaps we can move all of this code into yet another testHarness structure? I feel like every test in this file has the same boilerplate to setup the same testmatrix and from it instantiate the testCfg right?

e2e/config_fields.go
```diff
@@ -0,0 +1,27 @@
...
+import "fmt"
+
+type TestEnvironment int
+
+const (
+	// Local environment uses a local memstore analog for the actual eigenDA network
+	Local TestEnvironment = iota + 1
+	// Testnet runs tests against holesky testnet
+	Testnet
+)
```
samlaf, line 12: Do these do anything other than setting up the backend (memstore vs eigenda-holesky)? If not, could we use that explicit vocabulary (backend) instead of "Environment", which feels like another layer of uncertainty for test readers/maintainers.

e2e/benchmark_test.go
```diff
@@ -8,35 +8,49 @@ import (
...
-	testCfg := TestConfig(true, false)
-	testCfg.UseS3Caching = true
-	writeThreadCount := os.Getenv("WRITE_THREAD_COUNT")
-	threadInt, err := strconv.Atoi(writeThreadCount)
-	if err != nil {
-		panic(fmt.Errorf("Could not parse WRITE_THREAD_COUNT field %w", err))
-	}
-	testCfg.WriteThreadCount = threadInt
+	testMatrix := testmatrix.NewTestMatrix()
+	testMatrix.AddDimension(testmatrix.NewDimension(V2Enabled, []any{true, false}))
```
samlaf, line 19: not familiar with golang benchmarking tools, but will this play well with your testMatrix? Or will it just give a benchmark for the entire testMatrix time?

Makefile
litt3: This strategy is from claude, definitely open to a cleaner way

README.md
litt3: Yes, that's correct.

Combining allows test backend to be part of the test matrix, instead of being configured via env vars. I think this could be an issue if holesky tests took a really long time- but since they are quick, I think it makes sense to combine with memstore tests.

README.md
litt3: That's my interpretation, but I'm not sure. I don't know why it would be a problem to fuzz with a real backend... but I left the comment since I'm not sure

.github/workflows/per-pr.yml
litt3: [fe4424a](https://github.com/Layr-Labs/eigenda-proxy/pull/337/commits/fe4424ab371197957a52034eb4e3f9a812884d59)

Makefile
litt3: [3435b21](https://github.com/Layr-Labs/eigenda-proxy/pull/337/commits/3435b21e63d380abac50cbf61c51f58f1e54e117)

Makefile
litt3: Modified, to not bother excluding fuzz tests [2e0edc9](https://github.com/Layr-Labs/eigenda-proxy/pull/337/commits/2e0edc9c64eaadecc1e3f6ac9f5d224eed6d213c)

Makefile
litt3: It appears it's not possible to run all fuzz tests throughout the whole project with a single command. Specifically, this errors out with:

```
cannot use -fuzz flag with multiple packages
```

Claude claims that you need to home roll some solution to combine fuzz tests, either with a single fuzz package, or multiple fuzz commands. Not sure which route we prefer in the future, but single fuzz package seems fine for now.

e2e/optimism_test.go
litt3: In [3996e16](https://github.com/Layr-Labs/eigenda-proxy/pull/337/commits/3996e16b7f73fd0d4296f8c80e6dbd4c2d35f0a2), I did some reorganization, and added helper methods to decrease the amount of boilerplate per test. I may make some more modifications in following commits to make this more palatable, still working through your other comments.

e2e/config_fields.go
litt3: Renamed [657ada3](https://github.com/Layr-Labs/eigenda-proxy/pull/337/commits/657ada33d9cce2ae0f383ccd259ff55aae809103)

README.md
ethenotethan: I'd argue against having them consolidated since holesky tests can fail due to an assortment of issues and rely on external dependency for success

README.md
ethenotethan: > I don't know why it would be a problem to fuzz with a real backend... but I left the comment and test configuration alone, since I'm not sure

This is probably due to the fact that there'd be significantly incurred latency when dealing with a real network environment and the input seed corpus would take forever to exhaust. Fuzzing is typically done on isolated logic sets that can be exhausted over a large pseud-unbounded input space

e2e/benchmark_test.go
```diff
@@ -8,35 +8,47 @@ import (
...
-	testCfg.UseS3Caching = true
-	writeThreadCount := os.Getenv("WRITE_THREAD_COUNT")
-	threadInt, err := strconv.Atoi(writeThreadCount)
-	if err != nil {
-		panic(fmt.Errorf("Could not parse WRITE_THREAD_COUNT field %w", err))
-	}
-	testCfg.WriteThreadCount = threadInt
+	testMatrix := testmatrix.NewTestMatrix()
+	testMatrix.AddDimension(testmatrix.NewDimension(testutils.V2Enabled, []any{true, false}))
+	testMatrix.AddDimension(testmatrix.NewDimension(testutils.Backend, []any{testutils.Memstore}))
```
ethenotethan, line 20: this feels overly verbose and more error prone than what we had before since we're introducing the iteration within each test case - is it not possible to capture the matrix in a way that doesn't require introducing a table driven test over each e2e test?

fuzz/server_fuzz_test.go
ethenotethan: its latency and gurantees. Fuzzing at the memstore level gives us validation of the correctness of the server's request processing and handling logic. Fuzzing while using EigenDA via proxy wouldn't really do much imo and creates a surface area that's harder to navigate. It'd be better fit to fuzz at the disperser client instead vs using fuzzing with proxy against EigenDA

Makefile
anupsv: Yea usually people thread out multiple package fuzzing through Makefile.

anupsv: Running without -fuzz will run the tests with just the seed inputs.

samlaf: interesting so it can run multiple fuzz tests at the same time if they are in the same package? But not if they are in different packages? Don't understand why that makes sense

README.md
samlaf: hmm I think I agree with ethen here.

fuzz/server_fuzz_test.go
samlaf: That's a good point. Just looked at what its doing. Its just creating arbitrary byte inputs and sending them to proxy via standard_client? It's really not clear that its testing anything useful to me. After we've tested the edge cases (0 bytes, 1 byte, 32 byte, 16MiB), really not clear what other interesting inputs exist (actually I think the fuzz framework tells us if it found any interesting inputs while fuzzing... I'd bet it doesn't but would be curious to know).

I think i'd vote for getting rid of this fuzz test then. as you say, eigenda clients themselves should be fuzzed. don't think fuzzing here adds any value, and I think it actually creates fake value; I don't like having tests that feel like they are doing something when they really aren't personally.

e2e/config_fields.go
samlaf: tip: golang enums are super dumb, so best to name the options with the type as suffix, like `MemstoreTestBackend`

Makefile
samlaf: Oh I just realized I think we only have one function in the package right? That might explain it?

README.md
litt3: > holesky tests can fail due to an assortment of issues and rely on external dependency for success

@ethenotethan Fair enough 🤔 

@samlaf You weren't a fan of the env var test configuration, but I'm not sure I see a great alternative it we want to keep memstore vs testnet tests separate. 

What we could do here is build v1 vs v2 testing into the tests themselves, and configure backend with an environment variable, to allow us to separate execution.

Thoughts?

litt3: Ok, my most recent iteration takes the middle road: 

holesky vs local is decided with an env var, to allow the tests to be easily split into different actions. V1 vs V2 is decided in the test files themselves, though, instead of with an env var

e2e/config_fields.go
litt3: Good to know. Enum is gone in most recent iteration, though

e2e/benchmark_test.go
litt3: I split this apart into separate tests, for v1 and v2.

e2e/benchmark_test.go
litt3: Based on your feedback, I went back to the drawing board and made things much simpler. LMK what you think

fuzz/server_fuzz_test.go
samlaf: I guess. I looked at the fuzzer test closely and it was doing very weird stuff. First off I don't even think it was fuzzing.... its seed corpus was the first 1M integers being stringified and sent. I don't even think it managed to actually get to fuzzing lol... but not sure. When I cleaned up the seed corpus and removed the cluttering log outputs, this is what it output lol
image

samlaf: Fixed it, see https://github.com/Layr-Labs/eigenda-proxy/pull/339

anupsv: > I guess. I looked at the fuzzer test closely and it was doing very weird stuff. First off I don't even think it was fuzzing.... its seed corpus was the first 1M integers being stringified and sent. I don't even think it managed to actually get to fuzzing lol... but not sure. When I cleaned up the seed corpus and removed the cluttering log outputs, this is what it output lol
> 
> image
> 
> 

Not sure about this. If I remember correcy it added all the printable unicode characters right ? 

Although I'm not sure why it says nothing to fuzz. Maybe something changed in the previous changes.

benchmark/benchmark_test.go
```diff
@@ -0,0 +1,55 @@
...
+// constant blob sizes in parallel. Exercises V1 code pathways
+func BenchmarkPutsWithSecondaryV1(b *testing.B) {
+	testCfg := testutils.NewTestConfig(true, false)
+	putsWithSecondary(b, testCfg)
+}
+
+// BenchmarkPutsWithSecondaryV2  ... Takes in an async worker count and profiles blob insertions using
+// constant blob sizes in parallel. Exercises V2 code pathways
+func BenchmarkPutsWithSecondaryV2(b *testing.B) {
+	testCfg := testutils.NewTestConfig(true, true)
```
samlaf, line 24: can you change this constructor to use functional options like I did for the fuzzer? Makes it much easier to read what exactly is the option being passed I find.

e2e/safety_checks_test.go
samlaf, line 1: Can we refactor the tests in this file to use tables instead of a bunch of boilerplated t.Runs()? Would make it a lot easier to have all the test cases in one spot to quickly understand what exactly our test cases are. Right now they are dispersed through a lot of repeating boilerplate. Specifically thinking of `testProxyClientMalformedInputCases` but there must be more

litt3: @samlaf Are you ok fixing this pre-existing issue in a followup PR? I agree it could use some cleanup, but I'm not familiar with tables, so it will take some figuring out on my part.

samlaf (approved):
LGTM. Let's wait for @ethenotethan's stamp if he's still awake. Otherwise as we talked this is getting large so let's merge and do a subsequent PR on top.

ethenotethan (approved):
LGTM with some super minor nits

testutils/setup.go
```diff
@@ -106,7 +105,12 @@ func startRedisContainer() error {
 	return nil
 }
 
-type Cfg struct {
+func UseMemstore() bool {
+	envVar := "MEMSTORE"
```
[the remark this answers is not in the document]
litt3: Done [dc88402](https://github.com/Layr-Labs/eigenda-proxy/pull/337/commits/dc884023d90f0e6df84e6fb212900ac3a44433bf)

samlaf (approved):

PR #338, opened 2025-03-13
fix: Enforce V1 cert field sizes

ethenotethan:
## Fixes Issue

Fixes #

## Changes proposed
Mitigates risk where the disperser can increase the size of dynamic array fields to be larger than their expected size limit with verification still passing. If performed, the disperser could:
1. Overload DA commitment sizes used for batcher inbox submissions on OP x EigenDA rollups to grief
2. Break secondary storage on Arbitrum x EigenDA rollups by providing a cert that can't be regenerated perfectly when doing get queries when deriving L2 state from the sequencer inbox 

Having this size enforcement also securely enables https://github.com/Layr-Labs/nitro/pull/76

### DIFF
--- a/store/generated_key/eigenda/eigenda.go
+++ b/store/generated_key/eigenda/eigenda.go
@@ -136,7 +136,12 @@ func (e Store) Put(ctx context.Context, value []byte) ([]byte, error) {
 
 	err = cert.NoNilFields()
 	if err != nil {
-		return nil, fmt.Errorf("failed to verify DA cert: %w", err)
+		return nil, fmt.Errorf("failed to verify DA cert due to nil fields: %w", err)
+	}
+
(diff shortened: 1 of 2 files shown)


samlaf (approved):
LGTM. Really unfortunate that protobuf doesn't have better types, so we're stuck using bytes instead of uint256 for example. I did recommend https://github.com/Layr-Labs/eigenda/pull/1104#discussion_r1990369975 that we use https://github.com/bufbuild/protovalidate?tab=readme-ov-file on disperser to enforce the actual types we care about.

Given that we've made the unfortunate choice in v1 to use protobuf instead of solidity types, then adding this extra validation in proxy seems like the right solution.

anupsv (approved):

litt3 (approved):

PR #342, opened 2025-03-19
Add migration tests

litt3:


samlaf (commented):
First pass on architecture. Haven't looked at tests yet. We can chat offline regarding:
image

e2e/optimism_test.go
```diff
@@ -215,7 +226,10 @@ func testOptimismGenericCommitment(t *testing.T, v2Enabled bool) {
 
 	optimism.sequencer.ActL2PipelineFull(ot)
 	optimism.sequencer.ActL1SafeSignal(ot)
-	require.Equal(t, uint64(1), optimism.sequencer.SyncStatus().SafeL1.Number)
+
+	expectedBlockNumber := startingBlockNumber + 1
+	println("BLOCK NUMBER", expectedBlockNumber)
```
samlaf, line 231: better to use t.Log in tests, so they dont clutter when not needed (logs will only appear if test fails or `-v` flag is passed to `go test`

server/handlers.go
```diff
@@ -175,7 +175,7 @@ func (svr *Server) handlePostOPGenericCommitment(w http.ResponseWriter, r *http.
 		Version: commitments.CertV0,
 	}
 
-	if svr.cfg.DisperseV2 {
+	if svr.sm.IsV2Enabled() {
 		commitmentMeta.Version = commitments.CertV1
```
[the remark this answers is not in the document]
litt3: No- this is how the cert versions are defined

```
const (
	// EigenDA V1
	CertV0 EigenDACommitmentType = iota
	// EigenDA V2
	CertV1
)
```

Not crazy about this, I've just been carrying it forward

server/handlers.go
```diff
@@ -135,7 +135,7 @@ func (svr *Server) handlePostStdCommitment(w http.ResponseWriter, r *http.Reques
...
```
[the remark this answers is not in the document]
litt3: No, see https://github.com/Layr-Labs/eigenda-proxy/pull/342#discussion_r2004115872

e2e/optimism_test.go
```diff
@@ -204,6 +204,17 @@ func testOptimismGenericCommitment(t *testing.T, v2Enabled bool) {
...
+	exerciseGenericCommitments(t, ot, proxyTS, optimism)
+}
+
+// TODO: @ethen can you help me formulate a description of what this sequence of calls is doing?
+func exerciseGenericCommitments(
+	t *testing.T,
+	ot actions.StatefulTesting,
+	proxyTS testutils.TestSuite,
+	optimism *L2AltDA,
+) {
```
samlaf, line 216: let's try to refactor this function into smaller understandable parts. For example feels like we should be able to extract a "buildL1Block" function, but right now block1 and block2 are being built differently. Probably can refactor and extract out the FinalizationSignal in block2.

Also I personally think the require() should be moved out of this function and into tests. I prefer to see all the require statements when looking at a test. Helps see with a single look what exactly is being tested.

e2e/server_test.go
```diff
@@ -382,3 +383,60 @@ func testProxyMemConfigClientCanGetAndPatch(t *testing.T, v2Enabled bool) {
...
+	fetchedPayload1a, err := client.GetData(testSuite.Ctx, cert1a)
+	require.NoError(t, err)
+
+	require.Equal(t, payload1a, fetchedPayload1a)
+	require.Equal(t, payload2a, fetchedPayload2a)
+	require.Equal(t, payload1b, fetchedPayload1b)
+	require.Equal(t, payload2b, fetchedPayload2b)
+
+	// requireStandardClientSetGet(t, testSuite, testutils.RandBytes(100))
+	// requireDispersalRetrievalEigenDA(t, testSuite.Metrics.HTTPServerRequestsTotal, commitments.Standard)
```
samlaf, line 441: commtented out on purpose? can we delete? or did you want to test this but not sure how to?

server/server.go
```diff
@@ -139,6 +137,11 @@ func (svr *Server) Stop() error {
 	return nil
 }
 
+// SetDisperseV2 configures which version of eigenDA the server disperses to
+func (svr *Server) SetDisperseV2(disperseV2 bool) {
+	svr.sm.SetV2Enabled(disperseV2)
```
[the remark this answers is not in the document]
litt3: Yes, renamed [edff842](https://github.com/Layr-Labs/eigenda-proxy/pull/342/commits/edff84202a0909e694c8b168f38e74c3f0f869f2) across the board

e2e/server_test.go
```diff
@@ -382,3 +383,60 @@ func testProxyMemConfigClientCanGetAndPatch(t *testing.T, v2Enabled bool) {
 	require.NoError(t, err)
 	require.Equal(t, cfg.PutLatency, expectedChange)
 }
+
+// TestInterleavedVersions alternately disperses payloads to v1 and v2, and then retrieves them.
+func TestInterleavedVersions(t *testing.T) {
+	t.Parallel()
```
samlaf, line 389: this would be a good candidate test for fuzzing, or at least large scale random generation. Like instead of just testing 4 dispersals, generate a ton, or even better let fuzzer generate them (having `useV2` as a fuzzer input variable).

samlaf (commented):
Tests review

server/handlers.go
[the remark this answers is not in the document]
samlaf: oh.... kek. Should we rename? CertV0 makes no sense. We can still keep the iota numbering starting from 0. But at least the name will make sense. @ethenotethan thoughts?

store/manager.go
```diff
@@ -26,33 +29,41 @@ type Manager struct {
...
 	// secondary storage backends (caching and fallbacks)
 	secondary ISecondary
 }
 
+var _ IManager = &Manager{}
+
+// IsV2Enabled returns whether v2 dispersal is enabled
+func (m *Manager) IsV2Enabled() bool {
+	return m.writeV2.Load()
+}
```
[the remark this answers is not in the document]
litt3: Makes sense, renamed [edff842](https://github.com/Layr-Labs/eigenda-proxy/pull/342/commits/edff84202a0909e694c8b168f38e74c3f0f869f2)

server/server.go
```diff
@@ -24,7 +24,6 @@ var (
 )
 
 type Server struct {
-	cfg        config.ServerConfig
```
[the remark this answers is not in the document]
litt3: Moving some configs to the `config/` package was a consequence of how [flags are being used](https://github.com/Layr-Labs/eigenda-proxy/blob/b80bc7c3ee799de3e0bea58d9b70eaf3888373a1/config/flags.go#L63)

Since the flags for a given category are defined in the various packages, I was struggling with circular imports. It's possible there is a better package structure out there, I'm just not sure what it would be

e2e/optimism_test.go
litt3: Leaving this print was accidental, removed [58a7c40](https://github.com/Layr-Labs/eigenda-proxy/pull/342/commits/58a7c40bbc27d5e36332ace269c39d53b506c8e4)

e2e/server_test.go
litt3: I was still working on this test when preprod stopped working, so it was partially done.

in [58a7c40](https://github.com/Layr-Labs/eigenda-proxy/pull/342/commits/58a7c40bbc27d5e36332ace269c39d53b506c8e4) I tried my best to put this right, but I'll have to test once preprod is back to be sure

server/handlers.go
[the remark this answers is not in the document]
litt3: I've no strong feelings for changing vs not. I've been generally trying (maybe failing) to keep diffs as small as possible, which is why I left it thus far

e2e/server_test.go
samlaf: yep brainfart of my part, you're right.

server/server.go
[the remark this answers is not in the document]
samlaf: Ah I see. Prob just need to pop the config and flags in a subpackage of each respective package. Like `server/config/config.go` and `server/flags/flags.go`. I think that would resolve it. But also totally fine to leave as is.

e2e/optimism_test.go
litt3: [9bfbda9](https://github.com/Layr-Labs/eigenda-proxy/pull/342/commits/9bfbda9538fd0c3d2563f7f2c0fc3fef02f30ed2)

Made a few tweaks, and left a note to improve further in the future

server/server.go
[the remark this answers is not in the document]
litt3: Let's leave as-is for now, and keep in mind for potential future proxy package structure improvements (maybe as part of merging proxy into core)

e2e/optimism_test.go
```diff
@@ -238,6 +259,24 @@ func testOptimismGenericCommitment(t *testing.T, v2Enabled bool) {
...
 	optimism.ActL1Finalized(ot)
+}
+
+func TestOptimismGenericCommitmentMigration(t *testing.T) {
+	// initialize with v2 = true, so that the necessary components are constructed
+	proxyTS, shutDown := testutils.CreateTestSuite(testutils.GetBackend(), true)
+	defer shutDown()
+
+	// turn off v2 dispersal, so that we exercise v1, i.e. "pre migration"
+	proxyTS.Server.SetDisperseV2(false)
```
ethenotethan, line 270: Definitely out of scope for this PR but would be nice to remove this Boolean longterm in favor of some 'ProtocolVersion' enum - pseudo equivalent to how node software like geth maintains hard fork versioning

e2e/optimism_test.go
```diff
@@ -238,6 +259,24 @@ func testOptimismGenericCommitment(t *testing.T, v2Enabled bool) {
...
+
+	ot := actions.NewDefaultTesting(t)
```
ethenotethan, line 272: How are we actually asserting that proxy is dual processing both V1 and V2 if we're only verifying dispersal/retrieval based on a generalized metrics vector. Would there be merit in extending the telemetry to also support a versioning field?

store/manager.go
```diff
@@ -24,45 +27,53 @@ type Manager struct {
 
 	s3 common.PrecomputedKeyStore // for op keccak256 commitment
 	// For op generic commitments & standard commitments
-	eigenda   common.GeneratedKeyStore // v0 da commitment version
-	eigendaV2 common.GeneratedKeyStore // v1 da commitment version
-	writeV2   bool                     // write blobs to EigenDAV2 backend
+	eigenda    common.GeneratedKeyStore // v0 da commitment version
+	eigendaV2  common.GeneratedKeyStore // v1 da commitment version
+	disperseV2 *atomic.Bool             // disperse blobs to EigenDAV2 backend
```
ethenotethan, line 32: Nit - does this actually need to be atomic? Iiuc there's isn't a current test case which distributes client load in parallel when switching the version flag

e2e/optimism_test.go
```diff
@@ -215,7 +231,9 @@ func testOptimismGenericCommitment(t *testing.T, v2Enabled bool) {
 
 	optimism.sequencer.ActL2PipelineFull(ot)
```
ethenotethan, line 232: Does this enact the entire derivation pipeline using the first L1 batch or does it just catch up from a previous batch state checkpoint?

e2e/optimism_test.go
```diff
@@ -238,6 +259,24 @@ func testOptimismGenericCommitment(t *testing.T, v2Enabled bool) {
...
+	proxyTS.Server.SetDisperseV2(false)
+
+	ot := actions.NewDefaultTesting(t)
+
+	optimism := NewL2AltDA(ot, proxyTS.Address(), true)
+	exerciseGenericCommitments(t, ot, optimism)
+
+	// turn on v2 dispersal
+	proxyTS.Server.SetDisperseV2(true)
+	exerciseGenericCommitments(t, ot, optimism)
```
ethenotethan, line 279: Is this testing that a migration can be applied against OP which re-derives from an inbox that contains both V1 and V2 certs or is it just applying the migration to do single step update/derive?

e2e/optimism_test.go
samlaf: Wouldnt that just require us to set an arbitrary timestamp at which the setDisperseV2 should happen?

store/manager.go
samlaf: needs to be atomic to prevent race condition from proxy reading it, while test writing to it. Prob innocuous race condition but still better to be race free

e2e/optimism_test.go
ethenotethan: > Wouldnt that just require us to set an arbitrary timestamp at which the setDisperseV2 should happen?

hmm you could introduce scheduling but feel it'd be unnecessary since the rollup itself is the control plane. Maybe for migrating arbitrum we could expose some admin API that the nodes would call when an ArbSysConfig update happens. This might be over-engineering though but definitely something to consider in the future if we introduce versioning with actual breaking changes where proxy has to have stricter switching between version subsections (e.g, if we change how secondary storage keys are generated). This opens a greater question though if there should be some delineation between the DA byte version and the actual EigenDA protocol version. 

Currently we're mapping da byte 1:1 w/ EigenDA protocol but we need to bump this byte in the future irrespective of an EigenDA upgrade!

e2e/optimism_test.go
ethenotethan: I'd argue for guaranteed migration test coverage its absolutely necessary

litt3: I agree, the assertions here just aren't strong enough.

I think it makes sense to create a follow-up PR, I [created an issue](https://linear.app/eigenlabs/issue/DAINT-381/strengthen-proxy-test-assertions-for-v1-vs-v2-operation)

store/manager.go
litt3: Making it atomic also keeps the door open for exposing the flag to users, if there would ever be a use case. Maybe not likely, but the atomicity cost is negligible at the rates we'll see here.

e2e/optimism_test.go
litt3: I'm not sure, I just utilized what was already there for this test. Sam and I tried to go through and understand what precisely was happening in this test, but I didn't feel like we really arrived at a deep level of understanding.

e2e/optimism_test.go
litt3: [See above](https://github.com/Layr-Labs/eigenda-proxy/pull/342#discussion_r2010461549), I'm not super confident in my understanding of what these tests are doing.

PR #344, opened 2025-03-20
Config only

litt3:
- This is a pared down version of [this PR](https://github.com/Layr-Labs/eigenda-proxy/pull/340)

NOTE: I added back a deprecated flags file that was deleted during the merge to support V2, hence the larger diff than before


store/generated_key/memstore/memconfig/config.go
```diff
@@ -17,6 +17,8 @@ type Config struct {
 	// after sleeping PutLatency duration.
 	// This can be used to simulate eigenda being down.
 	PutReturnsFailoverError bool
+	// if true, memstore is enabled and will be used
+	Enabled bool
```
samlaf, line 21: Why are you adding this here? I think we had this outside right?
I originally thought about having it here, but think it can lead to some conflicts. Like what if we add another backend, which also has an enabled flag? I feel like what we want is the external config to be something like:
```
backend: enum(memstore, eigenda, ...)
memstoreconfig{..}
eigendaconfig{..}
otherbackendconfig{..}
```

So feels more natural to me if the backend type is outside of any single one backend config?

testutils/setup.go
```diff
@@ -122,10 +161,10 @@ type TestConfig struct {
 	UseS3Fallback      bool
 }
 
-func NewTestConfig(useMemory bool, useV2 bool) TestConfig {
+func NewTestConfig(backend Backend, disperseToV2 bool) TestConfig {
```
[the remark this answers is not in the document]
litt3: I agree the current state isn't ideal, but there are some reasons I don't think we should go the options route:

1. There are tests where we need direct access the the config struct, to be able to change arbitrary fields. While it would be possible to define a bunch of bespoke methods like `TestSuiteWithEndpoint`, `TestSuiteWithFieldX`, etc., it's more complex than it's worth
2. I'm personally not a fan of the functional options pattern in general
3. If we're gonna put more effort into improving configuration, I'd sooner apply it to implementing your external config proposal

testutils/setup.go
```diff
@@ -161,26 +199,26 @@ func createS3Config(eigendaCfg config.ProxyConfig) config.AppConfig {
...
 		CredentialType:  s3.CredentialTypeStatic,
 	}
-	return config.AppConfig{
-		EigenDAConfig: eigendaCfg,
-	}
+
+	return storeConfig
 }
 
 func BuildTestSuiteConfig(testCfg TestConfig) config.AppConfig {
```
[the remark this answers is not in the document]
litt3: [17b36a4](https://github.com/Layr-Labs/eigenda-proxy/pull/344/commits/17b36a4628e652000148d0587270c820727da08e)

testutils/setup.go
```diff
@@ -135,24 +174,23 @@ func NewTestConfig(useMemory bool, useV2 bool) TestConfig {
...
 	}
-	return config.AppConfig{
-		EigenDAConfig: eigendaCfg,
-	}
+
+	return storageConfig
 }
 
-func createS3Config(eigendaCfg config.ProxyConfig) config.AppConfig {
+func createS3Config(storeConfig store.Config) store.Config {
```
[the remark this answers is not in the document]
litt3: This change would require first implementing [this change](https://github.com/Layr-Labs/eigenda-proxy/pull/344#discussion_r2008331758), which I don't think we should do right now

config/flags.go
```diff
@@ -67,4 +67,8 @@ func init() {
 	Flags = append(Flags, s3.CLIFlags(common.GlobalPrefix, S3Category)...)
 	Flags = append(Flags, memstore.CLIFlags(common.GlobalPrefix, MemstoreFlagsCategory)...)
 	Flags = append(Flags, verify.CLIFlags(common.GlobalPrefix, VerifierCategory)...)
```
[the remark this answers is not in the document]
litt3: [85ca760](https://github.com/Layr-Labs/eigenda-proxy/pull/344/commits/85ca7602df1589b71a675c8b10279237e9fe9287)

I split apart the flag definitions into separate methods. I opted not to do any package reorganization, for the sake of keeping a small diff

config/flags.go
```diff
@@ -67,4 +67,8 @@ func init() {
...
+
+	Flags = append(Flags, eigendaflags.DeprecatedCLIFlags(common.GlobalPrefix, EigenDAClientCategory)...)
+	Flags = append(Flags, verify.DeprecatedCLIFlags(common.GlobalPrefix, VerifierCategory)...)
+	Flags = append(Flags, store.DeprecatedCLIFlags(common.GlobalPrefix, StorageFlagsCategory)...)
```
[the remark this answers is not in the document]
litt3: I'm not sure that's simpler, since it adds an additional layer of indirection between the flag definitions, and where they are being registered.

Maybe we can think more on this, and potentially address in a future PR

config/server_config.go
```diff
@@ -0,0 +1,8 @@
+package config
+
+// ServerConfig ... Config for the proxy HTTP server
+type ServerConfig struct {
+	DisperseToV2 bool
```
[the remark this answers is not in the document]
litt3: This field is being removed from this config in [the PR](https://github.com/Layr-Labs/eigenda-proxy/pull/342) that adds migration tests.

store/generated_key/memstore/memconfig/config.go
litt3: [fe4b76c](https://github.com/Layr-Labs/eigenda-proxy/pull/344/commits/fe4b76c3d5ac5ae938eb886dbc8b7daf770c32cb)

testutils/setup.go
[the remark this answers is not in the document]
samlaf: Right now its a 3 level thing:
```
1. create test config (this acts as an "external config", but created ad-hoc for testing)
modify it
2. create app config from the test config (this is the "internal config")
could still modify this
3. Create the testsuite from 2
```

For 1 I'm saying we merge 1 and 2 using the functional options pattern. I think this is a super simple refactor that would at least clean up our test config.
2. what do you prefer?
3. We could and prob should at some point but that's a much larger project that doesn't only affect tests. But I guess the result is that we get rid of step 2 (instead of getting rid of 1 like I'm proposing), which prob indeed makes more sense, but lots of unknown unknowns here..

config/flags.go
```diff
@@ -25,7 +25,8 @@ const (
 	StorageFlagsCategory    = "Storage"
 	RedisCategory           = "Redis Cache/Fallback"
 	S3Category              = "S3 Cache/Fallback"
-	VerifierCategory        = "KZG and Cert Verifier"
+	VerifierCategory        = "Cert Verifier"
```
samlaf, line 28:
```suggestion
	VerifierCategory        = "Cert Verifier (V1 only)"
```

samlaf: can we add something to show that this is V1 only. I would even be tempted to add to the KZG category that its shared between V1 and V2.

config/flags.go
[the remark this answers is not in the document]
samlaf: oh nvm thought these were new additions you made. Didn't realize it was our old deprecated flags.

docs/help_out.txt
```diff
@@ -18,6 +18,10 @@ GLOBAL OPTIONS:
    --port value   Server listening port (default: 3100) [$EIGENDA_PROXY_PORT]
    --version, -v  print the version
 
+   Cert Verifier
+
+   --eigenda.cert-verification-disabled  Whether to verify certificates received from EigenDA disperser. (default: false) [$EIGENDA_PROXY_EIGENDA_CERT_VERIFICATION_DISABLED]
+
    EigenDA Client
```
samlaf, line 25: Let's add V1 Suffix to both of these categories (EigenDA Client and Cert Verifier)?

docs/help_out.txt
```diff
@@ -33,6 +37,9 @@ GLOBAL OPTIONS:
    --eigenda.disable-tls                                                    Disable TLS for gRPC communication with the EigenDA disperser. Default is false. (default: false) [$EIGENDA_PROXY_EIGENDA_GRPC_DISABLE_TLS]
    --eigenda.disperser-rpc value                                            RPC endpoint of the EigenDA disperser. [$EIGENDA_PROXY_EIGENDA_DISPERSER_RPC]
    --eigenda.eth-rpc value                                                  URL of the Ethereum RPC endpoint. Needed to confirm blobs landed onchain. [$EIGENDA_PROXY_EIGENDA_ETH_RPC]
+   --eigenda.max-blob-length value                                          Maximum blob length to be written or read from EigenDA. Determines the number of SRS points
```
samlaf, line 40: just to be sure, this is just being moved from the KZG category (got mislabeled in previous V2 PR I believe?) to here right? But flag itself isnt changing?

docs/help_out.txt
```diff
@@ -43,34 +50,34 @@ GLOBAL OPTIONS:
...
+   --eigenda.v2.disperse-blob-timeout value      Maximum amount of time to wait for a blob to disperse against v2 protocol. (default: 2m0s) [$EIGENDA_PROXY_EIGENDA_V2_DISPERSE_BLOB_TIMEOUT]
+   --eigenda.v2.disperse-to-v2                   Enable blob dispersal and retrieval against EigenDA V2 protocol. (default: false) [$EIGENDA_PROXY_EIGENDA_V2_DISPERSE_TO_V2]
+   --eigenda.v2.disperser-rpc value              RPC endpoint of the EigenDA disperser. [$EIGENDA_PROXY_EIGENDA_V2_DISPERSER_RPC]
+   --eigenda.v2.eth-rpc value                    URL of the Ethereum RPC endpoint. [$EIGENDA_PROXY_EIGENDA_V2_ETH_RPC]
+   --eigenda.v2.max-blob-length value            Maximum blob length to be written or read from EigenDA. Determines the number of SRS points
+                                                           loaded into memory for KZG commitments. Example units: '30MiB', '4Kb', '30MB'. Maximum size
+                                                           slightly exceeds 1GB. (default: "16MiB") [$EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH]
```
samlaf, line 69: Just flagging this, no need to do anything about it right now.
First of all this file is super helpful.. quite a nice addition!

That being said this part is quite hard to review. I think only 1 or 2 flags changed but the diff being presented is prob doing some stupid line-by-line diff instead of structural diff. Think if we output to a markdown we might be able to get a nicer diff presentation? Not sure.. cc @ethenotethan

testutils/setup.go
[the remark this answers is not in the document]
litt3: > what do you prefer?

Depends on context, but where there are relatively few options, I prefer simply having multiple different methods, that accept different set of parameters. Or a single method, and let the user configure behavior by passing in `nil` for stuff they don't care about.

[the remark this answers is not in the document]
litt3: I'm not opposed to cleanup here in general, just not sure it's worth prioritizing, given that the "ideal" solution would undo the work.

samlaf (dismissed):
Overall LGTM. Can't say that I'm 100% confident, but we need to move forward. The flag output differ is a super nice addition, which'll help us catch external regressions. Also lots of good small improvements in here. I know that we all know this but going forward let's try to keep PRs small, so do feel free to answer my comments by "let's do this in a subsequent PR" like you are doing here :)

config/flags.go
litt3: [4e6ea09](https://github.com/Layr-Labs/eigenda-proxy/pull/344/commits/4e6ea09a7a6f352e6cc80a76ce0cb9aac41b8737)

> I would even be tempted to add to the KZG category that its shared between V1 and V2

I don't think it makes sense to specify this for the KZG category. As it now stands with the commit above, _every_ category that doesn't have v1 or v2 specified is for both versions.

docs/help_out.txt
litt3: [4e6ea09](https://github.com/Layr-Labs/eigenda-proxy/pull/344/commits/4e6ea09a7a6f352e6cc80a76ce0cb9aac41b8737)

docs/help_out.txt
litt3: > just to be sure, this is just being moved from the KZG category

Correct.

> (got mislabeled in previous V2 PR I believe?)

I don't think so- it was just in the verification category all along, and was being propagated through the rest of the config structs in a roundabout way.

docs/help_out.txt
litt3: Yeah, I'm not sure why github is so bad at showing this diff... the tool I use locally handles this in a sane way, and it's easy enough to review.

ethenotethan (approved):
LGTM

PR #349, opened 2025-03-24
Create env var test

litt3:
- closes DAINT-326
- Implements a test for v2 env vars, to mirror the existing v1 env var test

NOTE: since this PR gets rid of the `flags` CI action which is required, I will have to remove that as a required action before merging. Then, I will add the two new flag actions as required once it merges.


scripts/test-proxy-startup-with-env-vars.sh
```diff
@@ -2,24 +2,38 @@
...
 ./bin/eigenda-proxy &
 PID=$!
 
 # Ensure we kill the process on script exit
 trap "kill $PID" EXIT
 
-# Wait 5 seconds for startup to happen
-echo "sleeping 5 seconds to let the proxy start up"
-sleep 5
+# Wait 75 seconds for startup to happen (actual startup can take up to ~50 seconds, due to SRS parsing)
```
ethenotethan, line 34: is it just due to SRS or bootstrapping onchain registry states as well?

.github/workflows/per-pr.yml
```diff
@@ -26,19 +26,29 @@ jobs:
...
-  # This ensures that std output generated when running binary with `--help` is reflected in
-  # in docs/help_out.txt
+  # This checks that the flags in .env.exampleV2.holesky are valid and allow the proxy to start.
+  flagsV2:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION}}
```
samlaf, line 47: use a matrix strategy instead: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy

scripts/test-proxy-startup-with-env-vars.sh
```diff
@@ -2,24 +2,38 @@
...
-# If any deprecated flags are still being used in .env.example.holesky, the script will fail.
+# If any deprecated flags are still being used in the specified environment file, the script will fail.
+
+ENV_FILE=$1
+
+# Check if the environment file exists
+if [ ! -f "$ENV_FILE" ]; then
+    echo "Error: Environment file $ENV_FILE does not exist"
+    echo "Current working directory: $(pwd)"
+    echo "Files in current directory:"
```
samlaf, line 15: prob need to add a check in case someone forgets to pass argument. Then exit with usage comment

scripts/test-proxy-startup-with-env-vars.sh
samlaf, line 34: wtf...? is this new? what changed?

litt3: A couple things:
1. g2 srs parsing takes close to 30 seconds, and previously wasn't happening at all (since it wasn't required for v1)
2. with v2 enabled, we are instantiating two backends instead of one, so that adds some time too

.github/workflows/per-pr.yml
litt3: [eb93cad](https://github.com/Layr-Labs/eigenda-proxy/pull/349/commits/eb93cad30057e88ea3b6cc5d47fa0c89c52ccb9e)

scripts/test-proxy-startup-with-env-vars.sh
litt3: [3387be2](https://github.com/Layr-Labs/eigenda-proxy/pull/349/commits/3387be2c8f0f794223723cbead35a8a835d5f5f9)

samlaf (dismissed):

PR #350, opened 2025-03-24
chore(clients): retract v1.0.1

samlaf:
We published a `clients/v1.0.1` tag by mistake, which got picked up by pkg.go.dev, and now the latest version of the docs (pointed to in our README) leads to https://pkg.go.dev/github.com/Layr-Labs/eigenda-proxy/clients@v1.0.1/standard_client. This directive retracts that version, such that the docs (and go commands) no longer point to this version (at this point v0.2.0 is the latest). 
Once we reach semver v1.0.1, we can possibly remove this retract directive (??), or just skip this number and go to v1.0.2.

After this is merged, will create a new release v0.2.1 of the clients such that this change gets picked up by pkg.go.dev and our new docs (including runnable test examples) show up there.

## Fixes Issue

Fixes #

### DIFF
--- a/clients/go.mod
+++ b/clients/go.mod
@@ -63,3 +63,12 @@ require (
 	golang.org/x/sys v0.28.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+// See https://go.dev/ref/mod#go-mod-file-retract for more information about the retract directive.
+retract (
+	// We published a `clients/v1.0.1` tag by mistake, which got picked up by pkg.go.dev, and now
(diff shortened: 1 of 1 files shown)


clients/go.mod
```diff
@@ -63,3 +63,12 @@ require (
 	golang.org/x/sys v0.28.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+// See https://go.dev/ref/mod#go-mod-file-retract for more information about the retract directive.
+retract (
```
ethenotethan, line 68: oh I didn't even know that retraction was something that was possible - this is cool!

ethenotethan (approved):
LGTM!

samlaf:
Merging without preprod/testnet CI passing as we are in the process of refactoring preprod tests and testnet is down.

PR #351, opened 2025-03-24
Fix e2e tests

litt3:
- closes DAINT-397 https://linear.app/eigenlabs/issue/DAINT-397/fix-non-local-e2e-tests
- updates to the most recent core DA commit

### DIFF
--- a/store/manager.go
+++ b/store/manager.go
@@ -94,13 +94,15 @@ func (m *Manager) Get(ctx context.Context, key []byte, cm commitments.Commitment
 			return nil, errors.New("expected EigenDA backend for DA commitment type, but none configured")
 		}
 
-		var err error
-		var data []byte
+		verifyMethod, err := m.getVerifyMethod(cm.Version)
+		if err != nil {
+			return nil, fmt.Errorf("get verify method: %w", err)
+		}
 
 		// 1 - read blob from cache if enabled
 		if m.secondary.CachingEnabled() {
 			m.log.Debug("Retrieving data from cached backends")
-			data, err := m.secondary.MultiSourceRead(ctx, key, false, m.eigenda.Verify)
+			data, err := m.secondary.MultiSourceRead(ctx, key, false, verifyMethod)
 			if err == nil {
 				return data, nil
 			}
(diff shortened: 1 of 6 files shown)


store/manager.go
```diff
@@ -94,13 +94,15 @@ func (m *Manager) Get(ctx context.Context, key []byte, cm commitments.Commitment
...
+		verifyMethod, err := m.getVerifyMethod(cm.Version)
+		if err != nil {
+			return nil, fmt.Errorf("get verify method: %w", err)
+		}
 
 		// 1 - read blob from cache if enabled
 		if m.secondary.CachingEnabled() {
 			m.log.Debug("Retrieving data from cached backends")
-			data, err := m.secondary.MultiSourceRead(ctx, key, false, m.eigenda.Verify)
+			data, err := m.secondary.MultiSourceRead(ctx, key, false, verifyMethod)
```
samlaf, line 105: does this change not also need to be applied for the fallback? See https://github.com/Layr-Labs/eigenda-proxy/blob/4602d9ab31d02605761eee416cc9cc8d46fda21a/store/manager.go#L123

samlaf (approved):
LGTM also seems like this should also be applied to fallback?

Overall seems a bit unfortunate the way the code is structured since we need to add this custom verification everywhere. A nicer structure would have been to decouple reading from verification. And then we could have read (from either cache, main store, fallback), and only after that do verification in a single place. Guessing this refactor is something that's hard to do without changing a thousand files though?

store/manager.go
litt3: I had made the fix there too, but I must have managed to undo it... good catch, thanks. [aa0f951](https://github.com/Layr-Labs/eigenda-proxy/pull/351/commits/aa0f951dff22e674cbba738ec306acdc97c82165)

samlaf (dismissed):

samlaf (approved):
LGTM. iiuc after our conversation tests are passing locally but failing in CI for some reason. So goal is to merge this, and then experiment with fixing CI in a new PR.

PR #353, opened 2025-03-24
docs(readme): document 1account/proxy

### LINKED ISSUE #250: Support multiple accountIds for RaaS support
This is a feature suggestion coming from us. Hasn't been requested (yet) by any RaaS. Just opening as a tracker and for external visibility.

Right now eigenda-proxy uses a single eigenDAClient, which implies they can only have a single account_id loaded. This means all of the rollups they support are (for v1) being rate limited (and for v2) paying from the same account_id.

Not sure whether supporting multiple account_ids is needed or not, we would need to better understand how RaaSs' setup works. Perhaps we could hold a mapping from ip to account_id which they could load as an initial config?

samlaf:
## Fixes Issue

Closes #250 
Closes DAINT-246

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,8 @@ Features:
 - [Contributor Guide](#contributor-guide)
   - [Testing](#testing)
     - [Unit](#unit)
-    - [Integration / E2E against local deployment](#integration--e2e-against-local-deployment)
-    - [E2E against Holesky](#e2e-against-holesky)
-    - [E2E Fuzz](#e2e-fuzz)
+    - [Integration / E2E](#integration--e2e)
+    - [Fuzz](#fuzz)
 
 
 ## User Guide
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -202,6 +201,8 @@ This behavior is turned on by default, but configurable via the `--eigenda.confi
 
 In order to disperse to the EigenDA V1 network in production, or at high throughput on testnet, please register your authentication ethereum address through [this form](https://forms.gle/3QRNTYhSMacVFNcU8). Your EigenDA authentication keypair address should not be associated with any funds anywhere. For EigenDA V2, please see our [payments](https://docs.eigenda.xyz/releases/payments) doc.
 
+> Note: Proxy only supports using a single authn (v1) or payment (v2) key. We recommend running a single instance of the Proxy per Rollup.
```
ethenotethan, line 204:
```suggestion
> Note: Proxy only supports using a single authorization (v1) or payment (v2) key. We recommend running a single instance of the Proxy per Rollup (i.e, 1 proxy : 1 rollup) and avoid provisioning patterns like N rollups : 1 proxy.
```

ethenotethan (approved):
LGTM

README.md
samlaf: Changed it to [505f8bd](https://github.com/Layr-Labs/eigenda-proxy/pull/353/commits/505f8bdde25ca138f3dd6b08fc9588f6ef3c078f)

PR #358, opened 2025-04-02
Get preprod CI passing

litt3:
- Configure `PutRetries` to be 3 for both v1 and v2
    - previously was undefined for v1, which could cause tests to hang if dispersal kept failing
    - chose 3 for both v1 and v2 since it seems a reasonable value
- Increase parallelism, to speed up e2e tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,14 +38,14 @@ test-e2e-local:
 # E2E tests using holesky testnet backend, leveraging op-e2e framework. Also tests the standard client against the proxy.
 # If holesky tests are failing, consider checking https://dora.holesky.ethpandaops.io/epochs for block production status.
 test-e2e-testnet:
-	# Add the -v flag to observe logs as the run is happening on CI, given that this test takes ~40 minutes to run.
+	# Add the -v flag to observe logs as the run is happening on CI, given that this test takes ~20 minutes to run.
 	# Good to have early feedback when needed.
-	BACKEND=testnet go test -v -timeout 60m ./e2e -parallel 4
+	BACKEND=testnet go test -v -timeout 30m ./e2e -parallel 8
 
 test-e2e-preprod:
-	# Add the -v flag to observe logs as the run is happening on CI, given that this test takes ~40 minutes to run.
+	# Add the -v flag to observe logs as the run is happening on CI, given that this test takes ~20 minutes to run.
 	# Good to have early feedback when needed.
-	BACKEND=preprod go test -v -timeout 60m ./e2e -parallel 4
+	BACKEND=preprod go test -v -timeout 30m ./e2e -parallel 8
 
 # Very simple fuzzer which generates random bytes arrays and sends them to the proxy using the standard client.
 # To clean the cached corpus, run `go clean -fuzzcache` before running this.
--- a/testutils/setup.go
+++ b/testutils/setup.go
@@ -278,6 +278,7 @@ func BuildTestSuiteConfig(testCfg TestConfig) config.AppConfig {
 				SvcManagerAddr:           svcManagerAddress,
 			},
 			MaxBlobSizeBytes: maxBlobLengthBytes,
+			PutRetries:       3,
 		},
 		VerifierConfigV1: verify.Config{
 			VerifyCerts:          false,
@@ -320,7 +321,7 @@ func BuildTestSuiteConfig(testCfg TestConfig) config.AppConfig {
 				PayloadClientConfig: payloadClientConfig,
 				RelayTimeout:        5 * time.Second,
 			},
-			PutRetries:                 1,
+			PutRetries:                 3,
 			MaxBlobSizeBytes:           maxBlobLengthBytes,
 			EigenDACertVerifierAddress: certVerifierAddress,
 		},
@@ -340,15 +341,12 @@ func BuildTestSuiteConfig(testCfg TestConfig) config.AppConfig {
 	switch {
 	case testCfg.UseKeccak256ModeS3:
 		proxyConfig.StorageConfig = createS3Config(proxyConfig.StorageConfig)
-
 	case testCfg.UseS3Caching:
 		proxyConfig.StorageConfig.CacheTargets = []string{"S3"}
 		proxyConfig.StorageConfig = createS3Config(proxyConfig.StorageConfig)
-
 	case testCfg.UseS3Fallback:
 		proxyConfig.StorageConfig.FallbackTargets = []string{"S3"}
 		proxyConfig.StorageConfig = createS3Config(proxyConfig.StorageConfig)
-
 	case testCfg.UseRedisCaching:
 		proxyConfig.StorageConfig.CacheTargets = []string{"redis"}
 		proxyConfig.StorageConfig = createRedisConfig(proxyConfig.StorageConfig)


testutils/setup.go
```diff
@@ -234,7 +233,11 @@ func BuildTestSuiteConfig(testCfg TestConfig) config.AppConfig {
 		pollInterval = time.Minute * 1
 	}
 
-	maxBlobLengthBytes, err := common.ParseBytesAmount("16mib")
+	maxBlobLength := testCfg.MaxBlobLength
+	if maxBlobLength == "" {
+		maxBlobLength = "1mib"
```
anupsv, line 238: Should we randomize this ? Or can we add a test where we check if the 16mb blob also is fine ?

anupsv (approved):

.env.exampleV1AndV2.holesky
```diff
@@ -86,7 +86,7 @@ EIGENDA_PROXY_EIGENDA_V2_RELAY_TIMEOUT=5s
 EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 
 # Maximum blob length for V2
-EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=16MiB
+EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
```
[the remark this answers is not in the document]
litt3: it speeds up the flag tests considerably

.github/workflows/per-pr.yml
```diff
@@ -99,6 +105,8 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: ${{ env.GO_VERSION }}
+      - name: Install gotestsum
```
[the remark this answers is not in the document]
litt3: It makes the test output actually legible- previous test output is entirely useless, since logs from all concurrent tests are mixed together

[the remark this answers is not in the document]
ethenotethan: should there be an install target within Make for this?

.env.exampleV1AndV2.holesky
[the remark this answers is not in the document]
ethenotethan: understood - is it worth commenting this though since this example could be used by potential customers in the future? Also the existing docstring `Maximum blob length for V2` doesn't hold true anymore

testutils/setup.go
ethenotethan: @anupsv pretty sure this is being done to reduce test times by decreasing SRS ingestion sizes

.env.exampleV1AndV2.holesky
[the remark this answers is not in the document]
litt3: I added a description in [5cb0844](https://github.com/Layr-Labs/eigenda-proxy/pull/358/commits/5cb0844eea5af8a2ca2e5537529db6187d1ad243) to clarify

testutils/setup.go
litt3: > Should we randomize this

I don't see a benefit in randomizing, this just needs to be larger than the maximum length of the largest blob dispersal attempted.

> Or can we add a test where we check if the 16mb blob also is fine ?

Sure, I'll add a test for this

litt3: [20ea92f](https://github.com/Layr-Labs/eigenda-proxy/pull/358/commits/20ea92ffadb293c876f903b5e51524958e8e166a)

Makefile
```diff
@@ -89,6 +89,9 @@ op-devnet-allocs:
 benchmark:
 	go test -benchmem -run=^$ -bench . ./benchmark -test.parallel 4
 
+install-gotestsum:
```
samlaf, line 92: should just be called `deps`, and it should install all the requires dependencies/tooling that we need. My goal is to eventually use nix or mise for this, but in the meantime we can use this and add all the dependencies needed to run tests or develop locally.

.env.exampleV1AndV2.holesky
```diff
@@ -26,8 +26,12 @@ EIGENDA_PROXY_EIGENDA_DISABLE_POINT_VERIFICATION_MODE=false
 # Disable TLS for gRPC communication with the EigenDA disperser (V1)
 EIGENDA_PROXY_EIGENDA_GRPC_DISABLE_TLS=false
 
-# Maximum blob length for V1
-EIGENDA_PROXY_EIGENDA_MAX_BLOB_LENGTH=16MiB
+# Maximum blob length for V1. This value is permitted to be up to 16MiB
+# This value is configured to be lower than the permitted maximum, since this configuration determines the amount of
+# SRS data that will be read into memory. Reading in a large amount of SRS data can be expensive, and since you only
```
samlaf, line 31:
```suggestion
# SRS data that will be read into memory. Reading in a large amount of cause long startup times, and since you only
```

.env.exampleV2.holesky
```diff
@@ -44,8 +44,12 @@ EIGENDA_PROXY_EIGENDA_V2_RELAY_TIMEOUT=5s
 # Blob params version used when dispersing
 EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 
-# Maximum blob length for V2
-EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=16MiB
+# Maximum blob length for V2. This value is permitted to be up to 16MiB
+# This value is configured to be lower than the permitted maximum, since this configuration determines the amount of
+# SRS data that will be read into memory. Reading in a large amount of SRS data can be expensive, and since you only
```
samlaf, line 49:
```suggestion
# SRS data that will be read into memory. Reading in a large amount of SRS data can cause long startup time, and since you only
```

scripts/test-proxy-startup-with-env-vars.sh
```diff
@@ -38,9 +38,9 @@ PID=$!
 # Ensure we kill the process on script exit
 trap "kill $PID" EXIT
 
-# Wait 75 seconds for startup to happen (actual startup can take up to ~50 seconds, due to SRS parsing)
-echo "sleeping 75 seconds to let the proxy start up"
-sleep 75
+# Wait 10 seconds for startup to happen (actual startup takes ~5 seconds)
```
samlaf, line 41:
```suggestion
# Wait 10 seconds for startup to happen (actual startup takes ~5 seconds with EIGENDA_PROXY_EIGENDA_MAX_BLOB_LENGTH=1MiB)
```

testutils/setup.go
```diff
@@ -278,6 +281,7 @@ func BuildTestSuiteConfig(testCfg TestConfig) config.AppConfig {
 				SvcManagerAddr:           svcManagerAddress,
 			},
 			MaxBlobSizeBytes: maxBlobLengthBytes,
+			PutRetries:       3,
```
samlaf, line 284: is the default behavior when `PutRetries=0` that it tries forever, which was causing tests to hang? We should probably change that behavior. IMO zero values (which are the default when forgotten) should always be safe (in this case should probably mean try once, or we insead just dissalow value 0). Unbounded behavior could for example be when `PutRetries=-1` instead.

.env.exampleV1AndV2.holesky
litt3: [748a9fd](https://github.com/Layr-Labs/eigenda-proxy/pull/358/commits/748a9fdf0143d91faaff31fa4e136e11a0b373c7)

.env.exampleV2.holesky
litt3: [748a9fd](https://github.com/Layr-Labs/eigenda-proxy/pull/358/commits/748a9fdf0143d91faaff31fa4e136e11a0b373c7)

scripts/test-proxy-startup-with-env-vars.sh
litt3: [bdb4ebf](https://github.com/Layr-Labs/eigenda-proxy/pull/358/commits/bdb4ebf593194853b93f884901e072629cbc158e)

Added, but without naming a specific config field, since max blob length has multiple fields that can configure it

testutils/setup.go
litt3: Will make the change in a follow up pr, see https://linear.app/eigenlabs/issue/DAINT-419/make-putretries-configuration-more-user-friendly

samlaf (approved):

PR #359, opened 2025-04-06
chore(make): rename build command

samlaf:
Used to be "eigenda-proxy". Renamed for consistency with `docker-build` command. Also most people are used to just typing `make build`, which doesn't require knowing the binary name.

## Fixes Issue

Fixes #


ethenotethan (approved):
LGTM

PR #360, opened 2025-04-07
Implement admin endpoints for v2 dispersal

litt3:
- closes DAINT-420
    -  [link](https://linear.app/eigenlabs/issue/DAINT-420/create-admin-http-endpoints-to-get-and-set-dispersetov2)
- closes DAINT-389
    - [link](https://linear.app/eigenlabs/issue/DAINT-389/docs-document-v2-migration)

## Flag changes
- adds new flag 
```
		&cli.BoolFlag{
			Name:     AdminEndpointsEnabledFlagName,
			Usage:    "Enable administrative HTTP endpoints for runtime configuration",
			Value:    false,
			EnvVars:  common.PrefixEnvVar(envPrefix, "ADMIN_ENDPOINTS_ENABLED"),
			Category: category,
		},
```

### DIFF
--- a/.env.exampleV1AndV2.holesky
+++ b/.env.exampleV1AndV2.holesky
@@ -102,10 +102,10 @@ EIGENDA_PROXY_STORAGE_BACKENDS_TO_ENABLE=V1,V2
 
 # Enable blob dispersal and retrieval against EigenDA V2 protocol
 # For migration: Start with 'false' and toggle at runtime using admin endpoints
-EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=false
+EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=true
 
 # Enable administrative HTTP endpoints for runtime configuration (needed for on-the-fly migration)
-EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED=true
+EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED=false
 
 # === Shared KZG Configuration ===
 
--- a/.env.exampleV1AndV2.holesky
+++ b/.env.exampleV1AndV2.holesky
@@ -105,7 +105,8 @@ EIGENDA_PROXY_STORAGE_BACKENDS_TO_ENABLE=V1,V2
 EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=true
 
 # Enable administrative HTTP endpoints for runtime configuration (needed for on-the-fly migration)
-EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED=false
+# Set to "admin" to enable admin endpoints, or leave empty to disable
+EIGENDA_PROXY_API_ENABLED=
 
 # === Shared KZG Configuration ===
 
--- a/.env.exampleV1AndV2.holesky
+++ b/.env.exampleV1AndV2.holesky
@@ -101,7 +101,11 @@ EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 EIGENDA_PROXY_STORAGE_BACKENDS_TO_ENABLE=V1,V2
 
 # Enable blob dispersal and retrieval against EigenDA V2 protocol
-EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=true
+# For migration: Start with 'false' and toggle at runtime using admin endpoints
+EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=false
+
+# Enable administrative HTTP endpoints for runtime configuration (needed for on-the-fly migration)
+EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED=true
 
 # === Shared KZG Configuration ===
 
--- a/README.md
+++ b/README.md
@@ -126,6 +126,38 @@ Response:
   Body: <preimage_bytes>
 ```
 
+#### Admin Routes
+
+The proxy provides administrative endpoints to control runtime behavior. By default, these endpoints are disabled and must be explicitly enabled through configuration.
+
+To enable admin endpoints, set the `--admin-endpoints-enabled` flag to `true` or set the environment variable `EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED=true` when starting the proxy server.
+
+When enabled, the following admin endpoints are available:
+
+```text
+Request:
+  GET /admin/v2-dispersal
+
+Response:
+  200 OK
+  Content-Type: application/json
+  Body: {"disperseToV2": boolean}
+```
+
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -126,6 +130,84 @@ Response:
...
+When enabled, the following admin endpoints are available:
+
+```text
+Request:
+  GET /admin/v2-dispersal
+
+Response:
+  200 OK
+  Content-Type: application/json
+  Body: {"disperseToV2": boolean}
```
ethenotethan, line 148: longterm I'm curious if there's any value in changing this to something like `protocolVersion`

README.md
```diff
@@ -126,6 +130,84 @@ Response:
   Body: <preimage_bytes>
 ```
 
+#### Admin Routes
+
+The proxy provides administrative endpoints to control runtime behavior. By default, these endpoints are disabled and must be explicitly enabled through configuration.
```
ethenotethan, line 135: let's add a disclaimer specifying that this should never be publicly accessible - then again the namespace `/admin` already infers that

config/flags.go
```diff
@@ -30,8 +30,9 @@ const (
 )
 
 const (
-	ListenAddrFlagName = "addr"
-	PortFlagName       = "port"
+	ListenAddrFlagName            = "addr"
+	PortFlagName                  = "port"
+	AdminEndpointsEnabledFlagName = "admin-endpoints-enabled"
```
ethenotethan, line 35: probably not worth but curious if there's any value in putting the admin API on a same port space to allow operators to better manage network security ingress by setting per port policies

server/routing.go
```diff
@@ -85,6 +85,13 @@ func (svr *Server) RegisterRoutes(r *mux.Router) {
 	)
 
 	r.HandleFunc("/health", withLogging(svr.handleHealth, svr.log)).Methods("GET")
+
+	// Only register admin endpoints if explicitly enabled in configuration
+	if svr.config.AdminEndpointsEnabled {
```
ethenotethan, line 90: in the vain of using dynamic lists for configuration - curious if there's any value if instead taking a list of enabled "APIs"; e.g:
```
--api-enabled proxy, memconfig, admin 
```

analogous to how geth does ([e.g](https://github.com/offchainlabs/nitro-testnode/blob/release/docker-compose.yaml#L69))

config/flags.go
litt3: > on a same port space

Do you mean put it on a *different* port?

I think changes in this vein could be worth considering in the future, if driven by customer requirements.

README.md
litt3: [c7a17b9](https://github.com/Layr-Labs/eigenda-proxy/pull/360/commits/c7a17b9ccaea6e5c199aa8fd97d9c607dfc911fc)

server/routing.go
litt3: Given that it's a breaking change to how memstore is configured, and based on the complexity of this change, I don't think we should do this right now

.env.exampleV1AndV2.holesky
```diff
@@ -101,7 +101,11 @@ EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 EIGENDA_PROXY_STORAGE_BACKENDS_TO_ENABLE=V1,V2
 
 # Enable blob dispersal and retrieval against EigenDA V2 protocol
-EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=true
+# For migration: Start with 'false' and toggle at runtime using admin endpoints
+EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=false
+
+# Enable administrative HTTP endpoints for runtime configuration (needed for on-the-fly migration)
+EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED=true
```
samlaf, line 108:
```suggestion
EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2=true

# Enable administrative HTTP endpoints for runtime configuration (needed for on-the-fly migration)
EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED=false
```

samlaf: Feel like these example env vars should be production-ready, in case some users just copy them and run their service with them.

docs/help_out.txt
```diff
@@ -13,10 +13,11 @@ COMMANDS:
...
 
 GLOBAL OPTIONS:
-   --addr value   Server listening address (default: "0.0.0.0") [$EIGENDA_PROXY_ADDR]
-   --help, -h     show help
-   --port value   Server listening port (default: 3100) [$EIGENDA_PROXY_PORT]
-   --version, -v  print the version
+   --addr value               Server listening address (default: "0.0.0.0") [$EIGENDA_PROXY_ADDR]
+   --admin-endpoints-enabled  Enable administrative HTTP endpoints for runtime configuration (default: false) [$EIGENDA_PROXY_ADMIN_ENDPOINTS_ENABLED]
+   --help, -h                 show help
+   --port value               Server listening port (default: 3100) [$EIGENDA_PROXY_PORT]
```
samlaf, line 19: Feel like these should be in a "server" category?

server/routing.go
samlaf: Don't think its a breaking change though given that proxy is auto-enabled, and memconfig api is also auto-enabled (afair) when --memstore.enabled is passed.
So this would just be a future-compatible way to enable apis (we could eventually enable the memstore api that way as well (also I think breaking changes for testing apis are totally fine)

samlaf (commented):
Very nice PR

.env.exampleV1AndV2.holesky
litt3: [d7a1ff3](https://github.com/Layr-Labs/eigenda-proxy/pull/360/commits/d7a1ff310381340a3ef08246212b55cd99ace1b5)

samlaf (dismissed):
LGTM

server/routing.go
litt3: > we could eventually enable the memstore api that way as well

Fair point, when considering complexity and impact, I was assuming we would change how memstore works right now. But there isn't any reason we need to do that, we can simply use this new pattern for the admin API

With that in mind, I'll go ahead and implement this suggestion for the admin endpoints

litt3: [714f107](https://github.com/Layr-Labs/eigenda-proxy/pull/360/commits/714f10781b2d380cef2d133c401bba1319a24d24)

config/flags.go
ethenotethan: makes sense - feel free to resolve

ethenotethan (approved):
LGTM - only thing missing imo is an E2E test which asserts the get/set logic. pretty out-of-scope though and may require custom client work

PR #361, opened 2025-04-08
chore: configure dispersal backend flag with enum

litt3:
- This PR uses an enum to determine dispersal backend, rather than using a `disperseToV2` bool
- This is for the sake of consistency, and for potentially supporting future versions

## Flag Changes

Changes bool flag

```
		&cli.BoolFlag{
			Name:     DisperseToV2FlagName,
			Usage:    "Enable blob dispersal to EigenDA V2 protocol.",
			EnvVars:  withEnvPrefix(envPrefix, "DISPERSE_TO_V2"),
```

to string flag, that uses the `EigenDABackend` enum

```
		&cli.StringFlag{
			Name:     DispersalBackendFlagName,
			Usage:    "Target EigenDA backend version for blob dispersal (e.g. V1 or V2).",
			EnvVars:  withEnvPrefix(envPrefix, "DISPERSAL_BACKEND"),
```

Note that no V2 release has been cut yet so this change does not constitute a breaking change.


docs/help_out.txt
```diff
@@ -124,6 +124,6 @@ GLOBAL OPTIONS:
    --storage.backends-to-enable value [ --storage.backends-to-enable value ]  Comma separated list of eigenDA backends to enable (e.g. V1,V2) (default: "V1") [$EIGENDA_PROXY_STORAGE_BACKENDS_TO_ENABLE]
    --storage.cache-targets value [ --storage.cache-targets value ]            List of caching targets to use fast reads from EigenDA. [$EIGENDA_PROXY_STORAGE_CACHE_TARGETS]
    --storage.concurrent-write-routines value                                  Number of threads spun-up for async secondary storage insertions. (<=0) denotes single threaded insertions where (>0) indicates decoupled writes. (default: 0) [$EIGENDA_PROXY_STORAGE_CONCURRENT_WRITE_THREADS]
-   --storage.disperse-to-v2                                                   Enable blob dispersal to EigenDA V2 protocol. (default: false) [$EIGENDA_PROXY_STORAGE_DISPERSE_TO_V2]
+   --storage.dispersal-backend value                                          Target EigenDA backend version for blob dispersal (e.g. V1 or V2). (default: "V1") [$EIGENDA_PROXY_STORAGE_DISPERSAL_BACKEND]
```
samlaf, line 127: I don't like "(e.g. V1 or V2)"
I think (?) there's a way to use an enum for allowed values in urfave and urfave would automatically list all the options here. Don't spend too much time on this but if its easy would be a nice future-compatible change

e2e/safety_checks_test.go
```diff
@@ -20,19 +20,19 @@ func isNilPtrDerefPanic(err string) bool {
 }
 
 func TestOpClientKeccak256MalformedInputsV1(t *testing.T) {
-	testOpClientKeccak256MalformedInputs(t, false)
+	testOpClientKeccak256MalformedInputs(t, common.V1EigenDABackend)
 }
 
 func TestOpClientKeccak256MalformedInputsV2(t *testing.T) {
-	testOpClientKeccak256MalformedInputs(t, true)
+	testOpClientKeccak256MalformedInputs(t, common.V2EigenDABackend)
```
samlaf, line 27: yay this is already way more readable than a random boolean

store/cli.go
```diff
@@ -38,12 +38,13 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
...
-			Usage:    "Enable blob dispersal to EigenDA V2 protocol.",
-			EnvVars:  withEnvPrefix(envPrefix, "DISPERSE_TO_V2"),
+		&cli.StringFlag{
+			Name:     DispersalBackendFlagName,
+			Usage:    "Target EigenDA backend version for blob dispersal (e.g. V1 or V2).",
+			EnvVars:  withEnvPrefix(envPrefix, "DISPERSAL_BACKEND"),
 			Category: category,
 			Required: false,
+			Value:    "V1",
 		},
```
samlaf, line 48: using a generic flag to implement the actual enum would be nice: see https://github.com/urfave/cli/issues/602#issuecomment-416000118

samlaf (approved):
Nice refactor. Fine to me as is although I do think if you had time using an actual enum for the flag values would be even better (prob something we should use in other places as well)

docs/help_out.txt
litt3: I took a crack at this:

1. It's kinda ugly (if you're curious to see, I pushed it to a branch called `cli-enums`
2. I was struggling to get the help output to generate using the generic type

Given our current time crunch, I think it's best to more forward as is, since it's not proven super easy to implement, and I'm not yet 100% convinced it's actually better, considering the added complexity.

store/cli.go
litt3: See [this comment](https://github.com/Layr-Labs/eigenda-proxy/pull/361#discussion_r2035783143)

docs/help_out.txt
samlaf: LGTM def not a priority. Pretty sure op uses this in a few places, so we could also look at that for other examples if/when we decide this is worth doing (never is fine lol)

samlaf (approved):

PR #362, opened 2025-04-08
Add explanatory docs and logging for admin endpoints

litt3:
- Addresses comments left on https://github.com/Layr-Labs/eigenda-proxy/pull/360

### DIFF
--- a/server/routing.go
+++ b/server/routing.go
@@ -88,10 +88,16 @@ func (svr *Server) RegisterRoutes(r *mux.Router) {
 	r.HandleFunc("/health", withLogging(svr.handleHealth, svr.log)).Methods("GET")
 
 	// Only register admin endpoints if explicitly enabled in configuration
+	// Note: A common pattern for admin endpoints is to generate a random API key on startup for authentication.
+	// Since the proxy isn't meant to be exposed publicly, we haven't implemented this here, but it's something
+	// that might be done in the future.
 	if svr.config.IsAPIEnabled(config.AdminAPIType) {
+		svr.log.Warn("Admin API endpoints are enabled")
 		// Admin endpoints to check and set EigenDA backend used for dispersal
-		r.HandleFunc("/admin/eigenda-dispersal-backend", withLogging(svr.handleGetEigenDADispersalBackend, svr.log)).Methods("GET")
-		r.HandleFunc("/admin/eigenda-dispersal-backend", withLogging(svr.handleSetEigenDADispersalBackend, svr.log)).Methods("PUT")
+		r.HandleFunc("/admin/eigenda-dispersal-backend",
+			withLogging(svr.handleGetEigenDADispersalBackend, svr.log)).Methods("GET")
+		r.HandleFunc("/admin/eigenda-dispersal-backend",
+			withLogging(svr.handleSetEigenDADispersalBackend, svr.log)).Methods("PUT")
 	}
 }
 


server/routing.go
```diff
@@ -88,10 +88,16 @@ func (svr *Server) RegisterRoutes(r *mux.Router) {
 	r.HandleFunc("/health", withLogging(svr.handleHealth, svr.log)).Methods("GET")
 
 	// Only register admin endpoints if explicitly enabled in configuration
+	// Note: A common pattern for admin endpoints is to generate a random API key on startup for authentication.
```
samlaf, line 91: Is this really a thing? how does it work? how is the random api key used by anyone if its randomly generated?

samlaf (dismissed):

server/routing.go
litt3: This note reflects the comment from @anupsv [here](https://github.com/Layr-Labs/eigenda-proxy/pull/360#discussion_r2033540286), I'm not really familiar with how people "normally" do this to be able to opine

> literally remove the admin API from production binaries.

How would one go about doing this?

server/routing.go
```diff
@@ -88,10 +88,16 @@ func (svr *Server) RegisterRoutes(r *mux.Router) {
...
+	// Since the proxy isn't meant to be exposed publicly, we haven't implemented this here, but it's something
+	// that might be done in the future.
```
ethenotethan, line 93: also worth noting that we expose no telemetry for these endpoints meaning that behavior is only observable via log dumps

server/routing.go
samlaf: @anupsv can you comment? I don't understand how generating a random key makes sense.

> How would one go about doing this?

you put those routes in a separate file with a build tag, and that build tag (`go build -tags prod`) is only passed when building releases. See https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags#adding-build-tags for example

litt3: > literally remove the admin API from production binaries

@samlaf In this case, there are definitely production use-cases where you'd want these endpoints enabled. Customers who use proxy the "private" way may choose to do even a mainnet upgrade to v2 via the admin endpoints.

anupsv: > @anupsv can you comment? I don't understand how generating a random key makes sense.
> 
> > How would one go about doing this?
> 
> you put those routes in a separate file with a build tag, and that build tag (`go build -tags prod`) is only passed when building releases. See https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags#adding-build-tags for example

Sure, essentially it's another step to help prevent accidental exposing and usage of the admin endpoint. So with this random key which is generated, logged in console, even in the case that this is exposed, there's an api key at least preventing anyone random from calling it.

server/routing.go
samlaf: Addition sounds wrong to me. Its not only observable through log analysis. You can literally just call GET on the endpoint to know what the current state is. We could/should add prometheus instrumentation to track the dispersal backend over time, but that can prob just live in a ticket

samlaf (approved):

server/routing.go
litt3: > You can literally just call GET on the endpoint to know what the current state is.

Valid point, removed the comment [d26a186](https://github.com/Layr-Labs/eigenda-proxy/pull/362/commits/d26a1866f7964cb821d62559124671ee3e956122)

PR #365, opened 2025-04-09
chore!: Clean up put retries

litt3:
- closes DAINT-419
    - [link](https://linear.app/eigenlabs/issue/DAINT-419/make-putretries-configuration-more-user-friendly) 

## Flag changes

- `PUT_RETRIES` for both v1 and v2 flags is now an `int` instead of a `uint`, to support more intuitive configuration (old logic interprets 0 to mean retry forever)
```
		&cli.IntFlag{
			Name: PutRetriesFlagName,
			Usage: "Total number of times to try blob dispersals before serving an error response." +
				">0 = try dispersal that many times. <0 = retry indefinitely. 0 is not permitted (causes startup error).",
			Value:    3,
			EnvVars:  []string{withEnvPrefix(envPrefix, "PUT_RETRIES")},
			Category: category,
		},
```
- This is *technically* a breaking change, in case anyone was relying on the old behavior where config value `0` yields infinite retries. Practically, my guess would be that no one is relying on this behavior, but it's worth noting

### DIFF
--- a/common/client_config_v1.go
+++ b/common/client_config_v1.go
@@ -13,5 +13,9 @@ import (
 type ClientConfigV1 struct {
 	EdaClientCfg     clients.EigenDAClientConfig
 	MaxBlobSizeBytes uint64
-	PutRetries       uint
+	// Determines number of times to try blob dispersals:
+	// - If > 0: Try up to that many times total (first attempt + up to N-1 retries)
+	// - If = 0: Try only once (no retries)
+	// - If < 0: Try indefinitely until success
+	PutRetries int
 }
--- a/common/client_config_v1.go
+++ b/common/client_config_v1.go
@@ -13,9 +13,9 @@ import (
 type ClientConfigV1 struct {
 	EdaClientCfg     clients.EigenDAClientConfig
 	MaxBlobSizeBytes uint64
-	// Determines number of times to try blob dispersals:
-	// - If > 0: Try up to that many times total (first attempt + up to N-1 retries)
-	// - If = 0: Try only once (no retries)
-	// - If < 0: Try indefinitely until success
-	PutRetries int
+	// Number of times to try blob dispersals:
+	// - If > 0: Try N times total
+	// - If < 0: Retry indefinitely until success
+	// - If = 0: Not permitted
+	PutTries int
 }
--- a/config/eigendaflags/cli.go
+++ b/config/eigendaflags/cli.go
@@ -65,12 +65,12 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 		&cli.DurationFlag{
 			Name: ConfirmationTimeoutFlagName,
 			Usage: `The total amount of time that the client will spend waiting for EigenDA
-			to "confirm" (include onchain) a blob after it has been dispersed. Note that
-			we stick to "confirm" here but this really means InclusionTimeout,
-			not confirmation in the sense of confirmation depth.
-			
-			If ConfirmationTimeout time passes and the blob is not yet confirmed,
-			the client will return an api.ErrorFailover to let the caller failover to EthDA.`,
+				to "confirm" (include onchain) a blob after it has been dispersed. Note that
+				we stick to "confirm" here but this really means InclusionTimeout,
+				not confirmation in the sense of confirmation depth.
+				
+				If ConfirmationTimeout time passes and the blob is not yet confirmed,
+				the client will return an api.ErrorFailover to let the caller failover to EthDA.`,
 			Value:    15 * time.Minute,
 			EnvVars:  []string{withEnvPrefix(envPrefix, "CONFIRMATION_TIMEOUT")},
 			Category: category,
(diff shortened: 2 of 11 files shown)


ethenotethan (approved):
LGTM - very clean

config/eigendaflags/cli.go
```diff
@@ -150,9 +150,9 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 		},
 		// Flags that are proxy specific, and not used by the eigenda-client
 		// TODO: should we move this to a more specific category, like EIGENDA_STORE?
-		&cli.UintFlag{
+		&cli.IntFlag{
 			Name:     PutRetriesFlagName,
-			Usage:    "Number of times to retry blob dispersals.",
+			Usage:    "Total number of times to try blob dispersals before serving an error response. Set to 0 for no retries (try only once), set to a negative value to retry indefinitely.",
```
samlaf, line 155: find this msg a bit confusing because makes me wonder "if 0 for no retries, what does 1 do?"
According to your comments above think 1 doesn't retry either (same behavior as 0), but now I'm unsure.

samlaf: because could also be interpreted as number of times to retry AFTER the first attempt

samlaf: but also why are we not suggesting to set to 1 for no retries.

store/generated_key/v2/eigenda.go
```diff
@@ -31,14 +42,14 @@ var _ common.GeneratedKeyStore = (*Store)(nil)
...
-	putRetries uint,
+	putRetries int,
 	disperser *payloaddispersal.PayloadDisperser,
 	retriever clients.PayloadRetriever,
 	verifier clients.ICertVerifier,
 ) (*Store, error) {
 	return &Store{
 		log:        log,
-		putRetries: putRetries,
+		putRetries: utils.MapPutRetries(putRetries),
```
samlaf, line 52: Feel like MapPutRetries should be only called at the retry-go call site. Our configs should be self-consistent and only contain our own meaning of retry imo

store/generated_key/utils/store_utils.go
```diff
@@ -0,0 +1,22 @@
...
+// This function handles the translation between the user-facing API and the internal retry-go library.
+// IMPORTANT: In retry-go, the "attempts" parameter represents the TOTAL number of attempts, including the first try.
+//
+// The mapping works as follows:
+// - putRetries > 0: Maps to that value directly (e.g., 3 means "try once, then retry twice if needed")
+// - putRetries = 0: Maps to 1 (means "try once with no retries"). This mapping exists to provide sane default behavior
+// - putRetries < 0: Maps to 0 (means "retry indefinitely until success")
+//
+// Note: This function is used by both v1 and v2 EigenDA backends to maintain consistent retry behavior.
+func MapPutRetries(putRetries int) uint {
```
samlaf, line 13: rename to something more explicit like `convertToGoRetrySemantic` or something?

store/generated_key/eigenda/eigenda.go
```diff
@@ -27,7 +28,17 @@ type StoreConfig struct {
...
+	//
+	// IMPORTANT: In retry-go, this parameter means:
+	// - Value of 0: Try indefinitely until success
+	// - Value of 1: Try once with no retries
+	// - Value > 1: Try up to this many times total (first try + up to N-1 retries)
+	//
+	// Users should NEVER set this field directly. Always use NewStoreConfig()
+	// which correctly maps from the user-friendly int values to the internal uint values
+	// required by the retry-go library.
 	PutRetries uint
```
samlaf, line 42: don't really like this. I would prefer if all our configs were self-consistent (had same meaning) and we called MapPutRetries at the retry call site only (adds an inconsequential amount of processing overhead)

config/eigendaflags/cli.go
litt3: > According to your comments above think 1 doesn't retry either 

That's correct interpretation of the changeset you reviewed.

After making the change suggested in [this comment](https://github.com/Layr-Labs/eigenda-proxy/pull/365#discussion_r2041299232), though, `0` and `1` will no longer yield identical behavior: `0` will simply be disallowed

I think a source of confusion here is that the configuration field is called `PutRetries`, but it really ought to be called `PutTries`, since it is configuring the total number of times to attempt the `PUT`. Note that this name / functionality mismatch wasn't introduced with my changes: the `PutRetries` field already configured total number of tries.

I opted *not* to rename this field in this PR, for compatibility reasons. Do you think this is the correct decision, or should we bite the bullet and rename this to `PutTries`?

common/client_config_v1.go
```diff
@@ -13,5 +13,9 @@ import (
 type ClientConfigV1 struct {
 	EdaClientCfg     clients.EigenDAClientConfig
 	MaxBlobSizeBytes uint64
-	PutRetries       uint
+	// Determines number of times to try blob dispersals:
+	// - If > 0: Try up to that many times total (first attempt + up to N-1 retries)
+	// - If = 0: Try only once (no retries)
```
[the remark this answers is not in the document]
litt3: [7f89679](https://github.com/Layr-Labs/eigenda-proxy/pull/365/commits/7f896797084180ac218e619138c5d1501f36788c)

store/generated_key/utils/store_utils.go
litt3: Part of [7f89679](https://github.com/Layr-Labs/eigenda-proxy/pull/365/commits/7f896797084180ac218e619138c5d1501f36788c)

samlaf (approved):
LGTM I like this new set of changes. Good catch on renaming to `putTries` instead of `putRetries`. Think we should keep the flag as is though, to prevent more breaking changes.

PR #367, opened 2025-04-10
feat: Remove user facing load g2 option

litt3:
- Remove user facing "load g2 points" option, since it can be explicitly set correctly for each case, without user input

## Flag changes

- Deletes `ReadG2PointsFlag`. This isn't a breaking change, since a version of proxy has never been released with it.

```
		&cli.BoolFlag{
			Name:     ReadG2PointsFlagName,
			Usage:    "Whether to read in G2 SRS points.",
			EnvVars:  []string{withEnvPrefix(envPrefix, "READ_G2_POINTS")},
			Value:    false,
			Category: category,
		},
```

### DIFF
--- a/verify/cli.go
+++ b/verify/cli.go
@@ -19,7 +19,6 @@ var (
 	G2PowerOf2PathFlagNameDeprecated = withFlagPrefix("g2-power-of-2-path")
 	G2PathFlagName                   = withFlagPrefix("g2-path")
 	G2TrailingPathFlagName           = withFlagPrefix("g2-path-trailing")
-	ReadG2PointsFlagName             = withFlagPrefix("read-g2-points")
 	CachePathFlagName                = withFlagPrefix("cache-path")
 )
 
@@ -87,13 +86,6 @@ func KZGCLIFlags(envPrefix, category string) []cli.Flag {
 			Value:    "resources/g2.trailing.point",
 			Category: category,
 		},
-		&cli.BoolFlag{
-			Name:     ReadG2PointsFlagName,
-			Usage:    "Whether to read in G2 SRS points.",
-			EnvVars:  []string{withEnvPrefix(envPrefix, "READ_G2_POINTS")},
-			Value:    false,
-			Category: category,
-		},
 		&cli.StringFlag{
 			Name:     CachePathFlagName,
 			Usage:    "path to SRS tables for caching. This resource is not currently used, but needed because of the shared eigenda KZG library that we use. We will eventually fix this.",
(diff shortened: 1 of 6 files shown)


verify/cli.go
```diff
@@ -109,11 +101,13 @@ func ReadKzgConfig(ctx *cli.Context, maxBlobSizeBytes uint64) kzg.KzgConfig {
...
 		G2Path:          ctx.String(G2PathFlagName),
 		G2TrailingPath:  ctx.String(G2TrailingPathFlagName),
-		LoadG2Points:    ctx.Bool(ReadG2PointsFlagName),
 		CacheDir:        ctx.String(CachePathFlagName),
 		SRSOrder:        eigendaflags.SrsOrder,
 		SRSNumberToLoad: maxBlobSizeBytes / 32,         // # of fr.Elements
 		NumWorker:       uint64(runtime.GOMAXPROCS(0)), // #nosec G115
+		// we are intentionally not setting the `LoadG2Points` field here. `LoadG2Points` has different requirements
+		// for v1 vs v2. To make things foolproof, we just set this value locally prior to use, so that it can't
+		// ever be set incorrectly.
```
samlaf, line 110: Actually by not setting it, it defaults to golang's default zero value, which is false for booleans, which means its set for the V1 case.

samlaf (approved):

verify/cli.go
litt3: Right - but I thought it made sense to symbolically remove it from here, since we really don't care what the value is (even if the v1 use case, we still manually set it to `false`, even though it's already false)

Are there any changes you recommend here, for better clarity?

samlaf: you say "we just set this value locally prior to use", but you actually only set it in the v2 case right? Which means your code for v1 assumes that this is set to false.

idk its really just a nit, not even sure what I would change it to so lgtm

samlaf (approved):

verify/cli.go
litt3: > but you actually only set it in the v2 case right?

It is explicitly set in the `v1` case too (this was already being done, prior to this PR)

```
	// The verifier doesn't support loading trailing g2 points from a separate file. If LoadG2Points is true, and
	// the user is using a slimmed down g2 SRS file, the verifier will encounter an error while trying to load g2
	// points. Since the verifier doesn't actually need g2 points, it's safe to force LoadG2Points to false, to
	// sidestep the issue entirely.
	kzgConfig.LoadG2Points = false
```

https://github.com/Layr-Labs/eigenda-proxy/blob/5c193730fd8defcdc7e2849a4a06efd2a799a654/verify/verifier.go#L80

PR #369, opened 2025-04-11
feat: Implement validator node blob retrieval

litt3:
- closes DAINT-322
    - [link](https://linear.app/eigenlabs/issue/DAINT-322/v2-optional-distributed-retrieval-w-eigenda-operator-set) 
- closes EGDA-1094
    - [link](https://linear.app/eigenlabs/issue/EGDA-1094/add-optional-network-id-flag-to-populate-other-config-fields)  

DO NOT MERGE:
- Given that we have recently released V2 proxy support, it's likely we will need to release patches in the near future
- This PR update the eigenda core dependency to the latest `master` commit, since doing so is required to implement blob retrieval from validator nodes.
- proxy releases should only be made with a dependency on an official core release. Until a new core release is cut which includes the necessary changes to support distributed retrieval, this PR will need to remain as a feature branch

## Flag Changes

The following new *V2* flags are introduced:

```
		&cli.StringFlag{
			Name:     ServiceManagerAddrFlagName,
			Usage:    "Address of the EigenDA service manager contract.",
			EnvVars:  []string{withEnvPrefix(envPrefix, "SERVICE_MANAGER_ADDR")},
			Category: category,
			Required: false,
		},
		&cli.StringFlag{
			Name:     BLSOperatorStateRetrieverFlagName,
			Usage:    "Address of the BLS operator state retriever contract.",
			EnvVars:  []string{withEnvPrefix(envPrefix, "BLS_OPERATOR_STATE_RETRIEVER_ADDR")},
			Category: category,
			Required: false,
		},
		&cli.DurationFlag{
			Name: ValidatorTimeoutFlagName,
			Usage: "Timeout used when retrieving chunks directly from EigenDA validators. " +
				"This is a secondary retrieval method, in case retrieval from the relay network fails.",
			EnvVars:  []string{withEnvPrefix(envPrefix, "VALIDATOR_TIMEOUT")},
			Category: category,
			Value:    2 * time.Minute,
			Required: false,
		},
```

### DIFF
--- a/.env.exampleV1AndV2.holesky
+++ b/.env.exampleV1AndV2.holesky
@@ -86,6 +86,9 @@ EIGENDA_PROXY_EIGENDA_V2_CONTRACT_CALL_TIMEOUT=5s
 # Timeout used when querying an individual relay for blob contents
 EIGENDA_PROXY_EIGENDA_V2_RELAY_TIMEOUT=5s
 
+# Timeout used when retrieving payload chunks directly from EigenDA validators (fallback method)
+EIGENDA_PROXY_EIGENDA_V2_VALIDATOR_TIMEOUT=2m
+
 # Blob params version used when dispersing
 EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 
@@ -95,6 +98,12 @@ EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 # you only actually need to read the amount of SRS data that corresponds to the size of the largest blob that will be
 # sent, decreasing this value is a crude sort of optimization.
 EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
+
+# Address of the EigenDA service manager contract
+EIGENDA_PROXY_EIGENDA_V2_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
+
+# Address of the BLS operator state retriever contract
+EIGENDA_PROXY_EIGENDA_V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR=0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7
 
 # === Storage Configuration ===
 
--- a/.env.exampleV1AndV2.holesky
+++ b/.env.exampleV1AndV2.holesky
@@ -56,12 +56,6 @@ EIGENDA_PROXY_EIGENDA_V2_SIGNER_PRIVATE_KEY_HEX="0000000000000000000100000000000
 # JSON RPC node endpoint for the Ethereum network (V2)
 EIGENDA_PROXY_EIGENDA_V2_ETH_RPC=https://ethereum-holesky-rpc.publicnode.com
 
-# RPC endpoint of the EigenDA disperser (V2)
-EIGENDA_PROXY_EIGENDA_V2_DISPERSER_RPC=disperser-testnet-holesky.eigenda.xyz:443
-
-# Address of the EigenDACertVerifier contract for V2
-EIGENDA_PROXY_EIGENDA_V2_CERT_VERIFIER_ADDR=0xFe52fE1940858DCb6e12153E2104aD0fDFbE1162
-
 # Disable TLS for gRPC communication with the EigenDA disperser and retrieval subnet (V2)
 EIGENDA_PROXY_EIGENDA_V2_GRPC_DISABLE_TLS=false
 
@@ -99,11 +93,9 @@ EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 # sent, decreasing this value is a crude sort of optimization.
 EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 
-# Address of the EigenDA service manager contract
-EIGENDA_PROXY_EIGENDA_V2_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
-
-# Address of the BLS operator state retriever contract
-EIGENDA_PROXY_EIGENDA_V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR=0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7
+# The EigenDA network to run on. Specifying this chooses default values for V2_DISPERSER_RPC, V2_CERT_VERIFIER_ADDR,
+# V2_SERVICE_MANAGER_ADDR, and V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR.
+EIGENDA_PROXY_EIGENDA_V2_NETWORK=holesky_testnet
 
 # === Storage Configuration ===
 
--- a/.env.exampleV2.holesky
+++ b/.env.exampleV2.holesky
@@ -8,12 +8,6 @@ EIGENDA_PROXY_EIGENDA_V2_SIGNER_PRIVATE_KEY_HEX="0000000000000000000100000000000
 # JSON RPC node endpoint for the Ethereum network (V2)
 EIGENDA_PROXY_EIGENDA_V2_ETH_RPC=https://ethereum-holesky-rpc.publicnode.com
 
-# RPC endpoint of the EigenDA disperser (V2)
-EIGENDA_PROXY_EIGENDA_V2_DISPERSER_RPC=disperser-testnet-holesky.eigenda.xyz:443
-
-# Address of the EigenDACertVerifier contract for V2
-EIGENDA_PROXY_EIGENDA_V2_CERT_VERIFIER_ADDR=0xFe52fE1940858DCb6e12153E2104aD0fDFbE1162
-
 # Disable TLS for gRPC communication with the EigenDA disperser and retrieval subnet (V2)
 EIGENDA_PROXY_EIGENDA_V2_GRPC_DISABLE_TLS=false
 
@@ -51,11 +45,9 @@ EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 # sent, decreasing this value is a crude sort of optimization.
 EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 
-# Address of the EigenDA service manager contract (required for validator retrieval)
-EIGENDA_PROXY_EIGENDA_V2_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
-
-# Address of the BLS operator state retriever contract (required for validator retrieval)
-EIGENDA_PROXY_EIGENDA_V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR=0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7
+# The EigenDA network to run on. Specifying this chooses default values for V2_DISPERSER_RPC, V2_CERT_VERIFIER_ADDR,
+# V2_SERVICE_MANAGER_ADDR, and V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR.
+EIGENDA_PROXY_EIGENDA_V2_NETWORK=holesky_testnet
 
 # === Storage Configuration ===
 
--- a/.env.exampleV2.holesky
+++ b/.env.exampleV2.holesky
@@ -2,7 +2,6 @@
 
 # === V2 Configuration ===
 
-
 # Hex-encoded signer private key for payments with EigenDA disperser (V2)
 EIGENDA_PROXY_EIGENDA_V2_SIGNER_PRIVATE_KEY_HEX="0000000000000000000100000000000000000000000000000000000000000000"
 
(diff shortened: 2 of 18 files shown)


store/storage_manager_builder.go
```diff
@@ -344,6 +392,56 @@ func (smb *StorageManagerBuilder) buildRelayClient(
...
+
+	chainState := eth.NewChainState(ethReader, ethClient)
+
+	retrievalClient := clients_v2.NewRetrievalClient(
+		smb.log,
+		ethReader,
+		chainState,
+		kzgVerifier,
+		20, // Default connection count
+	)
```
anupsv, line 415: The retrieval client does the length verification and other things as well right ?

litt3: I don't believe it does the length verification, but it does verify each frame received.

The proxy does [verify the cert on chain](https://github.com/Layr-Labs/eigenda-proxy/blob/ff8365dbb4c936ea5996a5d35bc14486db056315/store/generated_key/v2/eigenda.go#L139), and blob->payload conversion [asserts](https://github.com/Layr-Labs/eigenda/blob/f4b3ef75c59ab138bd46df0e2328e5af41c3f580/api/clients/v2/coretypes/blob.go#L33) that blob length is < max bound

.env.exampleV2.holesky
```diff
@@ -49,6 +51,12 @@ EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 # sent, decreasing this value is a crude sort of optimization.
 EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 
+# Address of the EigenDA service manager contract (required for validator retrieval)
+EIGENDA_PROXY_EIGENDA_V2_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
+
+# Address of the BLS operator state retriever contract (required for validator retrieval)
+EIGENDA_PROXY_EIGENDA_V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR=0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7
+
```
ethenotethan, line 59: is that not readable from the service manager?

store/storage_manager_builder.go
bxue-l2: that is right, retriever need not to verify length proof. Sufficient to just re-compute the kzg.

samlaf:
@litt3 is this actually ready for review? You requested me for review but you've always said that you never request for reviews on DRAFT PRs. Plus the description says "DO NOT MERGE".

litt3:
> @litt3 is this actually ready for review? You requested me for review but you've always said that you never request for reviews on DRAFT PRs. Plus the description says "DO NOT MERGE".

@samlaf Sorry for the confusion. Did you see my message in slack?

```
NOTE: the PR is still in draft, since it cannot merge until a new core release is cut (see the PR body for more details). I'm notifying now to get feedback on the general strategy used to implement this feature.
```

> you never request for reviews on DRAFT PRs

You're right, I should have marked it as `ready`, to keep consistency.

TLDR; 
- ✅ ready for review
- ❌ not ready to merge

anupsv (dismissed):

.env.exampleV1AndV2.holesky
```diff
@@ -96,6 +99,12 @@ EIGENDA_PROXY_EIGENDA_V2_BLOB_PARAMS_VERSION=0
 # sent, decreasing this value is a crude sort of optimization.
 EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 
+# Address of the EigenDA service manager contract
+EIGENDA_PROXY_EIGENDA_V2_SERVICE_MANAGER_ADDR=0xD4A7E1Bd8015057293f0D0A557088c286942e84b
+
+# Address of the BLS operator state retriever contract
+EIGENDA_PROXY_EIGENDA_V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR=0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7
```
samlaf, line 106: Config management for contract interactions getting more and more complex :\
Out of scope for this PR but @pakim249CAL had suggested having a registry contract that would hold the addresses of all our contracts. This would be a much nicer UX for users, since everywhere they only ever have to give a single contract address, and then all the other contract addresses are fetched automatically for them.

.env.exampleV2.holesky
samlaf: Yeah its not because stateRetriever was meant to be a standalone shared contract that everyone would use (all its calls take a serviceManagerAddr as argument). But in the end each team ends up deploying their own anyways so.. we're stuck with this.

common/client_config_v2.go
```diff
@@ -24,6 +26,13 @@ type ClientConfigV2 struct {
 	PutTries                   int
 	MaxBlobSizeBytes           uint64
 	EigenDACertVerifierAddress string
+
+	// Fields required for validator payload retrieval
+	BLSOperatorStateRetrieverAddr string
+	EigenDAServiceManagerAddr     string
+
+	// RetrieversToEnable specifies which retrievers should be enabled
+	RetrieversToEnable []RetrieverType
```
samlaf, line 35: do we plan on having more than 2 types? Is there any reason why people would not always want both? Perhaps just to not have to setup all the contract addresses required to setup the validator retriever?

common/client_config_v2.go
```diff
@@ -24,6 +26,13 @@ type ClientConfigV2 struct {
 	PutTries                   int
 	MaxBlobSizeBytes           uint64
 	EigenDACertVerifierAddress string
+
+	// Fields required for validator payload retrieval
+	BLSOperatorStateRetrieverAddr string
+	EigenDAServiceManagerAddr     string
```
samlaf, line 32: if they are required for validator payload retrieval, why aren't they in the ValidatorPayloadRetrieverCfg?

.env.exampleV1AndV2.holesky
samlaf: An alternative which I've mentioned in the past is that we add a `--chain-id` or `--network` (given that we have both preprod and testnet on holesky) flag and that this sets all the correct addresses (that would be hardcoded in the binary for holesky and mainnet)

store/generated_key/v2/eigenda.go
```diff
@@ -64,12 +64,18 @@ func (e Store) Get(ctx context.Context, key []byte) ([]byte, error) {
 		return nil, fmt.Errorf("RLP decoding EigenDA v2 cert: %w", err)
 	}
 
-	payload, err := e.retriever.GetPayload(ctx, &cert)
-	if err != nil {
-		return nil, fmt.Errorf("getting payload: %w", err)
+	// Try each retriever in sequence until one succeeds
+	var errs []error
+	for _, retriever := range e.retrievers {
```
samlaf, line 69: There's a lot of code in this PR to possibly make the retrieval method more flexible in the future (list instead of just hardcoding 2 methods, flag that is unused just to set the config, etc.)

Do we really need this flexibility? Should we consider only adding it in the future if/when we need it? Is the idea that we would eventually want to add more retrieval methods, such as unifying all the different backends we currently have (s3, redis) as more retrievers in the future?

store/storage_manager_builder.go
```diff
@@ -118,17 +120,35 @@ func (smb *StorageManagerBuilder) Build(ctx context.Context) (*Manager, error) {
 		return nil, fmt.Errorf("dispersal backend is set to V1, but V1 backend is not enabled")
 	}
 
+	var kzgVerifier *kzgverifier.Verifier
+	// there are two cases in which we need to construct the kzgVerifier:
+	// 1. V1
+	// 2. V2, when validator retrieval is enabled
```
samlaf, line 126: why dont we need it when retrieving from relays?

store/storage_manager_builder.go
```diff
@@ -118,17 +120,35 @@ func (smb *StorageManagerBuilder) Build(ctx context.Context) (*Manager, error) {
...
+	if v1Enabled || v2Enabled && slices.Contains(smb.v2ClientCfg.RetrieversToEnable, common.ValidatorRetrieverType) {
+		// The verifier doesn't support loading trailing g2 points from a separate file. If LoadG2Points is true, and
+		// the user is using a slimmed down g2 SRS file, the verifier will encounter an error while trying to load g2
+		// points. Since the verifier doesn't actually need g2 points, it's safe to force LoadG2Points to false, to
+		// sidestep the issue entirely.
+		kzgConfig := smb.kzgConfig
```
samlaf, line 132: lol more beautiful srs loading code. I have no idea what this means... hoping you figured it out and its not too much bad tech debt.

config/v2/eigendaflags/cli.go
```diff
@@ -168,12 +194,20 @@ func ReadClientConfigV2(ctx *cli.Context) (common.ClientConfigV2, error) {
...
+		ValidatorPayloadRetrieverCfg:  readValidatorRetrievalConfig(ctx),
+		PutTries:                      ctx.Int(PutRetriesFlagName),
+		MaxBlobSizeBytes:              maxBlobLengthBytes,
+		EigenDACertVerifierAddress:    ctx.String(CertVerifierAddrFlagName),
+		BLSOperatorStateRetrieverAddr: ctx.String(BLSOperatorStateRetrieverFlagName),
+		EigenDAServiceManagerAddr:     ctx.String(ServiceManagerAddrFlagName),
+		// we don't expose this configuration to users, as all production use cases should have
+		// both retrieval methods enabled. This could be exposed in the future, if necessary.
+		// Note the order of these retrievers, which is significant: the relay retriever will be
+		// tried first, and the validator retriever will only be tried if the relay retriever fails
```
samlaf, line 209: Is this ordering thing true? In the store_manager_builder, you build them always relay first, regardless of config ordering.

common/client_config_v2.go
litt3: > Is there any reason why people would not always want both

For testing purposes, it's really nice to have either one or the other. For production, it's currently not even an option for users to opt out of distributed retrieval (by design). 

> do we plan on having more than 2 types?

Ideally, I think that we should eventually define the various caching / fallback "backends" in this way. I don't see any reason why they couldn't conform to the disperser / retriever interfaces, and then configuration would be a lot more straightforward. It was with this plan in mind that I decided to pursue this more dynamic strategy

common/client_config_v2.go
litt3: the `ValidatorPayloadRetrieverCfg` only contains values directly needed by the `ValidatorPayloadRetriever`.

These addresses are needed by the `RetrievalClient`, which is passed into the `ValidatorPayloadRetriever` already constructed.

The problem is that there is no `RetrievalClientConfig`: the `RetrievalClient` just accepts all of its required parameters directly. Ideally, a `RetrievalClientConfig` would be created that contains all necessary values for construction, but this would require core changes.

store/generated_key/v2/eigenda.go
litt3: > Is the idea that we would eventually want to add more retrieval methods, such as unifying all the different backends we currently have (s3, redis) as more retrievers in the future?

Yes, exactly. Pursuing this strategy is at the top of my proxy wishlist, as I think would *greatly* simplify things.

I think it makes sense to introduce this flexibility now, rather than adding additional hard coded stuff that would make this step more difficult in the future. I also don't think this flexibility cost us very much in terms of code complexity: this PR would be only marginally simpler if we went the hardcoded route

store/storage_manager_builder.go
litt3: The verifier is necessary for validator retrieval, to check individual chunks from validator nodes. Retrieval from a relay is simpler, since you get the fully formed blob, and just need to check the cert.

store/storage_manager_builder.go
litt3: Everything to do with the kzg usage is required to be held together by duct tape, unfortunately. It'll be the case until we buckle down and rework the underlying utilities. I don't think this is really adding to the tech debt: it's more just an expression of the existing debt

config/v2/eigendaflags/cli.go
litt3: Yes, the ordering thing is true. Relevant here isn't construction order: rather, it's the slice iteration that takes place in the `Get` pathway

```
	for _, retriever := range e.retrievers {
		payload, err := retriever.GetPayload(ctx, &cert)
		if err == nil {
			return payload.Serialize(), nil
		}

		e.log.Debugf("Payload retriever failed: %v", err)
	}
```

https://github.com/Layr-Labs/eigenda-proxy/blob/4d8097dac6d8442705ba2459e6fd38670d254869/store/generated_key/v2/eigenda.go#L69

common/client_config_v2.go
samlaf: Nice I think that would indeed be a really nice code refactor

store/storage_manager_builder.go
samlaf: for the relay, you also need to check that the blob matches the commitment in the cert right?

config/v2/eigendaflags/cli.go
samlaf: But isnt the ordering of e.retrievers populated by the build step, which changes the ordering? https://github.com/Layr-Labs/eigenda-proxy/blob/4d8097dac6d8442705ba2459e6fd38670d254869/store/storage_manager_builder.go#L262-L272

store/storage_manager_builder.go
litt3: Yes, but this uses the `Prover`, as opposed to the `Verifier`. This is because we "verify" the commitment by generating our own, and then comparing.

config/v2/eigendaflags/cli.go
litt3: You're 💯% correct, I didn't read my own code carefully enough...

bug fixed in [6b3f050](https://github.com/Layr-Labs/eigenda-proxy/pull/369/commits/6b3f0508a915e6605b141c8432dde04867e938b4)

samlaf (approved):
LGTM

PR #370, opened 2025-04-11
feat: Cert encoding byte

ethenotethan:
## Fixes Issue

Fixes #

### DIFF
--- a/clients/standard_client/client.go
+++ b/clients/standard_client/client.go
@@ -24,8 +24,21 @@ var (
 	ErrServiceUnavailable = fmt.Errorf("eigenda service is temporarily unavailable")
 )
 
+// EncodingType represents the serialization format for the certificate
+type EncodingType string
+
+const (
+	// RLPEncoding represents RLP encoding (default)
+	RLPEncoding EncodingType = "rlp"
+	// VerifyCertV2 ABI encoding represents ABI encoding for the verifyCertV2 method
+	// on the EigenDACertVerifier contract
+	// https://github.com/Layr-Labs/eigenda/blob/10d1fcfc44e8c7e35a3e585799fe10447dbec6a8/contracts/src/core/EigenDACertVerifier.sol#L114-L118
+	ABIVerifyV2CertEncoding EncodingType = "abi"
+)
+
 type Config struct {
-	URL string // EigenDA proxy REST API URL
+	URL         string       // EigenDA proxy REST API URL
+	EncodingType EncodingType // Certificate encoding type (optional)
 }
 
 type HTTPClient interface {
@@ -123,13 +136,20 @@ func (c *Client) GetData(ctx context.Context, comm []byte) ([]byte, error) {
 // SetData writes raw byte data to DA and returns the associated certificate
 // which should be verified within the proxy
 func (c *Client) SetData(ctx context.Context, b []byte) ([]byte, error) {
-	url := fmt.Sprintf("%s/put?commitment_mode=standard", c.cfg.URL)
+	baseURL := fmt.Sprintf("%s/put?commitment_mode=standard", c.cfg.URL)
+	
+	// Add encoding type if specified in config
+	url := baseURL
+	if c.cfg.EncodingType != "" {
+		url = fmt.Sprintf("%s&encoding=%s", baseURL, c.cfg.EncodingType)
+	}
+	
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(b))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/octet-stream")
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := c.httpClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
--- a/commitments/eigenda.go
+++ b/commitments/eigenda.go
@@ -7,6 +7,8 @@ const (
 	CertV0 EigenDACommitmentType = iota
 	// EigenDA V2
 	CertV1
+	// EigenDA V2 with encoding byte
+	CertV2
 )
 
 // CertCommitment is the binary representation of a commitment.
@@ -17,8 +19,9 @@ type CertCommitment interface {
 }
 
 type EigenDACommitment struct {
-	prefix EigenDACommitmentType
-	b      []byte
+	prefix   EigenDACommitmentType
+	b        []byte
+	encoding *EncodingType // Optional encoding type, used only for V2+
 }
 
 // NewEigenDACommitment creates a new commitment from the given input.
@@ -29,6 +32,16 @@ func NewEigenDACommitment(input []byte, commitmentType EigenDACommitmentType) Ei
 	}
 }
 
+// NewEigenDACommitmentWithEncoding creates a new commitment with encoding from the given input.
+func NewEigenDACommitmentWithEncoding(input []byte, commitmentType EigenDACommitmentType,
+	encoding EncodingType) EigenDACommitment {
+	return EigenDACommitment{
+		prefix:   commitmentType,
+		b:        input,
+		encoding: &encoding,
+	}
+}
+
 // CommitmentType returns the commitment type of EigenDACommitment.
 func (c EigenDACommitment) CommitmentType() EigenDACommitmentType {
 	return c.prefix
@@ -36,5 +49,11 @@ func (c EigenDACommitment) CommitmentType() EigenDACommitmentType {
 
 // Encode adds a commitment type prefix self describing the commitment.
 func (c EigenDACommitment) Encode() []byte {
+	// For V2+ certificates with encoding type
+	if c.prefix >= CertV2 && c.encoding != nil {
+		return append([]byte{byte(c.prefix), byte(*c.encoding)}, c.b...)
+	}
+
+	// For V0/V1 certificates (backward compatibility)
 	return append([]byte{byte(c.prefix)}, c.b...)
 }
--- a/commitments/mode.go
+++ b/commitments/mode.go
@@ -4,10 +4,22 @@ import (
 	"fmt"
 )
 
+// EncodingType represents the serialization format for the certificate
+type EncodingType byte
+
+const (
+	// RLPEncoding represents RLP encoding (default)
+	RLPEncoding EncodingType = iota
+	// ABIVerifyV2CertEncoding represents ABI encoding
+	ABIVerifyV2CertEncoding
+)
+
 type CommitmentMeta struct {
 	Mode CommitmentMode
 	// version is shared for all modes and denotes version of the EigenDA certificate
 	Version EigenDACommitmentType
+	// encoding type for the certificate (defaults to RLPEncoding)
+	Encoding EncodingType
 }
 
 type CommitmentMode string
@@ -31,22 +43,76 @@ func StringToCommitmentMode(s string) (CommitmentMode, error) {
 	}
 }
 
+func StringToEncodingType(s string) (EncodingType, error) {
+	switch s {
+	case "rlp", "RLP", "0":
+		return RLPEncoding, nil
+	case "abi", "ABI", "1":
+		return ABIVerifyV2CertEncoding, nil
+	default:
+		return RLPEncoding, fmt.Errorf("unknown encoding type: %s, using default RLP encoding", s)
+	}
+}
+
+// DecodeCommitmentType returns the encoding type from a certificate commitment
+func DecodeCommitmentType(cert []byte) (EigenDACommitmentType, EncodingType, []byte, error) {
+	if len(cert) == 0 {
+		return 0, 0, nil, fmt.Errorf("commitment is empty")
+	}
+
+	// Extract version byte (always present)
+	version := EigenDACommitmentType(cert[0])
+	var encoding EncodingType
+	var payload []byte
+
+	// For CertV2+, we have an encoding byte
+	if version >= CertV2 {
+		if len(cert) < 2 {
+			return 0, 0, nil, fmt.Errorf("commitment too short for CertV2")
+		}
+		encoding = EncodingType(cert[1])
+		payload = cert[2:] // Skip version and encoding bytes
+	} else {
+		// For CertV0/V1, there's no encoding byte (default to RLP)
+		encoding = RLPEncoding
+		payload = cert[1:] // Skip only version byte
+	}
+
+	return version, encoding, payload, nil
+}
+
 func EncodeCommitment(
 	bytes []byte,
 	commitmentMode CommitmentMode,
 	commitmentType EigenDACommitmentType,
+	encodingType EncodingType,
 ) ([]byte, error) {
 	switch commitmentMode {
 	case OptimismKeccak:
 		return Keccak256Commitment(bytes).Encode(), nil
 
 	case OptimismGeneric:
+		// For V2 certificates with encoding specified, add encoding byte
+		if commitmentType >= CertV2 {
+			certCommit := NewEigenDACommitmentWithEncoding(bytes, commitmentType, encodingType).Encode()
+			svcCommit := EigenDASvcCommitment(certCommit).Encode()
+			altDACommit := NewGenericCommitment(svcCommit).Encode()
+			return altDACommit, nil
+		}
+
+		// Fallback to previous behavior for V0/V1
 		certCommit := NewEigenDACommitment(bytes, commitmentType).Encode()
 		svcCommit := EigenDASvcCommitment(certCommit).Encode()
 		altDACommit := NewGenericCommitment(svcCommit).Encode()
 		return altDACommit, nil
 
 	case Standard:
+		// For V2 certificates with encoding specified, add encoding byte
+		if commitmentType >= CertV2 {
+			return NewEigenDACommitmentWithEncoding(bytes, commitmentType, encodingType).Encode(), nil
+		}
+
+		// Fallback to previous behavior for V0/V1
 		return NewEigenDACommitment(bytes, commitmentType).Encode(), nil
 	}
 
--- a/common/serialization.go
+++ b/common/serialization.go
@@ -1,0 +1,106 @@
+package common
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda-proxy/commitments"
+	"github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/rlp"
+
+	binding "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDACertVerifier"
+)
+
+var (
+	v2VerifyMethodInputs abi.Arguments
+)
+
+func init() {
+	certVerifierABI, err := binding.ContractEigenDACertVerifierMetaData.GetAbi()
+	if err != nil {
+		panic(err)
+	}
+
+	v2VerifyMethodInputs = certVerifierABI.Methods["verifyDACertV2"].Inputs
+}
+
+func DecodeCertFromCtx(ctx context.Context, commit []byte) (*coretypes.EigenDACert, error) {
+	encodingAlg, ok := ctx.Value(EncodingCtxKey).(commitments.EncodingType)
+	if !ok {
+		return nil, fmt.Errorf("could not read encoding type from context")
+	}
+	cert, err := DecodeV2CertFromBytes(encodingAlg, commit)
+	if err != nil {
+		return nil, fmt.Errorf("decoding V2 cert from bytes: %w", err)
+	}
+
+	return cert, nil
+}
+
+func EncodeV2CertToBytes(encoding commitments.EncodingType, cert *coretypes.EigenDACert) ([]byte, error) {
+	switch encoding {
+	case commitments.ABIVerifyV2CertEncoding:
+		return v2VerifyMethodInputs.Pack(
+			cert.BatchHeader,
+			cert.BlobInclusionInfo,
+			cert.NonSignerStakesAndSignature,
+			cert.SignedQuorumNumbers)
+
+	case commitments.RLPEncoding:
+		return rlp.EncodeToBytes(cert)
+
+	default:
+		return nil, fmt.Errorf("encoding v2 cert to bytes: %x", encoding)
+	}
+}
+
+func interfaceToBytesJSON(data interface{}) ([]byte, error) {
+	bytes, err := json.Marshal(data)
+	if err != nil {
+		return nil, err
+	}
+	return bytes, nil
+}
+
+func DecodeV2CertFromBytes(encoding commitments.EncodingType, commitment []byte) (*coretypes.EigenDACert, error) {
+	switch encoding {
+	case commitments.ABIVerifyV2CertEncoding:
+
+		var abiMap map[string]interface{}
+		err := v2VerifyMethodInputs.UnpackIntoMap(abiMap, commitment)
+		if err != nil {
+			return nil, fmt.Errorf("unpacking from encoding ABI: %w", err)
+		}
+
+		// use json as intermediary to cast abstract type to bytes to
+		// then deserialize into structured certificate type
+		b, err := interfaceToBytesJSON(abiMap)
+		if err != nil {
+			return nil, err
+		}
+
+		var cert *coretypes.EigenDACert
+		err = json.Unmarshal(b, &cert)
+
+		if err != nil {
+			return nil, fmt.Errorf("json unmarshal v2 cert: %w", err)
+		}
+
+		return cert, nil
+
+	case commitments.RLPEncoding:
+		var cert *coretypes.EigenDACert
+
+		err := rlp.DecodeBytes(commitment, cert)
+		if err != nil {
+			return nil, fmt.Errorf("rlp decoding v2 cert: %w", err)
+		}
+
+		return cert, nil
+
+	default:
+		return nil, fmt.Errorf("encoding v2 cert to bytes: %x", encoding)
+	}
+}
--- a/e2e/server_test.go
+++ b/e2e/server_test.go
@@ -2,7 +2,6 @@ package e2e
 
 import (
 	"net/http"
-	"strings"
 	"testing"
 	"time"
 
@@ -16,6 +15,7 @@ import (
 	"github.com/Layr-Labs/eigenda/common/testutils/random"
 	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
 	altda "github.com/ethereum-optimism/optimism/op-alt-da"
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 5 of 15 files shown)


README.md
```diff
@@ -352,25 +352,38 @@ Validator nodes proceed with the exact reverse process as that used by the seque
...
-| commitment_type (byte) | da_layer_byte | version_byte | payload           |
-| ---------------------- | ------------- | ------------ | ----------------- |
-| 0x00                   |               |              | keccak_commitment |
-| 0x01                   | 0x00          | 0x00         | eigenda_cert_v1   |
-| 0x01                   | 0x00          | 0x01         | eigenda_cert_v2   |
+| commitment_type (byte) | da_layer_byte | version_byte | encoding_byte | payload           |
+| ---------------------- | ------------- | ------------ | ------------- | ----------------- |
+| 0x00                   |               |              |               | keccak_commitment |
+| 0x01                   | 0x00          | 0x00         | N/A           | eigenda_cert_v0   |
+| 0x01                   | 0x00          | 0x01         | N/A           | eigenda_cert_v1   |
```
samlaf, line 368: I'd personally remove the N/A and follow the pattern we had previously - just empty slot when its not present. Find its easier to parse as opposed to a diagram cluttered with N/As

README.md
```diff
@@ -352,25 +352,38 @@ Validator nodes proceed with the exact reverse process as that used by the seque
...
+| version_byte | encoding_byte | payload         |
+| ------------ | ------------- | --------------- |
+| 0x00         | N/A           | eigenda_cert_v0 |
+| 0x01         | N/A           | eigenda_cert_v1 |
+| 0x02         | 0x00          | eigenda_cert_v2 (RLP) |
+| 0x02         | 0x01          | eigenda_cert_v2 (ABI_verifyDACertV2) |
+
+For certificates with version byte 0x00 or 0x01, no encoding byte is included and RLP encoding is assumed. For version 0x02 and higher, an encoding byte is included that specifies the serialization format:
+- 0x00: RLP encoding (default)
+* `0x01`: `EigenDACertVerifier` *verifyDACertV2* ABI encoding (for smart contract verification)
```
samlaf, line 386: do we want to wait until we've done the struct refactoring for verifyDACertV2? Then we can just say that its the ABI encoding of the solidity cert struct, which feels more natural than pointing to a function's arguments.

clients/standard_client/client.go
```diff
@@ -24,8 +24,21 @@ var (
 	ErrServiceUnavailable = fmt.Errorf("eigenda service is temporarily unavailable")
 )
 
+// EncodingType represents the serialization format for the certificate
+type EncodingType string
```
samlaf, line 28: maybe `CertEncoding` or `CertSerialization` instead?

clients/standard_client/client.go
```diff
@@ -24,8 +24,21 @@ var (
...
+	// VerifyCertV2 ABI encoding represents ABI encoding for the verifyCertV2 method
+	// on the EigenDACertVerifier contract
+	// https://github.com/Layr-Labs/eigenda/blob/10d1fcfc44e8c7e35a3e585799fe10447dbec6a8/contracts/src/core/EigenDACertVerifier.sol#L114-L118
+	ABIVerifyV2CertEncoding EncodingType = "abi"
+)
+
 type Config struct {
-	URL string // EigenDA proxy REST API URL
+	URL         string       // EigenDA proxy REST API URL
+	EncodingType EncodingType // Certificate encoding type (optional)
```
samlaf, line 41: if optional, then describe what the zero (empty string) value defaults to

clients/standard_client/client.go
samlaf: Also not a fan of defining fundamental types in the client. Could we define these in some proxy package and import it in the client instead?

commitments/eigenda.go
```diff
@@ -17,8 +19,9 @@ type CertCommitment interface {
 }
 
 type EigenDACommitment struct {
-	prefix EigenDACommitmentType
-	b      []byte
+	prefix   EigenDACommitmentType
+	b        []byte
+	encoding *EncodingType // Optional encoding type, used only for V2+
```
samlaf, line 24: move above b to align with serialization order.
Also the EigenDACommitmentType is defined as a byte, whereas EncodingType is a string. Feel like we should use the same structure (which is also why they should both be defined in the same file imo, which makes it easier to see this structure.

samlaf: also don't think encoding should be optional? we always know what the encoding is (even if only implicitly in the case for v0 and v1 certs.

commitments/eigenda.go
```diff
@@ -29,12 +32,28 @@ func NewEigenDACommitment(input []byte, commitmentType EigenDACommitmentType) Ei
...
 func (c EigenDACommitment) CommitmentType() EigenDACommitmentType {
 	return c.prefix
 }
 
 // Encode adds a commitment type prefix self describing the commitment.
 func (c EigenDACommitment) Encode() []byte {
+	// For V2+ certificates with encoding type
+	if c.prefix >= CertV2 && c.encoding != nil {
+		return append([]byte{byte(c.prefix), byte(*c.encoding)}, c.b...)
+	}
```
samlaf, line 55: hmm don't we want to always add the encoding byte for v2+ certs? why allow encoding to be nil?

commitments/mode.go
```diff
@@ -4,10 +4,22 @@ import (
...
+// EncodingType represents the serialization format for the certificate
+type EncodingType byte
+
+const (
+	// RLPEncoding represents RLP encoding (default)
+	RLPEncoding EncodingType = iota
+	// ABIVerifyV2CertEncoding represents ABI encoding
+	ABIVerifyV2CertEncoding
+)
+
```
samlaf, line 16: oh I see you redefine it here... cant the client just import this? and we can have a translation between byte and string if needed

commitments/mode.go
```diff
@@ -4,10 +4,22 @@ import (
...
 type CommitmentMeta struct {
 	Mode CommitmentMode
 	// version is shared for all modes and denotes version of the EigenDA certificate
 	Version EigenDACommitmentType
+	// encoding type for the certificate (defaults to RLPEncoding)
+	Encoding EncodingType
```
samlaf, line 22: I like this like this, we use the zero value as a sane default. Prefer this over the pointer to encode optionality you used elsewhere. Not sure why we have 2 different ways of representing this

commitments/mode.go
```diff
@@ -31,22 +43,76 @@ func StringToCommitmentMode(s string) (CommitmentMode, error) {
 	}
 }
 
+func StringToEncodingType(s string) (EncodingType, error) {
+	switch s {
+	case "rlp", "RLP", "0":
```
samlaf, line 48: maybe use s.toLower() instead?

samlaf: this would also allow Rlp for example

store/manager.go
```diff
@@ -129,6 +129,8 @@ func (m *Manager) Get(ctx context.Context, key []byte, cm commitments.Commitment
 			m.log.Warn("Failed to read from cache targets", "err", err)
 		}
 
+		ctx = context.WithValue(ctx, common.EncodingCtxKey, cm.Encoding)
```
samlaf, line 132: why use context key/value dict everywhere instead of passing as a proper (typed!) argument?

common/serialization.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+func interfaceToBytesJSON(data interface{}) ([]byte, error) {
+	bytes, err := json.Marshal(data)
+	if err != nil {
+		return nil, err
+	}
+	return bytes, nil
+}
```
samlaf, line 65: delete. just use json.Marshal directly

server/handlers.go
```diff
@@ -160,14 +174,44 @@ func (svr *Server) handleGetEigenDADispersalBackend(w http.ResponseWriter, _ *ht
 // =================================================================================================
 
 // handlePostStdCommitment handles the POST request for std commitments.
```
samlaf, line 176: comment is misplaced

server/handlers.go
```diff
@@ -160,14 +174,44 @@ func (svr *Server) handleGetEigenDADispersalBackend(w http.ResponseWriter, _ *ht
...
+// parseEncodingQueryParamType parses the encoding query parameter
+func parseEncodingQueryParamType(w http.ResponseWriter, r *http.Request) (commitments.EncodingType, bool, error) {
```
samlaf, line 178: what is the bool returned for?

server/routing.go
```diff
@@ -14,14 +13,15 @@ const (
...
+	routingQueryParamEncoding       = "encoding"
 )
 
 func (svr *Server) RegisterRoutes(r *mux.Router) {
 	subrouterGET := r.Methods("GET").PathPrefix("/get").Subrouter()
 	// std commitments (for nitro)
 	subrouterGET.HandleFunc("/"+
 		"{optional_prefix:(?:0x)?}"+ // commitments can be prefixed with 0x
-		"{"+routingVarNameVersionByteHex+":[0-9a-fA-F]{2}}"+ // should always be 0x00 for now but we let others through to return a 404
+		"{"+routingVarNameVersionByteHex+":[0-9a-fA-F]{2}}"+
```
samlaf, line 24: shouldn't we add the encodingByte regexp here as well?

samlaf (commented):
Approach LGTM. Left a bunch of nits on the specifics

server/handlers.go
```diff
@@ -54,6 +56,11 @@ func (svr *Server) handleGetStdCommitment(w http.ResponseWriter, r *http.Request
 		return fmt.Errorf("failed to decode commitment %s: %w", rawCommitmentHex, err)
 	}
 
+	if versionByte >= byte(commitments.CertV2) {
```
anupsv, line 59: wondering why >= instead of of just == ?

store/manager.go
ethenotethan: this is a workaround to conform with existing interface constraints

commitments/mode.go
```diff
@@ -4,10 +4,22 @@ import (
 	"fmt"
 )
 
+// EncodingType represents the serialization format for the certificate
+type EncodingType byte
```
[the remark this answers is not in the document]
ethenotethan: the issue with having an imported definition in the standard client package is that proxy's core dependencies would also be imported which isn't possible given how we've provisioned the go workspace

e2e/server_test.go
```diff
@@ -510,3 +516,81 @@ func testMaxBlobSize(t *testing.T, dispersalBackend common.EigenDABackend) {
...
+	t.Parallel()
+
+	// Set up a test environment with memstore backend and V2 dispersal
+	testCfg := testutils.NewTestConfig(testutils.GetBackend(), common.V2EigenDABackend, nil)
+	tsConfig := testutils.BuildTestSuiteConfig(testCfg)
+	ts, kill := testutils.CreateTestSuite(tsConfig)
+	t.Cleanup(kill)
+
+	// Create test payload
+	testPayload := []byte("test-payload-for-abi-encoding")
```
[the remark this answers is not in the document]
ethenotethan: is there any difference in testing guarantee by introducing non-determinism at the payload level?

clients/standard_client/client.go
```diff
@@ -24,8 +24,21 @@ var (
...
+// EncodingType represents the serialization format for the certificate
+type EncodingType string
+
+const (
+	// RLPEncoding represents RLP encoding (default)
+	RLPEncoding EncodingType = "rlp"
+	// VerifyCertV2 ABI encoding represents ABI encoding for the verifyCertV2 method
+	// on the EigenDACertVerifier contract
+	// https://github.com/Layr-Labs/eigenda/blob/10d1fcfc44e8c7e35a3e585799fe10447dbec6a8/contracts/src/core/EigenDACertVerifier.sol#L114-L118
+	ABIVerifyV2CertEncoding EncodingType = "abi"
```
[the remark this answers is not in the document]
ethenotethan: `ABIEncoding` could mean a plethora of things and we're specifically encoding using the `EigenDACertVerifier`'s `verifyV2Cert` method. In the future we could support packed encodings for other verification functions on the cert verifier as well so having the method specific suffix is necessary

server/routing.go
ethenotethan: how would you add that regex where you wouldn't also be unnecessarily truncating commitment bytes from an older version commitment that doesn't have the encoding prefix?

clients/standard_client/client.go
ethenotethan: how would you do that in a way that doesn't introduce proxy's dependency graph to the client?

README.md
ethenotethan: yes! and fully agree

server/handlers.go
```diff
@@ -40,9 +40,11 @@ func (svr *Server) handleGetStdCommitment(w http.ResponseWriter, r *http.Request
 	if err != nil {
 		return fmt.Errorf("error parsing version byte: %w", err)
 	}
+
 	commitmentMeta := commitments.CommitmentMeta{
-		Mode:    commitments.Standard,
-		Version: commitments.EigenDACommitmentType(versionByte),
+		Mode:     commitments.OptimismGeneric,
```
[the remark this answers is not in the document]
ethenotethan: good catch - was curious why tests were still passing and its because standard/op-generic are treated identically when doing downstream fetching. worth noting that there's currently no difference in logic expressions when setting between standard && op-generic

server/routing.go
```diff
@@ -14,14 +13,15 @@ const (
 	routingVarNamePayloadHex        = "payload_hex"
 	routingVarNameVersionByteHex    = "version_byte_hex"
 	routingVarNameCommitTypeByteHex = "commit_type_byte_hex"
+	routingQueryParamEncoding       = "encoding"
```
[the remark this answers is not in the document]
ethenotethan: > it hasn't actually been added to the routes being registered.

right its because we're reading it in differently or downstream of the gorilla mux parsing. couldn't find a way to support cross version compatibility at the regex level processing level

store/generated_key/memstore/v2/memstore.go
```diff
@@ -181,7 +181,7 @@ func (e *MemStore) Get(_ context.Context, commit []byte) ([]byte, error) {
 // ephemeral db key = keccak256(pseudo_random_cert)
 // this is done to verify that a rollup must be able to provide
 // the same certificate used in dispersal for retrieval
-func (e *MemStore) Put(_ context.Context, value []byte) ([]byte, error) {
+func (e *MemStore) Put(ctx context.Context, value []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
ethenotethan: would prefer that this be a follow up change or live as a separate PR given the large code overhead required

clients/standard_client/client.go
samlaf: You import a package that doesn't have many transitive dependencies. For example I tried adding a main.go in clients that only imports common/consts:
```
package clients

import (
	"fmt"

	"github.com/Layr-Labs/eigenda-proxy/common/consts"
)

func main() {
	fmt.Println(consts.EthHappyPathFinalizationDepthBlocks)

}
```

I don't fully understand it but I think golang does lazy module loading, and only based on PACKAGES that a build requires. The difference between package and module is important here. See https://go.dev/ref/mod#go-mod-file-go:
> The go.mod file includes an explicit [require directive](https://go.dev/ref/mod#go-mod-file-require) for each module that provides any package transitively imported by a package or test in the main module.

store/manager.go
samlaf: Why do we need to conform to existing interfaces?

server/routing.go
samlaf: Oh fuck right... now we're stuck with some context-dependent encoding. :(
Maybe one way is to have 2 separate routes? Something like:
```
// Route for version bytes that require encoding byte
router.HandleFunc("/"+
    "{optional_prefix:(?:0x)?}"+
    "{version:[0-9a-fA-F]{2}}"+
    "{encoding:[0-9a-fA-F]{2}}"+  // Only for versions that need encoding
    "{commitment:[0-9a-fA-F]+}",
    handler).Methods("GET")

// Route for version bytes that don't require encoding byte
router.HandleFunc("/"+
    "{optional_prefix:(?:0x)?}"+
    "{version:[0-9a-fA-F]{2}}"+   // For older versions
    "{commitment:[0-9a-fA-F]+}",
    handler).Methods("GET")
```

Idk if this is any simpler overall, but I really don't like having implicit header parsing inside the handlers. If we do go this route though please add a comment in the router regexps explaining that we will further parse the routingVarNamePayloadHex to extract the encoding byte when versionByte >= 2 inside the handler.

clients/standard_client/client.go
ethenotethan: gotcha but wouldn't there by be a cyclic issue here? https://github.com/Layr-Labs/eigenda-proxy/blob/main/go.mod#L30-L31

commitments/eigenda.go
ethenotethan: agree

commitments/mode.go
ethenotethan: that'd be nice but don't think its feasible given how the workspace is configured

server/handlers.go
ethenotethan: indicating whether the encoding query param was provided or not. Retrospectively we can enforce the misusage invariant (i.e, `v1Backend=true && encodingQueryParam=true`) in this function directly vs propagating to multiple callers. will just require making the function self referential or accessor to the server

ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

server/handlers.go
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

common/serialization.go
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

commitments/mode.go
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

commitments/eigenda.go
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

commitments/eigenda.go
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

README.md
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

server/handlers.go
```diff
@@ -250,7 +314,7 @@ func (svr *Server) handlePostShared(
 		return err
 	}
 
-	responseCommit, err := commitments.EncodeCommitment(commitment, meta.Mode, meta.Version)
+	responseCommit, err := commitments.EncodeCommitment(commitment, meta.Mode, meta.Version, meta.Encoding)
```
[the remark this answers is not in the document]
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

server/handlers.go
```diff
@@ -201,13 +253,25 @@ func (svr *Server) handlePostOPKeccakCommitment(w http.ResponseWriter, r *http.R
 
 // handlePostOPGenericCommitment handles the POST request for optimism generic commitments.
 func (svr *Server) handlePostOPGenericCommitment(w http.ResponseWriter, r *http.Request) error {
```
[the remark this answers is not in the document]
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

commitments/mode.go
```diff
@@ -31,22 +43,76 @@ func StringToCommitmentMode(s string) (CommitmentMode, error) {
...
+		return RLPEncoding, nil
+	case "abi", "ABI", "1":
+		return ABIVerifyV2CertEncoding, nil
+	default:
+		return RLPEncoding, fmt.Errorf("unknown encoding type: %s, using default RLP encoding", s)
+	}
+}
+
+// DecodeCommitmentType returns the encoding type from a certificate commitment
+func DecodeCommitmentType(cert []byte) (EigenDACommitmentType, EncodingType, []byte, error) {
```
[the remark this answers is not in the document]
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

commitments/mode.go
```diff
@@ -31,22 +43,76 @@ func StringToCommitmentMode(s string) (CommitmentMode, error) {
...
 	case OptimismGeneric:
+		// For V2 certificates with encoding specified, add encoding byte
+		if commitmentType >= CertV2 {
+			certCommit := NewEigenDACommitmentWithEncoding(bytes, commitmentType, encodingType).Encode()
+			svcCommit := EigenDASvcCommitment(certCommit).Encode()
+			altDACommit := NewGenericCommitment(svcCommit).Encode()
+			return altDACommit, nil
+		}
+
+		// Fallback to previous behavior for V0/V1
```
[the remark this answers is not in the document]
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

server/handlers.go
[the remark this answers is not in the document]
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

server/handlers.go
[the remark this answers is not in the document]
ethenotethan: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

server/handlers.go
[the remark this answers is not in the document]
ethenotethan: its easier to just enforce the misconfiguration invariant within the function directly: 04dba8c7e4368cfd65fdfc4b9bb95f480a8ec008

clients/standard_client/client.go
samlaf: server module only imports clients for testing right? I think that's allowed...

commitments/eigenda.go
```diff
@@ -21,7 +21,7 @@ type CertCommitment interface {
 type EigenDACommitment struct {
 	prefix   EigenDACommitmentType
 	b        []byte
-	encoding *EncodingType // Optional encoding type, used only for V2+
+	encoding EncodingType // Optional encoding type, used only for V2+
```
samlaf, line 24:
```suggestion
	encoding EncodingType // encoding type, used only for V2+
```

samlaf: not optional anymore right?

samlaf: also is it really only used for v2? doesnt ssem for v2 only in meta struct
> 	// encoding type for the certificate (defaults to RLPEncoding)
	Encoding EncodingType

server/handlers.go
```diff
@@ -173,28 +165,34 @@ func (svr *Server) handleGetEigenDADispersalBackend(w http.ResponseWriter, _ *ht
...
 	encodingParam := r.URL.Query().Get(routingQueryParamEncoding)
 	if encodingParam == "" {
 		// If no encoding is provided, use default RLP encoding
-		return commitments.RLPEncoding, false, nil
+		return commitments.RLPEncoding, nil
+	}
+
+	// if encoding param provided but historical V1 backend used then error
+	if svr.sm.GetDispersalBackend() == common.V1EigenDABackend {
+		http.Error(w, "Encoding type query parameter cannot be specified for V1 backend", http.StatusBadRequest)
```
samlaf, line 179: dont you need to return here?

server/handlers.go
```diff
@@ -34,19 +34,9 @@ func (svr *Server) handleHealth(w http.ResponseWriter, _ *http.Request) error {
...
-	}
-
-	commitmentMeta := commitments.CommitmentMeta{
-		Mode:     commitments.OptimismGeneric,
-		Version:  commitments.EigenDACommitmentType(versionByte),
-		Encoding: commitments.RLPEncoding,
-	}
-
+// handleGetEigenDA helper function used for processing gets for an eigenda backend type
+func (svr *Server) handleGetEigenDA(w http.ResponseWriter, r *http.Request,
```
samlaf, line 38: dont think this should be called handle because its not a routing table entrypoint

samlaf: or maybe prefix name with a `_`?

commitments/mode.go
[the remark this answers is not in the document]
samlaf: don't think that's true, see https://github.com/Layr-Labs/eigenda-proxy/pull/376

clients/standard_client/client.go
samlaf: see https://github.com/Layr-Labs/eigenda-proxy/pull/376 for an example

commitments/mode.go
samlaf: see https://github.com/Layr-Labs/eigenda-proxy/pull/376 for an example

commitments/mode.go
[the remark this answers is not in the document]
litt3: I don't see any changes related to this, let me explain a bit more what I meant:

After the block with condition `cm.Version >= CertV2`, I think a check like this would make sense:

```
		if cm.Encoding != RLPEncoding {
			return nil, fmt.Errorf("encoding type x not supported for cert version y")
		}
```

That way, if a user tries to specify an encoding type that isn't compatible with the chosen version, they will be warned.

e2e/server_test.go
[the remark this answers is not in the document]
litt3: IMO it's best practice to choose random inputs, in case there is an unanticipated relationship between input and correctness. It's also in line with other tests in this file that use random payload inputs.

Up to you though, since I agree it doesn't seem likely to have an influence in the particular case.

server/handlers.go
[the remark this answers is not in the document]
litt3: This comment wasn't addressed in this commit, this method and `handlePostStdCommitment` are still identical, apart from `CommitmentMeta.Mode`, which could be passed in as a parameter

store/generated_key/memstore/v2/memstore.go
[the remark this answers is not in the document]
litt3: Let's continue this conversation on [this thread](https://github.com/Layr-Labs/eigenda-proxy/pull/370#discussion_r2048254337), since they are getting at the same thing

store/manager.go
litt3: The current context-based solution is orthogonal to the correct parameter-based solution, IMO.

I don't think we should introduce the tech debt of context based parameters for the sake of maintaining these internal interfaces.

server/handlers.go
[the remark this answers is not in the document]
ethenotethan: wdym - any common logic here has been extracted into `parseEncodingQueryParamType`. Outside of this there's only redundancy or code overlap when setting a V2 cert commitment version

commitments/mode.go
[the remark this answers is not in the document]
samlaf: I support this. invariants should be encoded in as many places as possible

server/handlers.go
samlaf: He;s just saying that every future cert version is required to use the encoding byte. We should have enforced the encoding byte header from the start for everything, but now we're stuck with cert versions v0 and v1 not using an encoding byte.

anupsv: I see. I guess when there is V3, we'll coalesce the common functions.

PR #373, opened 2025-04-17
Update op-devnet-allocs.sh

dedyshkaPexto:
without **.git** is correct rn

you are welcome


ethenotethan:
@dedyshkaPexto why is removing the `.git` suffix correct?

samlaf:
I think github accepts both. In fact, its UI uses the `.git` suffix:
image
So this change is not necessary. Closing.

PR #376, opened 2025-04-23
example: encoding byte type alias proxy client

samlaf:
This is just an example for Ethen regarding discussions in his PR https://github.com/Layr-Labs/eigenda-proxy/pull/370

## Fixes Issue

Fixes #


ethenotethan (approved):
LGTM - never realized cyclic go mods like this were possible

PR #377, opened 2025-04-28
core(.envrc): load .env if it exists

samlaf:
This allows overwriting values in .env.exampleV1AndV2.holesky, for example for secret values

## Fixes Issue

Fixes #


litt3 (approved):

ethenotethan (approved):
LGTM

PR #378, opened 2025-04-28
feat: rollup rbn recency validation (version 2)

samlaf:
Replaces https://github.com/Layr-Labs/eigenda-proxy/pull/226, which was very old and had accumulated a lot of merge conflicts. merged master into that branch and fixed the issues.

========

We add logic to help rollup validators reject certs in their batcher inbox that don't respect the [recency invariant](https://layr-labs.github.io/eigenda/integration/spec/6-secure-integration.html#1-rbn-recency-validation):

```
cert.L1InclusionBlock - cert.RBN < RBNRecencyWindowSize
```

This feature is only included for eigenDA v2 backend, since we are only building secure integrations for v2.

## Changes proposed

- Added new flag `RBNRecencyWindowSize` (in units of L1 Blocks) which should be rollup specific
   - WARNING: this flag needs to be equal to the same value flag in the rollup's derivation pipeline (op-program, kona, etc)
   - set to 0 by default, which skips the check.
- GET requests to the proxy can include an optional `l1_inclusion_block_number` query param, which will cause a to be rejected if it was included onchain > `RBNRecencyWindowSize` L1 blocks after the blob's batch's RBN.
- When both `l1_inclusion_block_number` and `RBNRecencyWindowSize` are > 0, the above check is done
- If the check fails, a 418 TEAPOT error is returned, and rollups should treat these as CertVerificationFailed errors, and drop the cert from their derivationPipeline

### Tests
- [x] rbn recency logic test: just testing the logic and errors returned from this pure function
- [x] routing test: http requests can accept `l1_inclusion_block_number` query_param and pass it to storage `Get` function
- [ ] integration test: missing

## TODO (before merging)
- add pact contract test for 418 error?


verify/verifier.go
```diff
@@ -25,6 +26,14 @@ const (
 )
 
 type Config struct {
+	// Allowed distance (in L1 blocks) between the eigenDA reference block number (RBN) of the batch the blob is included in,
+	// and the L1 block number at which the blob cert was included in the batcher's inbox.
+	// If batch.RBN + RollupBlobInclusionWindow < cert.L1InclusionBlock, the batch is considered stale and verification will fail.
+	// This check is optional and will be skipped when RollupBlobInclusionWindow is set to 0.
+	RollupBlobInclusionWindow uint32
```
samlaf, line 33: We need to figure out how to manage this securely. See discussion in https://github.com/Layr-Labs/eigenda-proxy/pull/226#discussion_r1908276688

samlaf:
Closing as we have decided to not do this check inside proxy, and instead provide golang and rust libraries to do these checks inside the derivation pipeline directly.

verify/cli.go
```diff
@@ -20,6 +21,9 @@ var (
 	G2PathFlagName                   = withFlagPrefix("g2-path")
 	G2TrailingPathFlagName           = withFlagPrefix("g2-path-trailing")
 	CachePathFlagName                = withFlagPrefix("cache-path")
+
+	// rollup related flags
+	RollupBlobInclusionWindowFlagName = withFlagPrefix("rollup-blob-inclusion-window")
```
samlaf, line 26: don't think this should be here.. since `verify/` is v1 only, but we also need to do this check for v2. Maybe should have a `rollup` category?

common/store.go
```diff
@@ -71,8 +71,21 @@ func StringToBackendType(s string) BackendType {
...
+	// This is optional, and should be set to 0 to mean to not verify the reference block number distance check.
+	// Its impossible for a batch inbox tx to have been included in the genesis block,
+	// so we are free to give this special meaning to the zero value.
+	//
+	// Used to determine the validity of the eigenDA batch.
+	// The eigenDA batch header contains a reference block number (RBN) which is used
+	// to lookup the stake of the eigenda operators before verifying signature thresholds.
+	// The rollup batch containing the eigenDA cert is only valid if it was included
+	// within a certain number of blocks after the RBN.
+	// validity condition is: RBN < l1_inclusion_block_number < RBN + rollup_blob_inclusion_window
```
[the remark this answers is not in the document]
samlaf: regarding default value, prefer to default to 0 (no check) right now given that this should be opt in feature since we're rolling it out late. Perhaps in the future we can set a real default, or pass the value dynamically from the op-node to batcher

verify/cli.go
samlaf: [de36b93](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/de36b938fec11748571d35abd501ef09bf5a37d1)

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +149,59 @@ func (e Store) BackendType() common.BackendType {
...
+//     not be slashable anymore, even though because of the old RBN their signature would count for a lot of stake.
+//
+// Note that for a secure integration, this same check needs to be verified onchain.
+// There are 2 approaches to doing this:
+//  1. Pessimistic approach: use a smart batcher inbox to dissalow stale blobs from even beign included
+//     in the batcher inbox (see https://github.com/ethereum-optimism/design-docs/pull/229)
+//  2. Optimistic approach: verify the check in op-program or hokulea (kona)'s derivation pipeline. See
+//
+// https://github.com/Layr-Labs/hokulea/blob/8c4c89bc4f35d56a3cec2220575a9681d987105c/crates/eigenda/src/eigenda.rs#L90
+func (e Store) verifyRBNRecencyCheck(certRBN uint32, opts common.VerifyOpts) error {
```
samlaf, line 187: should this check be moved to the client itself? cc @litt3

samlaf:
@bxue-l2 @litt3 think this is ready for a review. There's been a lot of back and forth and merging master and rebases and what not in this, so its possible there's some weird stuff left hanging around. Please point them out! Otherwise I think the main open question left for me is https://github.com/Layr-Labs/eigenda-proxy/pull/378#discussion_r2085291290

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +149,59 @@ func (e Store) BackendType() common.BackendType {
...
+	if err != nil {
+		return fmt.Errorf("punctuality check failed: %w", err)
+	}
 
 	return e.verifier.VerifyCertV2(ctx, &eigenDACert)
 }
+
+// Certs in the rollup batcher-inbox that do not respect the below equation are discarded.
+//
+//	cert.RBN < cert.L1InclusionBlock <= cert.RBN + RBNRecencyWindowSize
```
litt3, line 168: Can't remember if we'd had this discussion already-

Is `cert.RBN < cert.L1InclusionBlock` the right condition? Why shouldn't we permit `cert.RBN <= cert.L1InclusionBlock`?

verify/cert.go
```diff
@@ -22,7 +22,15 @@ import (
...
-// CertVerifier verifies the DA certificate against on-chain EigenDA contracts
+type certVerifier interface {
+	verifyBatchConfirmedOnChain(ctx context.Context, batchID uint32, batchMetadata *disperser.BatchMetadata) error
+	verifyMerkleProof(inclusionProof []byte, root []byte, index uint32, blobHeader BlobHeader) error
+	// getters for the cached quorum parameters
+	quorumNumbersRequired() []uint8
+	quorumAdversaryThresholdPercentages(quorum uint8) (percentage uint8, ok bool)
+}
+
+// certVerification verifies the DA certificate against on-chain EigenDA contracts
```
litt3, line 33: Was it intentional to change this to say `certVerification` here, instead of saying `CertVerifier`?

verify/verifier.go
```diff
@@ -43,13 +54,29 @@ func (c Config) MarshalJSON() ([]byte, error) {
...
 }
 
+// Verifier verifies the integrity of a blob and its certificate. There are 3 main categories of verification:
+//  1. Blob: properties localized to a single blob, most specifically its commitment
+//  2. Batch: properties wrt the EigenDA network: signatures, quorum thresholds, onchain inclusion, etc.
+//  3. Rollup: properties wrt the rollup chain: block at which the cert was included in the batcher's inbox
+//     (make sure its not too stale trying to game fraud proof windows)
+//
 // TODO: right now verification and confirmation depth are tightly coupled. we should decouple them
 type Verifier struct {
```
litt3, line 64: I'm curious why you are adding the recency check to this V1 logic.

Wouldn't it be ok to leave V1 entirely as-is, and *only* worry about the recency check for V2?

store/generated_key/v2/eigenda.go
samlaf: Ahh ok I wasn't aware that our v2 clients were not doing verification. We synced offline, and think we should at smoe point create a `CertVerifyingPayloadRetrieverClient` which would do everything proxy is doing (validate cert, then proceed to query blob from relay + nodes if needed).

At that point we could move this verification logic to that client in a single swoop cleanup. Agree?

verify/verifier.go
samlaf: yes I actually totally agree... I started this PR loong ago before we had v2. So just brought it back to life with me. I think it might actually make sense to just remove the V1 logic entirely to keep proxy simple...

store/generated_key/v2/eigenda.go
litt3: Yeah, this sounds good to me

verify/verifier.go
```diff
@@ -61,47 +88,83 @@ func NewVerifier(cfg *Config, kzgVerifier *kzgverifier.Verifier, l logging.Logge
...
+	// any external calls to the blockchain. This is a sanity check that should always be performed.
+
+	// 1 - verify that the blob cert was submitted to the rollup's batch inbox within the allowed
+	// window of size RBNRecencyWindowSize. This is to prevent timing attacks where a rollup batcher
+	// could try to game the fraud proof window by including an old DA blob that is about to expire
+	// on the DA layer and is hence not retrievable.
+	//
+	// Note that for a secure integration, this same check needs to be verified onchain.
+	// There are 2 approaches to doing this:
+	//  1. Pessimistic approach: use a smart batcher inbox to dissalow stale blobs from even beign included
```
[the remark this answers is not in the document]
samlaf: 674a41cc

common/store.go
```diff
@@ -71,8 +71,21 @@ func StringToBackendType(s string) BackendType {
...
 	// BackendType returns the backend type provider of the store.
 	BackendType() BackendType
-	// Verify verifies the given key-value pair.
-	Verify(ctx context.Context, serializedCert []byte, payload []byte) error
+}
+
+type VerifyOpts struct {
+	// L1 block number at which the rollup batch was included in the batcher inbox.
+	// This is optional, and should be set to 0 to mean to not verify the reference block number recency check.
+	// Its impossible for a batch inbox tx to have been included in the genesis block,
```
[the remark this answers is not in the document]
samlaf: 674a41cc

store/generated_key/eigenda/eigenda.go
```diff
@@ -181,7 +181,10 @@ func (e Store) Put(ctx context.Context, value []byte) ([]byte, error) {
 		return nil, fmt.Errorf("failed to verify commitment: %w", err)
 	}
 
-	err = e.verifier.VerifyCert(ctx, cert)
+	// we set RollupL1InclusionBlockNum to -1 to skip the check, as it is meaningless in the PUT route.
```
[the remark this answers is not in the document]
samlaf: 674a41cc

store/generated_key/eigenda/eigenda.go
```diff
@@ -214,7 +217,8 @@ func (e Store) BackendType() common.BackendType {
 
 // Key is used to recover certificate fields and that verifies blob
 // against commitment to ensure data is valid and non-tampered.
-func (e Store) Verify(ctx context.Context, serializedCert []byte, payload []byte) error {
+// l1InclusionBlockNum is optional and used to validate the certificate: negative number means don't verify this check
```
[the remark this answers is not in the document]
samlaf: 674a41cc

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +149,59 @@ func (e Store) BackendType() common.BackendType {
...
-func (e Store) Verify(ctx context.Context, certBytes []byte, _ []byte) error {
+func (e Store) Verify(ctx context.Context, certBytes []byte, _ []byte, opts common.VerifyOpts) error {
 	var eigenDACert coretypes.EigenDACert
 	err := rlp.DecodeBytes(certBytes, &eigenDACert)
 	if err != nil {
 		return fmt.Errorf("RLP decoding EigenDA v2 cert: %w", err)
 	}
+	err = e.verifyRBNRecencyCheck(eigenDACert.BatchHeader.ReferenceBlockNumber, opts)
+	if err != nil {
+		return fmt.Errorf("punctuality check failed: %w", err)
```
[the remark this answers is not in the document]
samlaf: 674a41cc

verify/cert.go
samlaf: 674a41cc

server/handlers.go
```diff
@@ -105,25 +106,35 @@ func (svr *Server) handleGetOPGenericCommitment(w http.ResponseWriter, r *http.R
...
 	versionedCert := certs.NewVersionedCert(commitment, certVersion)
 
-	return svr.handleGetShared(r.Context(), w, versionedCert, commitments.OptimismGenericCommitmentMode)
+	return svr.handleGetShared(r.Context(), w, r, versionedCert, commitments.OptimismGenericCommitmentMode)
 }
 
 func (svr *Server) handleGetShared(
 	ctx context.Context,
 	w http.ResponseWriter,
+	r *http.Request,
```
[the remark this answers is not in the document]
samlaf: hmm interesting I had never noticed. Was assuming `request` itself implemented Reader interface but actually only `r.Body` does. Nonetheless we have this convention on all of our handlers, and are just following the gorilla mux handler pattern here: https://github.com/gorilla/mux
image

Might change all of them to `req` later if I feel like it but want to focus on actual PR logic for now if you don't mind.

store/generated_key/v2/eigenda.go
samlaf: You had pointed this out actually, its impossible to have =, because RBN is used by validators to verify the dispersal request and sign on it. So has to be in the past. And all of this happens before the cert can be sent to batcher inbox

samlaf:
@litt3 went ahead and resolved all the typo fixes threads (the commit is linked in each in case you do want to verify). Was getting lost among all the comments. I think this is a policy we can be fine with going forward (fine for author to resolve trivial typo-fixes requests)?

store/generated_key/v2/eigenda.go
samlaf: [df30640](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/df30640ccda39ea3584abae02a387a7c20993db0)

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +149,59 @@ func (e Store) BackendType() common.BackendType {
...
+//
+// Note that for a secure integration, this same check needs to be verified onchain.
+// There are 2 approaches to doing this:
+//  1. Pessimistic approach: use a smart batcher inbox to dissalow stale blobs from even beign included
+//     in the batcher inbox (see https://github.com/ethereum-optimism/design-docs/pull/229)
+//  2. Optimistic approach: verify the check in op-program or hokulea (kona)'s derivation pipeline. See
+//
+// https://github.com/Layr-Labs/hokulea/blob/8c4c89bc4f35d56a3cec2220575a9681d987105c/crates/eigenda/src/eigenda.rs#L90
+func (e Store) verifyRBNRecencyCheck(certRBN uint32, opts common.VerifyOpts) error {
+	if opts.RollupL1InclusionBlockNum > 0 && e.rbnRecencyWindowSize > 0 {
```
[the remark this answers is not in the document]
samlaf: oh this is a very valid nit actually. 6bab9b10

verify/verifier.go
```diff
@@ -43,13 +54,29 @@ func (c Config) MarshalJSON() ([]byte, error) {
...
 
+// Verifier verifies the integrity of a blob and its certificate. There are 3 main categories of verification:
+//  1. Blob: properties localized to a single blob, most specifically its commitment
+//  2. Batch: properties wrt the EigenDA network: signatures, quorum thresholds, onchain inclusion, etc.
+//  3. Rollup: properties wrt the rollup chain: block at which the cert was included in the batcher's inbox
+//     (make sure its not too stale trying to game fraud proof windows)
+//
 // TODO: right now verification and confirmation depth are tightly coupled. we should decouple them
 type Verifier struct {
+	log logging.Logger
```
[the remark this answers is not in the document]
samlaf: c2021ec3

litt3:
> @litt3 went ahead and resolved all the typo fixes threads (the commit is linked in each in case you do want to verify). Was getting lost among all the comments. I think this is a policy we can be fine with going forward (fine for author to resolve trivial typo-fixes requests)?

I'm ok if you want to resolve typo fixes.

What drives me up the wall is just when I make a typo comment, the PR owner "fixes" it and resolves, but then it turns out it's not actually fixed. I wouldn't expect this from you, though.

My personal strategy is just to collapse comments that I've addressed, without resolving them.

store/storage_manager_builder.go
```diff
@@ -288,6 +288,7 @@ func (smb *StorageManagerBuilder) buildEigenDAV2Backend(
 	eigenDAV2Store, err := eigenda_v2.NewStore(
 		smb.log,
 		smb.v2ClientCfg.PutTries,
+		smb.v1VerifierCfg.RBNRecencyWindowSize,
```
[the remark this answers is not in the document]
samlaf: Completely removed v1 verification, so moved this config to v2ClientCfg. Can't link the actual commit because I chnaged/cleanup a lot of stuff..... think you'll need to re-review the whole thing (don't think it'll be that bad).

verify/verifier.go
samlaf: Ok pulled the trigger. Completely removed v1 verification, so moved this config to v2ClientCfg. Can't link the actual commit because I chnaged/cleanup a lot of stuff..... think you'll need to re-review the whole thing (don't think it'll be that bad).

verify/verifier.go
```diff
@@ -43,6 +47,13 @@ func (c Config) MarshalJSON() ([]byte, error) {
 	return json.Marshal(aux)
 }
 
+// Verifier verifies the integrity of a blob and its certificate. There are 3 main categories of verification:
```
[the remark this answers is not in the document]
litt3: I didn't read closely enough. Makes sense 👍

store/generated_key/eigenda/eigenda.go
```diff
@@ -181,6 +181,9 @@ func (e Store) Put(ctx context.Context, value []byte) ([]byte, error) {
 		return nil, fmt.Errorf("failed to verify commitment: %w", err)
 	}
 
+	// we set RollupL1InclusionBlockNum to 0 to skip the check, as it is meaningless in the PUT route.
```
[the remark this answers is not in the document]
samlaf: [9217fe7](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/9217fe71b7b07d54e6bc5967becad478d2786c2b)

store/generated_key/eigenda/eigenda.go
```diff
@@ -214,7 +217,8 @@ func (e Store) BackendType() common.BackendType {
 
 // Key is used to recover certificate fields and that verifies blob
 // against commitment to ensure data is valid and non-tampered.
-func (e Store) Verify(ctx context.Context, serializedCert []byte, payload []byte) error {
+// l1InclusionBlockNum is optional and used to validate the certificate: 0 means don't verify this check
```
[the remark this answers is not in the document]
samlaf: [a99cab6](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/a99cab6d73f66e1d965dc56d4a25977f73e141c7)

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +163,74 @@ func (e Store) BackendType() common.BackendType {
...
+		opts.L1InclusionBlockNum, e.rbnRecencyWindowSize)
+	if err != nil {
+		return fmt.Errorf("rbn recency check failed: %w", err)
+	}
 
 	return e.verifier.VerifyCertV2(ctx, &eigenDACert)
 }
+
+// verifyCertRBNRecencyCheck arguments:
+//   - certRBN: ReferenceBlockNumber included in the cert iself at which operator stakes are referenced
```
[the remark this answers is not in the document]
samlaf: [0a11032](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/0a11032c535784d6f417ffafc0bdbf2ee18c78a1)

store/generated_key/v2/eigenda.go
```diff
@@ -18,14 +18,35 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+type RBNRecencyCheckFailedError struct {
```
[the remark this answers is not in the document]
samlaf: [0db4ab6](https://github.com/Layr-Labs/eigenda-proxy/pull/378/commits/0db4ab6c2c60b6e547efe4352c9a21a8b55504ea)

litt3 (approved):

litt3 (dismissed):

server/handlers.go
```diff
@@ -136,6 +158,25 @@ func (svr *Server) handleGetShared(
...
 
+// Parses the l1_inclusion_block_number query param from the request.
+// Happy path:
+//   - if the l1_inclusion_block_number is provided, it returns the parsed value.
+//
+// Unhappy paths:
+//   - if the l1_inclusion_block_number is not provided, it returns 0 (whose meaning is to skip the check).
+//   - if the l1_inclusion_block_number is provided but isn't a valid integer, it returns an error.
+func parseCommitmentInclusionL1BlockNumQueryParam(r *http.Request) (uint64, error) {
+	l1BlockNumStr := r.URL.Query().Get("l1_inclusion_block_number")
```
ethenotethan, line 169: Nit - L1 isn't true for L3s

.env.exampleV1AndV2.holesky
```diff
@@ -97,6 +97,13 @@ EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 # V2_SERVICE_MANAGER_ADDR, and V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR.
 EIGENDA_PROXY_EIGENDA_V2_NETWORK=holesky_testnet
 
+# Allowed distance (in L1 blocks) between the eigenDA cert's reference block number (RBN)
+# and the L1 block number at which the cert was included in the rollup's batch inbox.
+# If certL1InclusionBlock > cert.RBN + rbnRecencyWindowSize, the cert
+# is considered stale and verification will fail.
+# This check is optional and will be skipped when set to 0.
+EIGENDA_PROXY_EIGENDA_V2_RBN_RECENCY_WINDOW_SIZE=0
```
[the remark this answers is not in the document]
samlaf: It's not a sequencer property, its a rollup property. It has to be set in the rollup.json config, and used everywhere the same (proxy, as well as OSP). We could make it more robust but I opted for this simpler approach to start with (since I think the recency window is the least important and least likely to get hit of the properties).

Few approaches to robustify this in the future:
1. We could add an endpoint on proxy that returns the recency window. derivation pipeline upon startup would query it and make sure its equivalent to its rollup.json field. If its not the DP would stop.
2. Otherwise we could pass the value as another query_param in the queries everytime

WDYT? cc @bxue-l2

server/handlers.go
samlaf: Are you suggesting we use "settlement_inclusion_block_number" instead? That could work. Or remove the l1 entirely? Or `commitment_inclusion_block_number`?

.env.exampleV1AndV2.holesky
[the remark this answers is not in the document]
bxue-l2: I actually prefer to have it on certVerifier contract. So it does not have to copy Sequencing_window_size on OP. 

Just conceptually, RECENCY_WINDOW_SIZE is an EigenDA property. which needs to be used by Hokulea, Proxy, and Arbitrum onchain one step proof.

server/handlers.go
```diff
@@ -105,27 +107,47 @@ func (svr *Server) handleGetOPGenericCommitment(w http.ResponseWriter, r *http.R
...
 			fmt.Errorf("get request failed with serializedCert %v: %w", serializedCertHex, err),
-			versionedCert.Version,
-			mode,
-		)
-		if errors.Is(err, ErrNotFound) {
-			http.Error(w, err.Error(), http.StatusNotFound)
+			versionedCert.Version, mode)
+		var rbnRecencyCheckFailedErr eigendav2store.RBNRecencyCheckFailedError
+		if errors.As(err, &rbnRecencyCheckFailedErr) {
+			// We return a 418 TEAPOT error for any cert validation error.
```
bxue-l2, line 146: though unrelated. Consider a case, that cert is valid, and recency is good. It is just the blob encoding has problem. In that case, hokulea has to deal with that in the derivation pipeline. Question, can proxy return such invalid encoding cert? 
Currently, proxy only returns the decoded data to hokulea.

server/routing.go
```diff
@@ -27,20 +27,28 @@ func (svr *Server) RegisterRoutes(r *mux.Router) {
 		withLogging(withMetrics(svr.handleGetStdCommitment, svr.m, commitments.StandardCommitmentMode), svr.log),
 	).Queries("commitment_mode", "standard")
 	// op keccak256 commitments (write to S3)
-	subrouterGET.HandleFunc("/"+
-		"{optional_prefix:(?:0x)?}"+ // commitments can be prefixed with 0x
-		"{"+routingVarNameCommitTypeByteHex+":00}"+ // 00 for keccak256 commitments
-		"{"+routingVarNameKeccakCommitmentHex+"}",
-		withLogging(withMetrics(svr.handleGetOPKeccakCommitment, svr.m, commitments.OptimismKeccakCommitmentMode), svr.log),
+	subrouterGET.HandleFunc(
+		"/"+
```
bxue-l2, line 31: why I felt formatting is worse than before

common/store.go
```diff
@@ -68,11 +68,24 @@ func StringToBackendType(s string) BackendType {
...
+	// This is optional, and should be set to 0 to mean to skip the RBN recency check.
+	// It is impossible for a batch inbox tx to have been included in the genesis block,
+	// so we are free to give this special meaning to the zero value.
+	//
+	// Used to determine the validity of the eigenDA batch.
+	// The eigenDA cert contains a reference block number (RBN) which is used
+	// to lookup the stake of the eigenda operators before verifying signature thresholds.
+	// The rollup commitment containing the eigenDA cert is only valid if it was included
+	// within a certain number of blocks after the RBN.
+	// validity condition is: certRBN < L1InclusionBlockNum < RBN + RBNRecencyWindowSize
```
bxue-l2, line 82: I think in this file store/generated_key/v2/eigenda.go, the condition is 

certRBN < L1InclusionBlockNum <= RBN + RBNRecencyWindowSize

the equal sign

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +167,74 @@ func (e Store) BackendType() common.BackendType {
...
+//     https://github.com/Layr-Labs/hokulea/blob/8c4c89bc4f/crates/eigenda/src/eigenda.rs#L90
+func verifyCertRBNRecencyCheck(certRBN uint32, certL1IBN uint64, rbnRecencyWindowSize uint64) error {
+	// Input Validation
+	if certL1IBN == 0 || rbnRecencyWindowSize == 0 {
+		return nil
+	}
+	if certRBN == 0 {
+		return fmt.Errorf("certRBN should never be 0, this is likely a bug")
+	}
+	if !(uint64(certRBN) < certL1IBN) {
```
bxue-l2, line 222: Nit. CertRBN >= certL1IBN. hehe

common/store.go
bxue-l2: I think in the spec, you wrote, 

cert.L1InclusionBlock - cert.RBN < RecencyWindowSize

bxue-l2: https://github.com/Layr-Labs/eigenda/blob/master/docs/spec/src/integration/spec/6-secure-integration.md

.env.exampleV1AndV2.holesky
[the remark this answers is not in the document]
samlaf: Hmm that would be a good usecase for testing our CertVerifier abi changes (adding a new RecencyWindow storage+getter).

But how would hokulea get access to this value? It would be hardcoded in the binary and checked onchain?
For op-program I feel like this wouldnt work..? We'd have to load it at boot time right? But op-program doesn't have access to certVerifier address (only indirectly via mips.sol pointing to preimageoracle which points to router)

server/handlers.go
samlaf: Prob better to talk this one out using voice call.
1. wdym by "blob encoding has problem"? If a cert is valid, doesn't it commit to a single possible blob (let's say that of length = blob_length)? Or are you saying the payload is badly encoded inside the blob?
2. wdym by "proxy return such invalid encoding cert"? Is it the cert or blob that is invalid

Is the case you're saying just "cert is valid, but proxy is unable to download/fetch (from relay and validators) a blob that matches the commitment)? I think that should just return a 404 error no?

server/routing.go
samlaf: Think its just github UI. Formatting looks fine on my computer. getting deterministic formatting is so damn hard... idk if I updated my formatter or what... just ran `make format` and this is what came out. We're prob not enforcing it via CI, which is why it popped up in this pr

common/store.go
samlaf: good catch, fixed spec: https://github.com/Layr-Labs/eigenda/pull/1567

store/generated_key/v2/eigenda.go
samlaf: I prefer this way because it keeps the positive case (inside !(...)) highlighted, which is something that makes more semantic sense (because certRBN should always be before the block at which cert was included on L1) and is thus easier to parse visually.

I think other than for trivial cases (==0), sanization should always look like:
```
if !(some_property_we_need) {
   return err
}
if !(some_other_property_we_need) {
   return err
}
```

server/handlers.go
samlaf: hmm I agree but I'm also torn because op spec for example uses l1 terminology: https://specs.optimism.io/protocol/derivation.html#l1-traversal
Let's discuss during FFT?

common/store.go
bxue-l2: lgtm

.env.exampleV1AndV2.holesky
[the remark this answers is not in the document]
bxue-l2: Thinking it is checked onchain, routed by RBN (if for any reason it is changed).

For op-program, the host part can verify the recency_window and give empty response. When it is in the contract challenge mode, the onchain contract can take inputs of (inclusionBN, and CertRBN) and return empty value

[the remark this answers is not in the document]
samlaf: 3 approaches that we discussed in FFT today:
CFF5E2F9-D071-4200-A464-6DE61B396F70

Basically we need the triplet (RWS, RBN, IBN) to be together to compare them: either they can be together in the VM, or in the contract.

Seems like the simplest approach (h/t @litt3) is just to load define `RWS := SequencerWindowSize`, and since we already have SWS in the vm (loaded during bootinfo), then we have access to everything there already. Only downside (as we've already discussed) is that this requires parsing the RBN of the cert inside the derivation pipeline.

server/handlers.go
samlaf: Actually realized I can't change it because the check doesn't even make sense on L3s! Created this issue: https://linear.app/eigenlabs/issue/DAINT-498/figure-out-how-to-do-rbn-recency-check-on-l3s

bxue-l2 (approved):

server/handlers.go
bxue-l2: yeah good point

litt3 (approved):

Issue #372, opened 2025-04-17
On-demand payment account obervability


ethenotethan:
Telemetry should be provided which gives a rollup using V2 the ability to view their payment account's balance.

ethenotethan:
one mechanism for which the balance can be determined is calling read [methods](https://holesky.etherscan.io/address/0x4a7Fff191BCDa5806f1Bc8689afc1417c08C61AB#readProxyContract) on the PaymentVault contract

samlaf:
> one mechanism for which the balance can be determined is calling read [methods](https://holesky.etherscan.io/address/0x4a7Fff191BCDa5806f1Bc8689afc1417c08C61AB#readProxyContract) on the PaymentVault contract

Indeed... good catch. Right now we are [querying the eth balance of the accountID](https://github.com/Layr-Labs/eigenda-proxy/blob/55ed00fe74a4ca03f736310cc3d603eba083f32d/store/storage_manager_builder.go#L503), which is totally unrelated to whether one can make payments on eigenda...

PR #380, opened 2025-05-01
chore: use mise to manage deps

### LINKED ISSUE #236: Create reproducible dev env

samlaf:
## Fixes Issue

Fixes #236 

## Changes proposed

Replaces #326 
Decided to opt for mise instead of nix. Much much easier and it's enough for our current use cases.
Also I'm much more familiar with it now that I've had to use it in our op-fork and some other places.

To use it, just install mise: https://mise.jdx.dev/getting-started.html
then run `mise install`, and that's pretty much it.


samlaf:
@anupsv any idea what that weird codeql error is?

litt3 (approved):
LGTM

store/storage_manager_builder.go
```diff
@@ -500,7 +500,9 @@ func (smb *StorageManagerBuilder) buildLocalSigner(
 	}
 
 	accountID := crypto.PubkeyToAddress(signer.PrivateKey.PublicKey)
-	pendingBalance, err := ethClient.PendingBalanceAt(ctx, accountID)
```
ethenotethan, line 503: off topic but this pending balance doesn't do anything - we should instead be querying the `PaymentVault` directly. Updated here: https://github.com/Layr-Labs/eigenda-proxy/issues/372

mise.toml
```diff
@@ -0,0 +1,16 @@
...
+go = "1.22.7"
+
+# Go Dependencies
+"go:github.com/golangci/golangci-lint/cmd/golangci-lint" = "1.64.5"
+"go:go.uber.org/mock/mockgen" = "0.5.0"
+"go:gotest.tools/gotestsum" = "1.12.0"
+"go:github.com/segmentio/golines" = "0.12.0"
+
+# Forge Dependencies
+cast = "v1.0.0"
```
ethenotethan, line 13: why do we need forge as a dependency?

mise.toml
```diff
@@ -0,0 +1,16 @@
+[tools]
+
+# Core Dependencies
+go = "1.22.7"
+
+# Go Dependencies
+"go:github.com/golangci/golangci-lint/cmd/golangci-lint" = "1.64.5"
+"go:go.uber.org/mock/mockgen" = "0.5.0"
+"go:gotest.tools/gotestsum" = "1.12.0"
+"go:github.com/segmentio/golines" = "0.12.0"
```
ethenotethan, line 10: oh nice so mise provides a wrapper framework on top of dev dependencies that allows for steamlined version mgmt vs having to inexplicitly manage them across things like Make and CI?

ethenotethan (approved):
LGTM

mise.toml
samlaf: don't remember but had listed it in the issue ticket. I'm guessing some makefile target or some command in some script uses cast somewhere.

store/storage_manager_builder.go
samlaf: Oh snap good catch... cc @litt3

mise.toml
samlaf: Yep exactly and the mise github action that we use in ci caches all the dependencies so there's no extra incurred download cost on every ci run.

samlaf:
Merged despite preprod ci failing because preprod is borked right now. Tested ci passed anyways which gives us confidence for the code in here. Plus its just dependency management, no code changes.

store/storage_manager_builder.go
litt3: Yeah, this check was half baked. Probably should just be removed, until we properly consider what tools proxy needs to make payments easier for users

PR #382, opened 2025-05-02
Update codeql-scanning.yml

anupsv:
moving to a central codeql scanner config

Fixes #

### DIFF
--- a/mocks/manager.go
+++ b/mocks/manager.go
@@ -72,6 +72,21 @@ func (mr *MockIManagerMockRecorder) GetDispersalBackend() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetDispersalBackend", reflect.TypeOf((*MockIManager)(nil).GetDispersalBackend))
 }
 
+// GetOPKeccakValueFromS3 mocks base method.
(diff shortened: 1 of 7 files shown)


samlaf (approved):
LGTM also I would replace the golang installation with mise install (to use same golang version as everywhere else). See https://github.com/Layr-Labs/eigenda-proxy/blob/c153916b31a82d0143b8500bc3f9f65421f8229b/.github/workflows/per-pr.yml#L17

Also I think there's something wrong with how codeql is setup..? because one of the CI isn't running:
image

anupsv:
> LGTM although I would replace the golang installation with mise install (to use same golang version as everywhere else). See
> 
> https://github.com/Layr-Labs/eigenda-proxy/blob/c153916b31a82d0143b8500bc3f9f65421f8229b/.github/workflows/per-pr.yml#L17
> 
> Also I think there's something wrong with how codeql is setup..? because one of the CI isn't running: image

added :)

samlaf (approved):
LGTM, although you haven't defined the `MISE_VERSION` env var, so I'm guessing its just defaulting to latest?

anupsv:
> LGTM, although you haven't defined the `MISE_VERSION` env var, so I'm guessing its just defaulting to latest?

Yea, but i can set it specifically. I wonder if it's better to set this version as a git env so that if it changes there it'll change everywhere else.

samlaf (approved):

PR #384, opened 2025-05-02
feat: Cert Verifier Router Support

ethenotethan:
## Fixes Issue

Fixes #

## Changes proposed
  - Adds support for the EigenDA Certificate Verifier Router contract, allowing the proxy to verify certificates using both legacy and v3 router/verifiers for cross compatibility with existing testnets

 **Key Changes**

  - Added support for certificate verifier router with backward compatibility for legacy certificate verification
  - Implemented RBN recency window validation to prevent stale certificates
  - Updated CI workflows to use mise for dependency management
  - Enhanced E2E tests to validate certificate verification through router


common/client_config_v2.go
```diff
@@ -51,15 +52,16 @@ type ClientConfigV2 struct {
 // Check checks config invariants, and returns an error if there is a problem with the config struct
 func (cfg *ClientConfigV2) Check() error {
```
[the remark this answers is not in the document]
ethenotethan: [b706d25](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/b706d25c08deeab8eb8161b9d235a98138e75f55)

common/store.go
```diff
@@ -100,7 +102,18 @@ type EigenDAStore interface {
 	Verify(ctx context.Context, serializedCert []byte, payload []byte, opts CertVerificationOpts) error
 }
 
-// SecondaryStore is the interface for a key-value data store that uses keccak(value) as the key.
+type EigenDAV2Store interface {
```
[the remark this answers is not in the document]
ethenotethan: [b706d25](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/b706d25c08deeab8eb8161b9d235a98138e75f55)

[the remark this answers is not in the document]
ethenotethan: [b706d25](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/b706d25c08deeab8eb8161b9d235a98138e75f55)

common/eigenda_network.go
```diff
@@ -66,6 +52,19 @@ func (n EigenDANetwork) GetBLSOperatorStateRetrieverAddress() (string, error) {
...
 
+// GetRegistryCoordinatorAddress returns, as a string, the address of the RegistryCoordinator contract for the
+// network
+func (n EigenDANetwork) GetRegistryCoordinatorAddress() (string, error) {
+	switch n {
+	case HoleskyTestnetEigenDANetwork:
+		return "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490", nil
+	case HoleskyPreprodEigenDANetwork:
+		return "0x2c61EA360D6500b58E7f481541A36B443Bc858c6", nil
+	default:
```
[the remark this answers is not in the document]
ethenotethan: [b706d25](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/b706d25c08deeab8eb8161b9d235a98138e75f55)

common/types/certs/eigenda.go
```diff
@@ -10,6 +14,7 @@ const (
 	V0VersionByte VersionByte = iota
 	// All future CertVersions will be against EigenDA V2 Blazar (https://docs.eigenda.xyz/releases/blazar)
 	V1VersionByte
+	V2VersionByte
 )
 
 func ByteToVersion(b byte) (VersionByte, error) {
```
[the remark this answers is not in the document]
samlaf: Hmm good point, especially as we plan to move proxy to monorepo.
core type is a uint64, whereas here its a byte. The point of this proxy type is because we parse the byte header when we read certs. But prob we can just parse the byte into a uint64 directly?

common/client_config_v2.go
```diff
@@ -23,9 +23,9 @@ type ClientConfigV2 struct {
 	// - If > 0: Try N times total
 	// - If < 0: Retry indefinitely until success
 	// - If = 0: Not permitted
-	PutTries                   int
-	MaxBlobSizeBytes           uint64
-	EigenDACertVerifierAddress string
+	PutTries                         int
+	MaxBlobSizeBytes                 uint64
+	EigenDACertVerifierRouterAddress string // >= V3 cert
```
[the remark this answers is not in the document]
samlaf: Def needs to be documented at the very least. Maybe `EigenDACertVerifierOrRouterAddress`?

common/client_config_v2.go
```diff
@@ -35,8 +35,9 @@ type ClientConfigV2 struct {
 	RetrieversToEnable []RetrieverType
 
 	// Fields required for validator payload retrieval
-	BLSOperatorStateRetrieverAddr string
-	EigenDAServiceManagerAddr     string
+	BLSOperatorStateRetrieverAddr  string
+	EigenDARegistryCoordinatorAddr string
+	EigenDAServiceManagerAddr      string
```
samlaf, line 40: if these are required for validator payload retrieval, why are they not part of the payload retriever config? I completely forget why this is separate, might be good to comment.

common/store.go
```diff
@@ -88,8 +90,7 @@ type Store interface {
 	BackendType() BackendType
 }
 
-// EigenDAStore is the interface for an EigenDA data store, which stores payloads that are retrievable
-// from a DACert. Implementations include EigenDA V1 and V2, as well as their memstore versions for testing.
+// EigenDAStore is the interface for an EigenDA V1 data store as well as V1 memstore.
 type EigenDAStore interface {
```
samlaf, line 94: Let's call it `EigenDAV1Store` then?

common/types/certs/eigenda.go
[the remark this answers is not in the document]
samlaf: ALso idk if we do it in core but might be good to add a small description of what the differences between the different versions are. We at least need this somewhere in some centralized place, and maybe we can point to that doc from everywhere, including here?

config/v2/eigendaflags/cli.go
```diff
@@ -128,6 +129,13 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			Category: category,
 			Required: false,
 		},
+		&cli.StringFlag{
+			Name:     RegistryCoordinatorAddrFlagName,
+			Usage:    "Address of the registry coordinator contract.",
+			EnvVars:  []string{withEnvPrefix(envPrefix, "REGISTRY_COORDINATOR_ADDR")},
+			Category: category,
+			Required: false,
+		},
```
samlaf, line 138: this can be fetched from the ServiceManager right?

server/handlers.go
```diff
@@ -284,7 +284,7 @@ func (svr *Server) handlePostShared(
 	case common.V1EigenDABackend:
 		certVersion = certs.V0VersionByte
 	case common.V2EigenDABackend:
-		certVersion = certs.V1VersionByte
+		certVersion = certs.V2VersionByte
```
samlaf, line 287: add comment here that dispersals only support the latest version.
Is that really what we want though? What if a rollup is using an old certVerifier...?
Shouldnt we query the latest cert version their certverifier (or router) supports and then disperse to that?

store/builder/storage_manager_builder.go
```diff
@@ -206,19 +210,52 @@ func buildEigenDAV2Backend(
...
+	if err != nil {
+		return nil, fmt.Errorf("new cert verifier router caller: %w", err)
+	}
 
-	certVerifier, err := verification.NewCertVerifier(
-		log, ethClient, certVerifierAddressProvider)
+	isRouter := true
+	// Check if the router address is actually a router. if method `getRouterAddress` fails, it means that the
+	// address is not a router, and we should treat it as an immutable cert verifier instead
+	_, err = caller.GetCertVerifierAt(&bind.CallOpts{Context: ctx}, 0)
```
samlaf, line 222:
```suggestion
	// Check if the router address is actually a router. if method `GetCertVerifierAt` fails, it means that the
	// address is not a router, and we should treat it as an immutable cert verifier instead
	_, err = caller.GetCertVerifierAt(&bind.CallOpts{Context: ctx}, 0)
```

samlaf: did you mean this? not sure I'm following where `getRouterAddress` comes from otherwise

store/builder/storage_manager_builder.go
```diff
@@ -206,19 +210,52 @@ func buildEigenDAV2Backend(
...
 	if err != nil {
-		return nil, fmt.Errorf("new cert verifier: %w", err)
+		log.Warn(`EigenDA cert verifier router address was detected to not be a router, using it as an
+			immutable cert verifier instead`)
```
samlaf, line 225: don't think this works. This call could fail for many reasons unrelated to this right? like a network error, or a context timeout, etc. Think we need to be a lot more strict and make sure that the error is `function doesn't exist` (not sure if that error exists though...)

samlaf: hmm https://pkg.go.dev/github.com/ethereum/go-ethereum/accounts/abi/bind#pkg-variables doesn't seem to have an error for this.. but there must be a way to figure this out?

store/generated_key/v2/eigenda.go
```diff
@@ -77,23 +80,40 @@ func NewStore(
...
+
+	case certs.V2VersionByte:
+		var v3Cert coretypes.EigenDACertV3
+		err := rlp.DecodeBytes(versionedCert.SerializedCert, &v3Cert)
+		if err != nil {
+			return nil, fmt.Errorf("RLP decoding EigenDA v3 cert: %w", err)
+		}
+		cert = &v3Cert
+
+	default:
```
samlaf, line 109: this is nice! Think it should be turned into a function `parseCert` or something, which returns the sum-type interface that `GetPayload` accepts. It should prob also be in core such that we can use it elsewhere (for example in some debugging library where people with proxy errors can just input their commitment and itll tell them which cert version it is from).

Prob can move/refactor this once we move proxy to core repo though? I had this ticket for adding such a function: https://linear.app/eigenlabs/issue/DAINT-465/certrbn-parsing-golang-library

store/generated_key/v2/eigenda.go
```diff
@@ -155,35 +175,62 @@ func (e Store) Put(ctx context.Context, value []byte) ([]byte, error) {
...
+	case coretypes.VersionTwoCert:
+		return nil, fmt.Errorf("EigenDA V2 certs are not supported anymore, use V3 instead")
+
+	case coretypes.VersionThreeCert:
+		eigenDACertV3, ok := cert.(*coretypes.EigenDACertV3)
+		if !ok {
+			return nil, fmt.Errorf("expected EigenDACertV3, got %T", cert)
+		}
+
+		return eigenDACertV3.Serialize(coretypes.CertSerializationRLP)
```
samlaf, line 188: Isnt the serialization type enforced by each cert version? So it feels bug-prone that we're doing this here. For example if someone refactors and changes this by mistake.
That's why originally I thought we should have 2 separate serialization functions: `SerializeABI()` and `SerializeBatcherInbox()` (or whatever).

Haven't thought this through in very deep details though.. so maybe we can discuss before making any changes

samlaf (commented):
Haven't finished everything but retro time so will submit my review.

store/generated_key/v2/eigenda.go
ethenotethan: sounds like a plan!

store/generated_key/v2/eigenda.go
ethenotethan: serialization should be determined by a custom encoding byte which we still need to integrate support for! in the interim it makes the most sense to default to RLP

store/builder/storage_manager_builder.go
ethenotethan: 🤔

ethenotethan: good callout - we could leverage what arbitrum uses for their BoLD activation check:
https://github.com/OffchainLabs/nitro/blob/master/staker/multi_protocol/multi_protocol_staker.go#L171-L188

ethenotethan: but even this makes assumptions on the provider - in the past we've seen issues where customers inject networking middleware between proxy and RPC.... 

https://github.com/OffchainLabs/nitro/blob/master/util/headerreader/header_reader.go#L29-L43

config/v2/eigendaflags/cli.go
litt3: IIRC the `Reader` struct doesn't expose this, and would need to be modified to do so.

server/handlers.go
samlaf: Oh I might have wrongly extrapolated here, or misremembered our previous conversations. I agree this is fine for this single migration.
But what about when we add a V4 cert for example. Is the idea to deal with that case only then? If so then feel free to resolve this discussion

store/builder/storage_manager_builder.go
samlaf: Maybe an easy solution for now is just to separate the CertVerifier and CertRouter flags and let the user input the correct choice?

store/generated_key/v2/eigenda.go
samlaf: The plan being for me to refactor this after we move proxy to monorepo? If so I'll update my ticket

common/types/certs/eigenda.go
```diff
@@ -41,3 +48,14 @@ func NewVersionedCert(serializedCert []byte, certVersion VersionByte) VersionedC
 func (c VersionedCert) Encode() []byte {
 	return append([]byte{byte(c.Version)}, c.SerializedCert...)
 }
+
+func (c VersionedCert) ToCoreCertType() (coretypes.CertificateVersion, error) {
+	switch c.Version {
+	case V0VersionByte, V1VersionByte:
+		return coretypes.VersionTwoCert, nil
```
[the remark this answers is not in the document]
ethenotethan: technically V0 should be touched in this code path - updated it to explicitly error though

config/v2/eigendaflags/cli.go
ethenotethan: [df6273a](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/df6273a049b984a44254fe2c7c412d3b3b19219c)

server/handlers.go
ethenotethan: > Oh wait does the payload disperser already set a version? From reading the certVerifier contract? 
Yes there is a version method returned in the EigenDA Cert interface. Added an issue for tracking:
https://github.com/Layr-Labs/eigenda-proxy/issues/404

common/store.go
```diff
@@ -95,7 +96,18 @@ type EigenDAStore interface {
...
+	Store
+	// Put inserts the given value into the key-value (serializedCert-payload) data store.
+	Put(ctx context.Context, payload []byte) (serializedCert []byte, err error)
+	// Get retrieves the given key if it's present in the key-value (serializedCert-payload) data store.
+	Get(ctx context.Context, versionedCert certs.VersionedCert) (payload []byte, err error)
+	// Verify verifies the given key-value pair.
+	Verify(ctx context.Context, versionedCert certs.VersionedCert, opts CertVerificationOpts) error
+}
+
+// PrecomputedKeyStore is the interface for a key-value data store that uses keccak(value) as the key.
```
litt3, line 110: Looks like your merge reverted a change made by Sam. Interface is now called `SecondaryStore`, but you've modified the doc back to the old name `PrecomputedKeyStore`

.env.exampleV1AndV2.sepolia
```diff
@@ -48,6 +48,11 @@ EIGENDA_PROXY_EIGENDA_STATUS_QUERY_TIMEOUT=30m0s
 # Number of times to retry blob dispersals for V1
 EIGENDA_PROXY_EIGENDA_PUT_RETRIES=3
 
+# The EigenDA network to run on. Specifying this chooses default values for V2_DISPERSER_RPC, V2_CERT_VERIFIER_ADDR,
```
litt3, line 51: V2_CERT_VERIFIER_ADDR isn't actually configured with this network field

litt3 (approved):

.env.exampleV2.sepolia
```diff
@@ -47,6 +47,8 @@ EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 # V2_SERVICE_MANAGER_ADDR, and V2_BLS_OPERATOR_STATE_RETRIEVER_ADDR.
 EIGENDA_PROXY_EIGENDA_V2_NETWORK=sepolia_testnet
```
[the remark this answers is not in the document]
ethenotethan: [fdd84a2](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/fdd84a22e1cbd38cd0b82dc8675702e31e44bbdc)

.env.exampleV1AndV2.sepolia
ethenotethan: [fdd84a2](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/fdd84a22e1cbd38cd0b82dc8675702e31e44bbdc)

store/builder/storage_manager_builder.go
```diff
@@ -206,19 +231,54 @@ func buildEigenDAV2Backend(
...
 	if err != nil {
 		return nil, fmt.Errorf("build eth reader: %w", err)
 	}
+	certVerifier, err := verification.NewCertVerifier(
+		log,
+		ethClient,
+		provider,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("new generic cert verifier: %w", err)
```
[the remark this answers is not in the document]
ethenotethan: [fdd84a2](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/fdd84a22e1cbd38cd0b82dc8675702e31e44bbdc)

common/store.go
ethenotethan: [fdd84a2](https://github.com/Layr-Labs/eigenda-proxy/pull/384/commits/fdd84a22e1cbd38cd0b82dc8675702e31e44bbdc)

litt3 (dismissed):

litt3 (approved):

PR #379, opened 2025-04-30
chore: refactor certs and commitments

samlaf:
This PR is a cleanup in preparation for the cert/rbn parsing library (see https://linear.app/eigenlabs/issue/DAINT-465/certrbn-parsing-golang-library)

We had commitmentMeta which consisted of commitmentMode + certVersion. This didn't really make sense to me as the certVersion should go with the cert.

So created a new struct VersionedCert which contains the certVersion and serializedCert. And then functions take this and a commitmentMode, as opposed to the serializedCert and commitmentMeta.

I believe this is a simplifying change. Its also forward looking because we will need to allow for new cert types. I plan to follow up with a cert/rbn parsing library, which will further simplify some of the logic in proxy.

## Fixes Issue

Fixes #

## Note to reviewers

I have more simplifying changes I want to make overall, but I believe the PR is big enough like this, so would appreciate a review of these initial changes. If they make sense, we can quickly merge and I'll start new PRs.

PR #386, opened 2025-05-02
chore: isolate keccak commitment POST route

samlaf:
Followup to discussion in https://github.com/Layr-Labs/eigenda-proxy/pull/379#discussion_r2070252750

keccak has a very different semantic from standard and op generic commitment modes... because it writes to s3 and doesn't need to deal with eigenda certs.

It didn't make sense for it to be behind the same store interface.

There is still some cleaning up to do (left some TODO comments in the code), but this is at least a start.

## Fixes Issue

Fixes #


store/precomputed_key/s3/errors.go
```diff
@@ -0,0 +1,19 @@
+package s3
+
+// TODO: this doesn't belong in the s3 package, but currently the Verify function returns
+// this error and is on S3. That also should be moved elsewhere.
+type Keccak256KeyValueMismatchError struct {
```
litt3, line 5: Could you describe what this error is? Why should we expect the key to match the value?

store/manager.go
```diff
@@ -277,16 +279,19 @@ func (m *Manager) getFromCorrectEigenDABackend(
...
 }
 
-// putKeccak256Mode ... put blob into S3 compatible backend
-func (m *Manager) putKeccak256Mode(ctx context.Context, key []byte, value []byte) ([]byte, error) {
+// PutOPKeccakPairInS3 puts a key/value pair, where key=keccak(value), into S3
+// This is only used for OP keccak256 commitments.
+func (m *Manager) PutOPKeccakPairInS3(ctx context.Context, key []byte, value []byte) error {
 	if m.s3 == nil {
-		return nil, errors.New("S3 is disabled but is only supported for posting known commitment keys")
+		return errors.New("S3 is disabled but is only supported for posting known commitment keys")
```
litt3, line 286: > but is only supported for posting known commitment keys

I'm not sure how to interpret this.

litt3 (dismissed):

samlaf:
Rebased on top of the 3 new commits in the target PR

server/handlers.go
```diff
@@ -167,38 +160,58 @@ func (svr *Server) handlePostStdCommitment(w http.ResponseWriter, r *http.Reques
...
+	payload, err := io.ReadAll(http.MaxBytesReader(w, r.Body, maxPOSTRequestBodySize))
+	if err != nil {
+		err = NewPOSTError(fmt.Errorf("failed to read request body: %w", err), commitments.OptimismKeccakCommitmentMode)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return err
+	}
+	err = svr.sm.PutOPKeccakPairInS3(r.Context(), keccakCommitment, payload)
+	if err != nil {
+		err = NewPOSTError(
+			fmt.Errorf("keccak POST request failed for commitment %v: %w", keccakCommitment, err),
```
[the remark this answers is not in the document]
samlaf: no, [46715ba](https://github.com/Layr-Labs/eigenda-proxy/pull/386/commits/46715ba2fc0a17bb209f198916569378f2c07538)

store/precomputed_key/s3/errors.go
samlaf: That was a very badly written error sorry... refactored and added docs in [c2424a1](https://github.com/Layr-Labs/eigenda-proxy/pull/386/commits/c2424a12632dfb195eb540c1cf979edb78e724fd)

store/manager.go
```diff
@@ -277,16 +279,20 @@ func (m *Manager) getFromCorrectEigenDABackend(
...
 	}
-
 	err := m.s3.Verify(ctx, key, value)
 	if err != nil {
-		return nil, err
+		return fmt.Errorf("PutOPKeccakPairInS3: %w", err)
```
litt3, line 291: the error message should probably indicate that it was the verification which failed

litt3 (dismissed):

store/manager.go
samlaf: I always get confused by this when wrapping error: should the wrapping document the caller or callee? I feel like both conventions work, we just need to settle on one. If its always the caller context then s3.Verify itself would have wrapped errors with the verify function context, so it would be fine.

samlaf: Also I just checked the only error possible is the mismatch key/value error, so I think its fine. Clearly its verification that is failing if you get that error.
Or is your general argument that every wrapping error in a given function should have a different message, to make it easy to find the exact call site when an error happens?

samlaf: I like that story, thanks for writing it. I like the "do a single thing well" approach, which in this case is "only answer what were you doing".

But on the other hand, I still feel like its just a simple matter of convention. I would have personally written the error in the second case as:
```
func divideInts(numerator int, denominator int) (int, error) {
    if denominator == 0 {
        return 0, errors.New("divideInts: denominator is 0")
    }

    return numerator / denominator, nil
}

func program() error {
    ...
    
    _, err := divideInts(a, b)
    if err != nil {
        return err
   }
}
```

Then you get the exact same error printed as in your first case.
If you think of it like: there are lib and bin writers (library and apps).
As a library author, you know exactly the name of the method where the error happened, so why not just include it? Then you save the trouble of every single app who integrates your library having to wrap and add that information. I think its nice actually of the library author does that. Then all the errors in program() can simply be returned without wrapping.

I guess this philosophy would be "add all the information that you have, when you have it, such that dependencies can use it for free"

litt3: > Then you save the trouble of every single app who integrates your library having to wrap and add that information.

My problem with this philosophy is trains people to not wrap errors.

Additionally, error messages are non-public API. So following this strategy means you have to dig into the non-public API of a library you're using in order to make your implementation decisions. On the other hand, the `what were you doing?` strategy lets you determine the correct log context without needing to consider the library's private implementation details.

samlaf: convinced: [17c3626](https://github.com/Layr-Labs/eigenda-proxy/pull/386/commits/17c362641795a429328927f5fd47247e4ee9f626)

PR #387, opened 2025-05-02
feat: Use reader as source for relay registry address

litt3:
- closes DAINT-471
    - [link](https://linear.app/eigenlabs/issue/DAINT-471/proxy-should-get-relay-registry-address-from-service-manager)

### DIFF
--- a/store/storage_manager_builder.go
+++ b/store/storage_manager_builder.go
@@ -246,17 +246,18 @@ func (smb *StorageManagerBuilder) buildEigenDAV2Backend(
 		return nil, fmt.Errorf("new cert verifier: %w", err)
 	}
 
+	ethReader, err := smb.buildEthReader(ethClient)
+	if err != nil {
+		return nil, fmt.Errorf("build eth reader: %w", err)
+	}
+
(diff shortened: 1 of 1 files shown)


store/storage_manager_builder.go
```diff
@@ -408,14 +409,10 @@ func (smb *StorageManagerBuilder) buildRelayClient(
 // payloads directly from EigenDA validators
 func (smb *StorageManagerBuilder) buildValidatorPayloadRetriever(
 	ethClient common_eigenda.EthClient,
+	ethReader *eth.Reader,
```
samlaf, line 412: does the ethReader expose the ethClient methods? If so can we only use the ethReader?

samlaf (approved):
LGTM. simple change

store/storage_manager_builder.go
ethenotethan: iiuc `ethReader` is really just `eigenDASysContractReader` - seems like there could be some benefit to reabstracting this thing

ethenotethan (approved):
LGTM

PR #389, opened 2025-05-05
ci: run workflows against all PRs (regardless of target branch)

samlaf:
Started working with stacked PRs, so all the PRs except for the first didn't have any CI running. Changing so we can catch bugs earlier.

## Fixes Issue

Fixes #


litt3 (approved):

.github/workflows/release-drafter.yml
```diff
@@ -4,9 +4,6 @@ name: Release Drafter
 
 on:
   push:
-    # branches to consider in the event; optional, defaults to all
-    branches:
-      - master
   # pull_request event is required only for autolabeler
```
ethenotethan, line 7: what's the rationale for enabling release-drafter for all branches?

samlaf: ah you're right. We have a main branch and not a master branch so I was very confused and figured this was probably not working properly in any case? Unless release-drafter is smart enough to convert from master to main?

Anyways you're right it should only draft on pushes to main, whatever its doing.

samlaf: [a776598](https://github.com/Layr-Labs/eigenda-proxy/pull/389/commits/a7765986ba65150c1bd93491f421218795030398)

PR #390, opened 2025-05-09
Support Enabling CORS on Proxy

jbrower95:
Adds flags to facilitate CORS:

- `--cors-allowed-domains <domain1>,<domain2>,<domain3>` which allows enabling CORS for a subset of domains. Can also specify "*" to allow all domains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,13 +316,7 @@ When CORS is enabled, the proxy will:
 2. Add CORS headers to all API responses
 3. Allow browsers to make cross-origin requests to the proxy API endpoints
 
-OPTIONS requests (known as "preflight requests") are a crucial part of the CORS security model. When a web application makes certain types of cross-origin requests (like those using certain HTTP methods or custom headers), browsers automatically send an OPTIONS request first to check if the server allows the actual request. This is a security mechanism built into browsers that:
-
-1. Helps protect servers from unexpected cross-origin requests
-2. Allows servers to specify which origins, methods, and headers are permitted
-3. Prevents potentially harmful operations from being executed without server approval
-
-For example, when a JavaScript client on website A.com tries to make a POST request with JSON data to our proxy on B.com, the browser first sends an OPTIONS request to verify that this cross-origin POST is allowed. Our proxy needs to respond appropriately to these OPTIONS requests for browser-based clients to work correctly.
+OPTIONS requests (known as "preflight requests") are a crucial part of the CORS security model. For more information on this part of the CORS model, see the following [resource](http://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#preflighted_requests).
 
 ### Requirements / Dependencies
 
--- a/README.md
+++ b/README.md
@@ -296,6 +296,28 @@ In the event that the EigenDA disperser or network is down, the proxy will retur
 
 This behavior is turned on by default, but configurable via the `--eigenda.confirmation-timeout` flag (set to 15 mins by default currently). If a blob is not confirmed within this time, the proxy will return a 503 status code. This should be set long enough to accomodate for the disperser's batching interval (typically 10 minutes), signature gathering, and onchain submission.
 
+#### CORS Support <!-- omit from toc -->
+The proxy can be configured to support Cross-Origin Resource Sharing (CORS), enabling web applications to make requests to the proxy server from different origins. This is useful for JavaScript clients or web applications that need to interact with the proxy directly from the browser.
+
+To enable CORS, use the following CLI flag:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -296,6 +296,27 @@ In the event that the EigenDA disperser or network is down, the proxy will retur
...
+./bin/eigenda-proxy --cors
+
+# Allow CORS only from specific domains
+./bin/eigenda-proxy --cors --cors-allowed-domains="myapp.com,localhost:3000"
+```
+
+When CORS is enabled, the proxy will:
+1. Respond to preflight OPTIONS requests with appropriate CORS headers
+2. Add CORS headers to all API responses
+3. Properly handle credentials and custom headers in cross-origin requests
```
samlaf, line 318: what credentials is this talking about? We don't use any authn in proxy afair

README.md
```diff
@@ -296,6 +296,27 @@ In the event that the EigenDA disperser or network is down, the proxy will retur
...
+- `--cors-allowed-domains`: A comma-separated list of domains to allow CORS from (e.g., `google.com,localhost:4157`). If this is empty and `--cors` is set, all origins will be allowed.
+
+Example usage:
+```bash
+# Allow CORS from any origin
+./bin/eigenda-proxy --cors
+
+# Allow CORS only from specific domains
+./bin/eigenda-proxy --cors --cors-allowed-domains="myapp.com,localhost:3000"
+```
```
samlaf, line 313: why not use a single flag `--cors--allowed-domains` which requires explicitly setting to * to allow all domains? Seems simpler UX and also safer (won't turn on CORS for all origins by mistake)?

server/middleware.go
```diff
@@ -102,3 +104,60 @@ func withLogging(
 		log.Info("request", args...)
 	}
 }
+
+// withCORS is a middleware that adds CORS headers to responses.
+// It intercepts OPTIONS requests for handling CORS preflight requests.
```
samlaf, line 109: why is this needed? Is it because our POST returns binary data?
image

server/middleware.go
```diff
@@ -102,3 +104,60 @@ func withLogging(
...
+			return
+		}
+
+		// Check if origin is allowed
+		allowOrigin := "*"
+		if len(corsConfig.CORSAllowedDomains) > 0 {
+			allowed := false
+			for _, domain := range corsConfig.CORSAllowedDomains {
+				if strings.TrimSpace(domain) == origin || 
+				   strings.HasSuffix(origin, strings.TrimSpace(domain)) {
```
samlaf, line 135: Seems to have a security flaw
image

Do we really need to match for subdomains? If yes, is there not a golang library that does proper matching of subdomains?
cc @anupsv

server/middleware.go
```diff
@@ -102,3 +104,60 @@ func withLogging(
...
+				log.Info("CORS rejected", "origin", origin)
+				handleFn(w, r)
+				return
+			}
+			allowOrigin = origin
+		}
+
+		// Set CORS headers
+		w.Header().Set("Access-Control-Allow-Origin", allowOrigin)
+		w.Header().Set("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
```
samlaf, line 150: asked this somewhere else but why is OPTIONS required? might be good to document here since most of us proxy developers are not super familiar with web stuff (at least talking for myself)

server/server.go
```diff
@@ -97,7 +97,26 @@ func BuildAndStartProxyServer(
...
+		// Copy all routes from the original router
+		corsRouter.PathPrefix("/").Handler(r)
+		
+		// Wrap the router with CORS middleware
+		handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			withCORS(corsRouter.ServeHTTP, svr.config, svr.log)(w, r)
+		})
+	}
+	
+	svr.httpServer.Handler = handler
```
samlaf, line 119: would prefer if we did this elsewhere than in the Start() function.
Can we abstract https://github.com/Layr-Labs/eigenda-proxy/blob/main/server/server.go#L84-L88 into a function `buildRouter()` or something and then add this code there?

samlaf (commented):
LGTM! Few changes I would prefer to make before merging. Not familiar with CORS myself so had to google some stuff, so feel free to let me know if I misunderstood something.

Also to make CI pass you'll need to:
1. fix lint issue (run `make lint` and fix)
2. run `make gen-static-help-output` and commit
3. we've never figured out how to allow external contributors to run our CI that requires secrets... if you know how would love to learn. Otherwise maybe best to make a PR from a branch directly inside our repo instead of from your own account

server/middleware.go
jbrower95: i think this is one of those things we should just handle now so that if you add more endpoints in the future, with other content types it's already adjusted.

PR #392, opened 2025-05-20
feat: Use most recent core commit

litt3:
- Updates to the most recent core commit, to be able to make a proxy release candidate
- I wasn't able to point to the exact `v0.9.0-rc.3` core tag, since that commit is missing the v1 bindings, which weren't added back in until core commit `dbc01a215a2811a8d4c30178b892dbd44694fc5b`, which is what I had to use


samlaf (approved):
LGTM. Not a big fan of the new binding struct names though.. pretty confusing :\ Might need to look into ways to make them more readable. Perhaps use type redefinitions to shield ourselves from 180 char contract names?

samlaf (approved):

PR #393, opened 2025-05-20
test: op altda client contract-test for rbn-recency-check

samlaf:
Main goal of this PR is to add [contract-tests](https://pactflow.io/blog/what-is-contract-testing/) for the recencyWindowCheck wrt op altda client, building on top of the new feature introduced in https://github.com/Layr-Labs/eigenda-proxy/pull/378

In order to add this I had to point our op dependency to our own fork. We were previously pointing at op v1.9.5 which is quite old, and a lot of things have changed in their op-e2e framework, which caused massive breaking changes, and it was literally impossible for me to make our previous op tests even run. I then looked and them and came to the conclusion that they don't belong in this repo, because they are testing op related stuff, and not much proxy stuff. For example we had an op migration test, which is a duplicate of the one we have in our op fork, the real place where it belongs.

Having removed our reliance on op-e2e framework, I was able to update our op dependency to our fork and test the new `l1_inclusion_block_number` query_param added to the client in https://github.com/Layr-Labs/optimism/pull/45.

Please see the commit full descriptions for more detail.

## Fixes Issue

Fixes #


store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +167,76 @@ func (e Store) BackendType() common.BackendType {
 //
 // Since v2 methods for fetching a payload are responsible for verifying the received bytes against the certificate,
 // this Verify method only needs to check the cert on chain. That is why the third parameter is ignored.
-func (e Store) Verify(ctx context.Context, certBytes []byte, _ []byte) error {
+//
+// TODO: this whole function should be upstreamed to a new eigenda VerifyingPayloadRetrieval client
```
litt3, line 171: Are we sure we want to put the recency check in that client? I can't remember if we came to a conclusion on that yet

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +167,76 @@ func (e Store) BackendType() common.BackendType {
...
+// that would verify certs, and then retrieve the payloads (from relay with fallback to eigenda validators if needed).
+// Then proxy could remain a very thing server wrapper around eigenda clients.
```
litt3, line 173: ~~thing~~ *thin*

store/generated_key/v2/eigenda.go
```diff
@@ -141,12 +167,76 @@ func (e Store) BackendType() common.BackendType {
...
+//
+// This check serves 2 purposes:
+//  1. liveness: prevents derivation pipeline from stalling on blobs that are no longer available on the DA layer
+//  2. safety: prevents a malicious EigenDA sequencer from using a very stale RBN whose operator distribution
+//     does not represent the actual stake distribution. Operators that withdrew a lot of stake would
+//     not be slashable anymore, even though because of the old RBN their signature would count for a lot of stake.
+//
+// Note that for a secure integration, this same check needs to be verified onchain.
+// There are 2 approaches to doing this:
+//  1. Pessimistic approach: use a smart batcher inbox to dissalow stale blobs from even beign included
```
litt3, line 210: ~~beign~~ *being*

litt3, line 210: ~~disallow~~ *disallow*

common/store.go
```diff
@@ -68,11 +68,24 @@ func StringToBackendType(s string) BackendType {
...
+	// This is optional, and should be set to 0 to mean to skip the RBN recency check.
+	// It is impossible for a batch inbox tx to have been included in the genesis block,
+	// so we are free to give this special meaning to the zero value.
+	//
+	// Used to determine the validity of the eigenDA batch.
+	// The eigenDA cert contains a reference block number (RBN) which is used
+	// to lookup the stake of the eigenda operators before verifying signature thresholds.
+	// The rollup commitment containing the eigenDA cert is only valid if it was included
+	// within a certain number of blocks after the RBN.
+	// validity condition is: certRBN < L1InclusionBlockNum < RBN + RBNRecencyWindowSize
```
litt3, line 82: I think you've misstated the validity case, it ought to read:

`certRBN < L1InclusionBlockNum <= certRBN + RBNRecencyWindowSize`

Personally, I try to pick either the "valid" or the "not valid" perspective, and make _every_ doc and piece of code approach from that same perspective. As it currently stands, some of your docs take the "valid" perspective, and others take the "invalid" perspective. It's really easy to introduce these off-by-one errors, trying to convert between the positive and negative case in your head.

store/generated_key/v2/eigenda.go
samlaf: What's your argument for not adding it to the client? After what bowen said last week that the "CertVerifier" should really be called a "DataAvailabilityVerifier" because its whole purpose is to parse a cert and make sure that if we try to download a blob, that blob should be available. And the recencyCheck is a big part of that (we could said it super laxly to a 14 day window for example for non-rollups). Its also not used by default (if the WindowSize is not set, it defaults to 0 which skips the check), so I don't personally see a reason why not to upstream it. But curious why you think otherwise

go.mod
```diff
@@ -6,41 +6,46 @@ toolchain go1.22.7
...
 	go.uber.org/mock v0.4.0
 	golang.org/x/exp v0.0.0-20241009180824-f66d83c29e7c
-	google.golang.org/grpc v1.64.1
+	google.golang.org/grpc v1.69.4
 )
 
 // TODO: Remove this after we have published v0.1.0 of the new clients module.
 replace github.com/Layr-Labs/eigenda-proxy/clients => ./clients/
 
+replace github.com/ethereum-optimism/optimism => github.com/Layr-Labs/optimism v0.0.0-20250522074124-9251a34831a7
```
samlaf, line 33: TODO: wait until https://github.com/Layr-Labs/optimism/pull/45 merges and then update this commit

store/generated_key/v2/eigenda.go
samlaf: [6892102](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/68921022877020328299a3bc7a6f58b4df822487)

store/generated_key/v2/eigenda.go
samlaf: [6892102](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/68921022877020328299a3bc7a6f58b4df822487)

common/store.go
samlaf: good catch: [f0e1e9a](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/f0e1e9a5f0a87d31ebd61ae4c73767598eeef197)
Agree with you.. I think the main issue is just that I have this (in)validity condition restated in way too many places. I tried to use the doc symlinks but it doesn't work as well as I would have liked. I'm not sure how to best do this, as I still think its good to have the same validity condition stated at least on the flag, config, and main verifying function logic.

I think everywhere left is using the validity perspective now? But let me know if not.

samlaf:
Merged master after merging https://github.com/Layr-Labs/eigenda-proxy/pull/378, and repointed to latest commit in https://github.com/Layr-Labs/optimism/pull/45, which we need to merge before merging this one.

store/generated_key/v2/eigenda.go
litt3: > But curious why you think otherwise

I didn't have a strong opinion one way or the other, just wanted to make sure this is the way to go.

Your arguments are persuasive, I'm on board with the plan of putting this into the client.

common/store.go
litt3: I agree, it's worth restating the validity criteria, even if it's duplicated

There are still some places in the proxy where you state the "stale" case, instead of the validity case:

- `.env` files
- `client_config_v2.go`
- `config/v2/eigendaflags/cli.go`

go.mod
samlaf: [ef23999](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/ef23999ba2f48f2abfaf85ef321ab29d091d6064)

common/store.go
samlaf: [48be85c](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/48be85cdd61398e71660fe8c24dbfab1ac90e38a)

litt3 (dismissed):

ethenotethan (dismissed):
LGTM

verify/verifier.go
```diff
@@ -23,6 +23,10 @@ const (
 )
 
 type Config struct {
+	// Cert verification is optional, and verifies certs retrieved from eigenDA when turned on.
+	// It is optional because it requires making calls to the blockchain, which is not necessarily always possible.
+	// For eg, some rollups are running on sepolia testnet which doesn't have an eigenlayer/eigenda contracts
+	// deployment.
```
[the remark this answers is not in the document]
samlaf: [0e815bb](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/0e815bb36f5563dac3aa13a78daa218c9cb79a58)

verify/verifier.go
```diff
@@ -43,6 +47,13 @@ func (c Config) MarshalJSON() ([]byte, error) {
 	return json.Marshal(aux)
 }
 
+// Verifier verifies the integrity of a blob and its certificate. There are 3 main categories of verification:
```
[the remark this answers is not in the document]
samlaf: [1327f4d](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/1327f4da47251b27be855b339c4694777c9d7b08)

config/v2/eigendaflags/cli.go
```diff
@@ -201,9 +201,9 @@ EigenDANetwork values include %s, %s, & %s.`,
 			Name: RBNRecencyWindowSizeFlagName,
 			Usage: `Allowed distance (in L1 blocks) between the eigenDA cert's reference 
 block number (RBN) and the L1 block number at which the cert was included 
-in the rollup's batch inbox. If certL1InclusionBlock > cert.RBN + rbnRecencyWindowSize, 
-the cert is considered stale and verification will fail. This check is 
-optional and will be skipped when set to 0.`,
+in the rollup's batch inbox. A cert is valid when certL1InclusionBlock <= cert.RBN + rbnRecencyWindowSize,
```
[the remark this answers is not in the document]
samlaf: [bafaad4](https://github.com/Layr-Labs/eigenda-proxy/pull/393/commits/bafaad40d62035e153e1abbd062562b6910caeab)

PR #395, opened 2025-05-22
feat: Sepolia Network Support

ethenotethan:
## Fixes Issue

Fixes https://github.com/Layr-Labs/eigenda-proxy/issues/394

## Changes proposed
  - Added Sepolia network constants and addresses for
   EigenDA contracts
  - Created environment configuration files for
  Sepolia deployments
  - Updated CI workflow for testing with Sepolia
  network
  - Minor code improvements for network handling

### DIFF
--- a/.env.exampleV1.sepolia
+++ b/.env.exampleV1.sepolia
@@ -1,0 +1,90 @@
+# An example sepolia testnet configuration
+
+# Hex-encoded signer private key. In V1, this key should not be associated with an Ethereum address holding any funds.
+EIGENDA_PROXY_EIGENDA_SIGNER_PRIVATE_KEY_HEX=
+
+# JSON RPC node endpoint for the Ethereum network used for finalizing DA blobs. See available list here: https://docs.eigenlayer.xyz/eigenda/networks/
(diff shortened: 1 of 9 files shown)


common/eigenda_network.go
```diff
@@ -61,11 +70,14 @@ func (n EigenDANetwork) String() string {
 	return string(n)
 }
 
+// TODO: Consider deriving network from chain ID via eth_getChainId or similar rpc call
+//
+//	vs hardcoding the network names and forcing user specification
```
samlaf, line 75: I think its prob best to let users specify chainID but then crash if specified chainID doesnt match the rpc_url. Prevent someone from dispersing to mainnet for a while when they really wanted a testnet proxy

.env.exampleV1.sepolia
samlaf, line 1: We talked about this during standup. This is a lot of copy-pasta that is impossible to know if all env files are up-to-date and have same content. Fine with merging this although I do think eventually we should clean this up. @litt3 said he might take a stab at it while doing memstore refactor

samlaf (approved):

litt3 (approved):

PR #406, opened 2025-06-03
feat: return HTTP teapot error for invalid certs

samlaf:
## Fixes Issue

Fixes DAINT-239.

## Changes proposed

Incorporates the structured error added in https://github.com/Layr-Labs/eigenda/pull/1617, and converts it to an HTTP 418 teapot error, the same way we already do for RBN check failures.

## Note to Reviewers

Making use of https://github.com/Layr-Labs/optimism/pull/49 in tests to make sure op client is parsing our TEAPOT errors correctly, so need that to merge first.

### DIFF
--- a/common/proxyerrors/4xx.go
+++ b/common/proxyerrors/4xx.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/Layr-Labs/eigenda-proxy/common"
+	_ "github.com/Layr-Labs/eigenda-proxy/store/generated_key/v2"
 	"github.com/Layr-Labs/eigenda-proxy/store/secondary/s3"
 	"github.com/Layr-Labs/eigenda/api/clients/v2/verification"
 	"google.golang.org/grpc/codes"
@@ -129,7 +130,8 @@ func (me UnmarshalJSONError) Error() string {
 
 // ParsingError is a very coarse-grained error that's used as a catch-all for any parsing errors
 // like parsing a hex string, or parsing a version byte from the request path, reading a query param, etc.
-// TODO: should all of these be returned as []
+// TODO: should all of these be returned as [eigenda.StatusCertParsingFailed] errors instead,
+// to return TEAPOTs instead of 400s?
 type ParsingError struct {
 	err error
 }


common/proxyerrors/4xx.go
```diff
@@ -36,7 +37,8 @@ func Is400(err error) bool {
 // used in optimism's derivation pipeline.
 func Is418(err error) bool {
 	var rbnRecencyCheckFailedErr eigendav2store.RBNRecencyCheckFailedError
-	return errors.As(err, &rbnRecencyCheckFailedErr)
+	var invalidCertErr verification.CertVerificationFailedError
+	return errors.As(err, &rbnRecencyCheckFailedErr) || errors.As(err, &invalidCertErr)
```
bxue-l2, line 41: Though theoretically correct that we can overload 418 to rbnRecencyCheckFailedErr and invalidCertErr, and get the desired function.

I don't think it is a good idea to overload, because it requires some baked assumption on the receiver end to process in order to find out where the error occurs. Currently, whether the cert is too old or is invalid. 

For instance in hokulea, the host queries proxy about eigenda blob and got this error, it now needs to do a local recency check and determine if it is too old, and inform hokulea client about it. The check itself is light, but this approach is more difficult to handle more error, and all receiver are forced to implement the error processing logics. 

Ideally, hokulea host only needs to pass through the error message from proxy to hokulea client, which is less error prone

samlaf: No you should never have to compare error messages. I didn't think you needed to know which error failed. If you do need that error switch, then how about we return a json with StatusCode as a field which you can switch on?

samlaf: now writing StatusCode + Msg as json fields in body: [3a6c23b](https://github.com/Layr-Labs/eigenda-proxy/pull/406/commits/3a6c23b6f43bbf8559b46b6ab1b4ab276f34748c)

server/middleware/error.go
```diff
@@ -24,7 +28,27 @@ func withErrorHandling(
...
+					StatusCode: -1,
+					Msg:        rbnRecencyCheckFailedErr.Error(),
+				}
+				_ = json.NewEncoder(w).Encode(statusCodeAndMsg)
+			}
+			// otherwise the error is a [verification.CertVerificationFailedError]
+			// so we json marshal it into the body to give access to the status code
+			// and message.
+			w.WriteHeader(http.StatusTeapot)
+			_ = json.NewEncoder(w).Encode(err)
```
bxue-l2, line 51: it will produce two json struct, if rbnRecencyCheckFailedErr case is triggered? Then it does not seem conventional.

samlaf: oh no good catch this needs to be removed... I was tired last night :\

samlaf: fixed in [f6a683e](https://github.com/Layr-Labs/eigenda-proxy/pull/406/commits/f6a683e4efd48c535cda6bd7365bb4eb149571d6)

server/middleware/error.go
```diff
@@ -24,7 +28,27 @@ func withErrorHandling(
...
+			if errors.As(err, &rbnRecencyCheckFailedErr) {
+				// We convert to a [verification.CertVerificationFailedError] like error,
+				// such that 418s always contain the same json body with StatusCode and Msg fields.
+				// All positive uint8 StatusCodes are reserved for errors coming from the CertVerifier contract,
+				// so we use negative errors to indicate other errors (of which we only have RBNRecencyCheck right now).
+				var statusCodeAndMsg = struct {
+					StatusCode int
+					Msg        string
+				}{
+					StatusCode: -1,
```
bxue-l2, line 42: It is better if we have a global error structure somewhere, to include both coretypes.VerificationStatusCode which is uint8, and those negative error. -1 seems a bit ad hoc

samlaf: I know I know... will leave a TODO. Will do this after we move proxy to monorepo. Too annoying to have to change the error there and then rebase yadi yada.

litt3 (dismissed):

server/middleware/error.go
```diff
@@ -20,11 +25,40 @@ func withErrorHandling(
...
 		case proxyerrors.Is400(err):
 			http.Error(w, err.Error(), http.StatusBadRequest)
-		case proxyerrors.Is418(err):
-			http.Error(w, err.Error(), http.StatusTeapot)
+		// 418 TEAPOT errors don't follow the pattern proxyerrors.Is418(err),
+		// because we have 2 very different errors that we need to marshal into the body of the 418 TEAPOT response.
+		case errors.As(err, &certVerificationFailedErr):
+			errBytes, _ := json.Marshal(certVerificationFailedErr)
+			fmt.Println("err", err)
+			fmt.Println(string(errBytes))
```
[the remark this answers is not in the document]
samlaf: [46c9324](https://github.com/Layr-Labs/eigenda-proxy/pull/406/commits/46c9324e4c601eea13b015c6f9fba08ed7ad9b55)

server/middleware/error.go
```diff
@@ -20,11 +25,40 @@ func withErrorHandling(
...
+			encodingErr := json.NewEncoder(w).Encode(certVerificationFailedErr)
+			if encodingErr != nil {
+				panic(encodingErr)
+			}
+		case errors.As(err, &rbnRecencyCheckFailedErr):
+			// We convert to a [verification.CertVerificationFailedError] like error,
+			// such that 418s always contain the same json body with StatusCode and Msg fields.
+			// All positive uint8 StatusCodes are reserved for errors coming from the CertVerifier contract,
+			// so we use negative errors to indicate other errors (of which we only have RBNRecencyCheck right now).
+			// TODO: we should probably fit the RBNRecencyCheckFailed errors into CertVerificationFailedErrors?
```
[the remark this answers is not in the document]
bxue-l2: i mean define an error spec, that maps from number(status code) to specific cause

We have two certVerificatio and recency. But they don't have standardization among them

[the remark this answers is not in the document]
samlaf: standardized in [b16caeb](https://github.com/Layr-Labs/eigenda-proxy/pull/406/commits/b16caebb08de2e1ddd76ac2cd0311fbb9844b9c5)

common/proxyerrors/4xx.go
```diff
@@ -129,6 +130,8 @@ func (me UnmarshalJSONError) Error() string {
 
 // ParsingError is a very coarse-grained error that's used as a catch-all for any parsing errors
 // like parsing a hex string, or parsing a version byte from the request path, reading a query param, etc.
+// TODO: should all of these be returned as [eigenda.StatusCertParsingFailed] errors instead,
+// to return TEAPOTs instead of 400s?
 type ParsingError struct {
```
samlaf, line 135: TODO: need to look at these error cases in more details to figure out what exactly needs to be returned...
Can potentially do as part of a followup PR given that this PR is a blocked for other work.

samlaf (commented):
@bxue-l2 refactored the error logic quite a bit. Still a TODO that I need to mull over... If the PR in its current state is fine with you then feel free to approve and merge and I'll fix the TODO in a follow-up PR tomorrow. Otherwise apologies but I'm super tired so complete fix will have to wait until tomorrow.

common/proxyerrors/4xx.go
bxue-l2: It can wait tomorrow, but the refactored looks great, thanks

bxue-l2 (approved):

PR #411, opened 2025-06-10
chore: Update eigenda v2 store to use latest legacy verification

ethenotethan:
## Fixes Issue

Fixes #

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -13,6 +13,8 @@ github.com/Layr-Labs/cerberus-api v0.0.2-0.20250117193600-e69c5e8b08fd h1:prMzW4
 github.com/Layr-Labs/cerberus-api v0.0.2-0.20250117193600-e69c5e8b08fd/go.mod h1:Lm4fhzy0S3P7GjerzuseGaBFVczsIKmEhIjcT52Hluo=
 github.com/Layr-Labs/eigenda v0.9.0-rc.3.0.20250602235625-e9611587885a h1:VlcjIk2Z1C3E4WQL8IFCg4j1eMiQ71tOGAldiN9Yjac=
 github.com/Layr-Labs/eigenda v0.9.0-rc.3.0.20250602235625-e9611587885a/go.mod h1:Zk1ScdOX1c2K77kI0vVz84085fgGtwTEgqyPxFjplhk=
+github.com/Layr-Labs/eigenda v0.9.0-rc.3.0.20250610184531-da8095c1c9b5 h1:xYBS7lBEQxW0tumZaQ0cysH2QWTnzsDl5Cm1nzdf3rs=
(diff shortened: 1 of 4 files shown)


store/generated_key/v2/eigenda.go
```diff
@@ -188,32 +188,52 @@ func (e Store) BackendType() common.BackendType {
 // Since v2 methods for fetching a payload are responsible for verifying the received bytes against the certificate,
 // this Verify method only needs to check the cert on chain. That is why the third parameter is ignored.
 func (e Store) Verify(ctx context.Context, versionedCert certs.VersionedCert, opts common.CertVerificationOpts) error {
-	switch versionedCert.Version {
-	case certs.V0VersionByte, certs.V1VersionByte:
-		e.log.Warn("EigenDA V2 certs are not supported anymore more for verification. Defaulting to successful verification.")
-		return nil
+	var retrievableDACert coretypes.RetrievableEigenDACert
```
litt3, line 191: I think it would be cleaner if you just did `var rbn uint64` here, instead of keeping track of the `retrievableDACert`. The `rbn` would be set in each case below, just like you're currently doing for the `retrievableDACert`

Then there is only one "cert" variable, so you could rename `sumDACert` to `eigenDACert`

go.sum
```diff
@@ -13,6 +13,8 @@ github.com/Layr-Labs/cerberus-api v0.0.2-0.20250117193600-e69c5e8b08fd h1:prMzW4
 github.com/Layr-Labs/cerberus-api v0.0.2-0.20250117193600-e69c5e8b08fd/go.mod h1:Lm4fhzy0S3P7GjerzuseGaBFVczsIKmEhIjcT52Hluo=
 github.com/Layr-Labs/eigenda v0.9.0-rc.3.0.20250602235625-e9611587885a h1:VlcjIk2Z1C3E4WQL8IFCg4j1eMiQ71tOGAldiN9Yjac=
```
litt3, line 14: nit: there are a ton of `eigenda` entries here- `go mod tidy` removes all the ones we don't actually need

litt3: Not sure whether we actually care about this or not

store/generated_key/v2/eigenda.go
ethenotethan: tbh the cleanest solution would be updating the core interfaces to converge on a single one. if additional cert fields become necessary for verification interfaces then doing the declare -> conditional populate flow would scale poorly. will make the change

go.sum
ethenotethan: [dd2f792](https://github.com/Layr-Labs/eigenda-proxy/pull/411/commits/dd2f792ac7c0230108a8be82b27510c45b855a14)

ethenotethan: we probably should given it'll bloat the dependency graph unnecessarily

ethenotethan: tangentially we could add a CI validation once merged into monorepo that runs `go mod tidy` and compares manifests to ensure no diff

PR #418, opened 2025-06-13
feat: configure memstore return instructed status code

bxue-l2:
## Fixes Issue

Fixes #DAINT-574

## Changes proposed 

The instructed mode allows users to set a desired returned status code for some payloads the user is about to write. The next time when a user requests to get the payloads with keys, the proxy returns the error according to the status code set earlier. The status code is sticky, and
can affect all subsequent writes. By default, the memstore isinitialized without the instructive mode.

```bash
 curl -X PATCH http://localhost:3100/memstore/config -d '{"InstructedMode": {"GetReturnsStatusCode": 3, "IsActivated": true }}'
 {"MaxBlobSizeBytes":2048,"BlobExpiration":"45m0s","PutLatency":"0s","GetLatency":"0s","PutReturnsFailoverError":false,"InstructedMode":{"GetReturnsStatusCode":3,"IsActivated":true}}
```

A user can only activate the instructed mode via http PATCH method above. A user can switch to other status code by sending a new `PATCH`
request, the retrieveal for subsequent writes would contain the new status code. There are currently seven available status codes
```
-1: Recency error. The proxy returns 418 with json string containing status code -1
1: Success. The proxy returns 200 with user payload, as if it is not using the instructed mode
0,2,3,4,5: Invalid Cert. The proxy returns 418 with json string containing the corresponding status code
```

A very important invariant which the instructed mode hold is that no key can ever be overwritten. This is important for all rollup use cases.

## Note to reviewers

Where it is used
- https://github.com/Layr-Labs/optimism/blob/869c01633cb80d0afd3727dadd0b8a2b2e8e687b/op-e2e/actions/proofs/helpers/eigenda_proxy.go#L23
- https://github.com/Layr-Labs/optimism/blob/869c01633cb80d0afd3727dadd0b8a2b2e8e687b/op-e2e/actions/proofs/eigenda_drop_cert_test.go#L113-L127

Built on top of #406, need to plan for merge

### DIFF
--- a/store/generated_key/memstore/README.md
+++ b/store/generated_key/memstore/README.md
@@ -48,6 +48,25 @@ $ curl http://localhost:3100/memstore/config | \
   curl -X PATCH http://localhost:3100/memstore/config -d @-
 ```
 
+### Instructed Mode
+
+The instructed mode allows users to set a desired returned status code for some payloads the user is about to write. The next time when a user requests to get the payloads with keys, the proxy returns the error according to the status code set earlier. The status code is sticky, and
+can affect all subsequent writes. By default, the memstore isinitialized without the instructive mode.
+
+```bash
+ curl -X PATCH http://localhost:3100/memstore/config -d '{"InstructedMode": {"GetReturnsStatusCode": 3, "IsActivated": true }}'
+ {"MaxBlobSizeBytes":2048,"BlobExpiration":"45m0s","PutLatency":"0s","GetLatency":"0s","PutReturnsFailoverError":false,"InstructedMode":{"GetReturnsStatusCode":3,"IsActivated":true}}
+```
+
+A user can only activate the instructed mode via http PATCH method above. A user can switch to other status code by sending a new `PATCH`
+request, the GET for subsequent writes contains the new status code. There are currently seven available status codes
+```
+-1: Recency error. The proxy returns 418 with json string containing status code -1
+1: Success. The proxy returns 200 with user payload, as if it is not using the instructed mode
+0,2,3,4,5: Invalid Cert. The proxy returns 418 with json string containing the corresponding status code
+```
+
+A very important invariant which the instructed mode hold is that no key can ever be overwritten. This is important for all rollup use cases.
 
 ### Golang client
 A simple HTTP client implementation lives in `/clients/memconfig_client/` and can be imported for manipulating the config using more structured types.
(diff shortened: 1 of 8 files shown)


store/generated_key/memstore/ephemeraldb/ephemeral_db.go
```diff
@@ -25,18 +28,20 @@ type DB struct {
...
 func New(ctx context.Context, cfg *memconfig.SafeConfig, log logging.Logger) *DB {
 	db := &DB{
-		config:    cfg,
-		keyStarts: make(map[string]time.Time),
-		store:     make(map[string][]byte),
-		log:       log,
+		config:                         cfg,
+		keyStarts:                      make(map[string]time.Time),
+		store:                          make(map[string][]byte),
+		instructedNonSuccessStatusCode: make(map[string]int),
```
bxue-l2, line 43: Not super happy about this treatment.

Alternative is to add a header byte or suffix byte to indicate the error and use the existing store.

bxue-l2: 9764a31d15d9ab42af769379b1f83dcb18deb259

store/generated_key/memstore/README.md
```diff
@@ -48,6 +48,25 @@ $ curl http://localhost:3100/memstore/config | \
   curl -X PATCH http://localhost:3100/memstore/config -d @-
 ```
 
+### Instructed Mode
```
samlaf, line 51: "instructed mode" is not a good choice. It's too generic and unclear what it refers to. Why not name it something related to CertVerification error or something more specific?

samlaf (commented):
Didn't actually review but just wanted to say that I don't think status 0 should be interpreted as invalid cert. I don't think it should be interpreted at all, because it should never happen. It should be treated as any status code that is not defined in the contract: a 500 error that must be resolved. Don't think the cert should be discarded from the derivation pipeline as it means the CertVerifier has a major flaw/bug, so that should be fixed.
image

bxue-l2:
> Didn't actually review but just wanted to say that I don't think status 0 should be interpreted as invalid cert. I don't think it should be interpreted at all, because it should never happen. It should be treated as any status code that is not defined in the contract: a 500 error that must be resolved. Don't think the cert should be discarded from the derivation pipeline as it means the CertVerifier has a major flaw/bug, so that should be fixed. 

I got your point that it is a bug. But for secure integration, nothing should ever stall. So derivation pipeline should ignore it. I am fine to call it something else

store/generated_key/memstore/README.md
bxue-l2: changed to InstructedStatusCodeReturn

bxue-l2:
replaced by #419

store/generated_key/memstore/README.md
bxue-l2: addressed in the new PR based on latest master. #419

PR #419, opened 2025-06-17
Feat  memstore return instructed status code

bxue-l2:
## Fixes Issue

Fixes #

### DIFF
--- a/clients/memconfig_client/client.go
+++ b/clients/memconfig_client/client.go
@@ -19,16 +19,24 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
+type InstructedStatusCodeReturn struct {
+	// return status code
+	GetReturnsStatusCode uint8
+	// if activated, GetReturnsStatusCode can be set to 1 to ensure normal operation
+	IsActivated bool
+}
+
 // MemConfig ... contains properties that are used to configure the MemStore's behavior.
 // this is copied directly from /store/generated_key/memstore/memconfig.
 // importing the struct isn't possible since it'd create cyclic dependency loop
 // with core proxy's go.mod
 type MemConfig struct {
-	MaxBlobSizeBytes        uint64
-	BlobExpiration          time.Duration
-	PutLatency              time.Duration
-	GetLatency              time.Duration
-	PutReturnsFailoverError bool
+	MaxBlobSizeBytes           uint64
+	BlobExpiration             time.Duration
+	PutLatency                 time.Duration
+	GetLatency                 time.Duration
+	PutReturnsFailoverError    bool
+	InstructedStatusCodeReturn InstructedStatusCodeReturn
 }
 
 // MarshalJSON implements custom JSON marshaling for Config.
@@ -36,22 +44,24 @@ type MemConfig struct {
 // which is hard to read.
 func (c MemConfig) MarshalJSON() ([]byte, error) {
 	return json.Marshal(intermediaryCfg{
-		MaxBlobSizeBytes:        c.MaxBlobSizeBytes,
-		BlobExpiration:          c.BlobExpiration.String(),
-		PutLatency:              c.PutLatency.String(),
-		GetLatency:              c.GetLatency.String(),
-		PutReturnsFailoverError: c.PutReturnsFailoverError,
+		MaxBlobSizeBytes:           c.MaxBlobSizeBytes,
+		BlobExpiration:             c.BlobExpiration.String(),
+		PutLatency:                 c.PutLatency.String(),
+		GetLatency:                 c.GetLatency.String(),
+		PutReturnsFailoverError:    c.PutReturnsFailoverError,
+		InstructedStatusCodeReturn: c.InstructedStatusCodeReturn,
 	})
 }
 
 // intermediaryCfg ... used for decoding into a less rich type before
 // translating to a structured MemConfig
 type intermediaryCfg struct {
-	MaxBlobSizeBytes        uint64
-	BlobExpiration          string
-	PutLatency              string
-	GetLatency              string
-	PutReturnsFailoverError bool
+	MaxBlobSizeBytes           uint64
+	BlobExpiration             string
+	PutLatency                 string
+	GetLatency                 string
+	PutReturnsFailoverError    bool
+	InstructedStatusCodeReturn InstructedStatusCodeReturn
 }
 
 // IntoMemConfig ... converts an intermediary config into a memconfig
@@ -72,12 +82,18 @@ func (cfg *intermediaryCfg) IntoMemConfig() (*MemConfig, error) {
 		return nil, fmt.Errorf("failed to parse blobExpiration: %w", err)
 	}
 
+	InstructedStatusCodeReturn := InstructedStatusCodeReturn{
+		IsActivated:          cfg.InstructedStatusCodeReturn.IsActivated,
+		GetReturnsStatusCode: cfg.InstructedStatusCodeReturn.GetReturnsStatusCode,
+	}
+
 	return &MemConfig{
-		MaxBlobSizeBytes:        cfg.MaxBlobSizeBytes,
-		BlobExpiration:          blobExpiration,
-		PutLatency:              putLatency,
-		GetLatency:              getLatency,
-		PutReturnsFailoverError: cfg.PutReturnsFailoverError,
+		MaxBlobSizeBytes:           cfg.MaxBlobSizeBytes,
+		BlobExpiration:             blobExpiration,
+		PutLatency:                 putLatency,
+		GetLatency:                 getLatency,
+		PutReturnsFailoverError:    cfg.PutReturnsFailoverError,
+		InstructedStatusCodeReturn: InstructedStatusCodeReturn,
 	}, nil
 }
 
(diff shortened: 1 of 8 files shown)


clients/memconfig_client/client.go
```diff
@@ -72,12 +82,18 @@ func (cfg *intermediaryCfg) IntoMemConfig() (*MemConfig, error) {
 		return nil, fmt.Errorf("failed to parse blobExpiration: %w", err)
 	}
 
+	GetReturnsInstructedStatusCode := GetReturnsInstructedStatusCode{
+		IsActivated:          cfg.GetReturnsInstructedStatusCode.IsActivated,
+		GetReturnsStatusCode: cfg.GetReturnsInstructedStatusCode.GetReturnsStatusCode,
+	}
```
samlaf, line 88: Don't think this is needed? Just use cfg.GetReturnsInstructedStatusCode directly?

store/generated_key/memstore/README.md
```diff
@@ -48,6 +49,18 @@ $ curl http://localhost:3100/memstore/config | \
   curl -X PATCH http://localhost:3100/memstore/config -d @-
 ```
 
+### Instructed Status Code Return
+The instructed status code return allows users to set a desired returned status code for some payloads the user is about to write. The next time when a user requests to get the payloads with keys, the proxy returns the result corresponding to the status code set earlier. The status code is sticky, and
```
samlaf, line 53: wdym by "user is about to write"? Isn't there no write? You just set this option and then reads return this error.

clients/memconfig_client/client.go
```diff
@@ -19,39 +19,49 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
+type GetReturnsInstructedStatusCode struct {
+	// return status code
+	GetReturnsStatusCode uint8
+	// if activated, GetReturnsStatusCode can be set to 1 to ensure normal operation
+	IsActivated bool
```
samlaf, line 26: Do we really need this? Why not just make this feature always activated, and make StatusCode=1 the default case (read normally from the memstore)?

store/generated_key/memstore/README.md
```diff
@@ -48,6 +49,18 @@ $ curl http://localhost:3100/memstore/config | \
...
+can affect all subsequent writes. By default, the memstore isinitialized without the instructed status code return.
 
+```bash
+ curl -X PATCH http://localhost:3100/memstore/config -d '{"GetReturnsInstructedStatusCode": {"GetReturnsStatusCode": 3, "IsActivated": true }}'
+ {"MaxBlobSizeBytes":2048,"BlobExpiration":"45m0s","PutLatency":"0s","GetLatency":"0s","PutReturnsFailoverError":false,"GetReturnsInstructedStatusCode":{"GetReturnsStatusCode":3,"IsActivated":true}}
+```
+
+A user can only activate the instructed status code via http PATCH method above. A user can switch to other status code from an existing activated status code
+by sending a new `PATCH` request, the GET for subsequent writes contains the new status code.
```
samlaf, line 62:
```suggestion
by sending a new `PATCH` request. Subsequent GET requests will return the new status code.
```
```suggestion
by sending a new `PATCH` request, the GET for subsequent writes contains the new status code.
```

store/generated_key/memstore/memconfig/config.go
```diff
@@ -118,6 +132,34 @@ func (sc *SafeConfig) SetMaxBlobSizeBytes(maxBlobSizeBytes uint64) {
 	sc.config.MaxBlobSizeBytes = maxBlobSizeBytes
 }
 
+func (sc *SafeConfig) GetGETReturnsInstructedStatusCode() (bool, coretypes.VerificationStatusCode) {
```
samlaf, line 135:
```suggestion
func (sc *SafeConfig) GETReturnsInstructedStatusCode() (bool, coretypes.VerificationStatusCode) {
```

clients/memconfig_client/client.go
```diff
@@ -19,39 +19,49 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
+type GetReturnsInstructedStatusCode struct {
```
samlaf, line 22:
```suggestion
type GetReturnsStatusCode struct {
```

store/generated_key/memstore/memconfig/config.go
```diff
@@ -118,6 +132,34 @@ func (sc *SafeConfig) SetMaxBlobSizeBytes(maxBlobSizeBytes uint64) {
...
+	sc.mu.RLock()
+	defer sc.mu.RUnlock()
+
+	return sc.config.GetReturnsInstructedStatusCode.IsActivated,
+		sc.config.GetReturnsInstructedStatusCode.GetReturnsStatusCode
+}
+
+func (sc *SafeConfig) SetGETReturnsInstructedStatusCode(mode GetReturnsInstructedStatusCode) error {
```
samlaf, line 143: please don't use "mode", that's already super overloaded in proxy. Just use `sc` here maybe if you want to be brief?

store/generated_key/memstore/memconfig/config.go
```diff
@@ -118,6 +132,34 @@ func (sc *SafeConfig) SetMaxBlobSizeBytes(maxBlobSizeBytes uint64) {
...
+		return fmt.Errorf("undefined status code, unable to configure memstore")
+	}
+
+	// reset to default value
+	if !mode.IsActivated {
+		mode.GetReturnsStatusCode = 0
+	}
+
+	// after it is activated, statusCode can be set to 1 to ensure normal operation
+	sc.config.GetReturnsInstructedStatusCode = mode
```
samlaf, line 159: wdym by set to 1? I think we should get rid of `isActivated` and just always treat status codes as activated. Wouldn't need all this bug-prone logic, also wouldn't need a struct in the configs (all the other are simple fields)

store/generated_key/memstore/memconfig/config.go
```diff
@@ -118,6 +132,34 @@ func (sc *SafeConfig) SetMaxBlobSizeBytes(maxBlobSizeBytes uint64) {
...
+}
+
+func (sc *SafeConfig) SetGETReturnsInstructedStatusCode(mode GetReturnsInstructedStatusCode) error {
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+
+	// If instructed to return a non Success Status(1), the memstore stores the error message
+	// on return.
+	if !eigenda.CheckValidStatusCode(mode.GetReturnsStatusCode) {
+		return fmt.Errorf("undefined status code, unable to configure memstore")
```
samlaf, line 150:
```suggestion
		return fmt.Errorf("undefined status code %v, unable to configure memstore", mode.GetReturnsStatusCode)
```

store/generated_key/memstore/memconfig/config.go
samlaf, line 143: document that the only error this function can return is a 4xx error. You use this assumption in your code

store/generated_key/memstore/ephemeraldb/ephemeral_db.go
```diff
@@ -68,18 +77,27 @@ func (db *DB) InsertEntry(key []byte, value []byte) error {
...
 	if exists {
 		return fmt.Errorf("payload key already exists in ephemeral db: %s", strKey)
 	}
 
-	db.store[strKey] = value
-	// add expiration if applicable
+	// force status code for all valid inputs
+	if !activated {
+		statusCode = 1
+	}
```
samlaf, line 91: this is sign of dangerous implicit dependencies/assumptions. I've said this in a few places but getting rid of activated will make this code simpler

samlaf (commented):
iiuc the semantics of having `GetReturnsInstructedStatusCode` turned on is that writes now store the StatusCode, and future GET requests for those payloads return the StatusCode. Feel like the name should be changed to `PUTStoreStatusCode` to reflect this semantics better. I was very confused for a while.

Otherwise we could keep the name `GetReturnsInstructedStatusCode` but flip the semantics and make it so that ALL GET requests started from this moment (regardless of when they were inserted) return the status code. This would have a much simpler semantics (similar to the other config settings). Would this work for the test cases you're trying to test though?

clients/memconfig_client/client.go
bxue-l2: There are too many placed that has to initialize, leaving statusCode =1 leaves little clue on what it means to the reader. By doing this, there is no contextual change at all

store/generated_key/memstore/ephemeraldb/ephemeral_db.go
bxue-l2: Activated is intended to hide statusCode from the rest of code. Given when the pr is written, we don't have status code clearly defined. 

I can remove this code too, given if it is not activated, statusCode is not looked at

store/generated_key/memstore/memconfig/config.go
bxue-l2: I really don't like the name, SetGETReturnsInstructedStatusCode. I would prefer the name in this commit, 404e73c3aa6b5a4197851ec33f74666f40c9487a

store/generated_key/memstore/memconfig/config.go
bxue-l2: instead only use isActivated

bxue-l2:
Currently this PR relies on statusCode =1 to determine if this is an instructed error, but I am super down to replace this usage

PR #421, opened 2025-06-25
chore: Update sepolia example env cert verifier

ethenotethan:
## Fixes Issue

Fixes #

## Changes proposed
Updates example env cert verifier address for v2 sepolia to use latest verifier vs legacy.


samlaf (commented):
How did we not catch this in tests/CI before your manual tests? Feel like we should add a test to catch these kinds of regressions in the future.

Oh is it because the tests are populated with addresses manually which are different from the ones here? This is another issue with keeping all these different configs everywhere, and why @litt3 was against having these example files.

Is there an easy way we can compare the values here to those hardcoded in the proxy binary and fail if they differ?

PR #422, opened 2025-06-26
Write on cache miss flag

gastonponti:
Flag to allow a proxy to save a blob in the cache after retrieving it from EigenDa's disperser/retriever.

There's already a flag that allows a proxy to fill the cache/fallback when is writing (Put) the blob.

This flag, allows a proxy that is only reading to fill its cache

Possible scenarios:
 * A l2-node operator following a rollup with multiple nodes but pointing to one proxy. The first call will retrieve it from the disperser and after that will be served by the cache
 * A l2-node operator following a rollup that wants to have possession of the blobs read for more than 14 days

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -211,7 +211,7 @@ issues:
         - unparam
 
     # Allow certain patterns to be ignored by lll (long lines)
-    - source: '".{120,}"' # Ignores double-quoted strings longer than 120 chars
+    - source: '".{100,}"' # Ignores double-quoted strings longer than 120 chars
       linters: [lll]
     - source: "// https?://" # This pattern matches comments containing URLs
       linters: [lll]


docs/help_out.txt
```diff
@@ -139,4 +139,5 @@ This check is optional and will be skipped when set to 0. (default: 0) [$EIGENDA
    --storage.concurrent-write-routines value                                  Number of threads spun-up for async secondary storage insertions. (<=0) denotes single threaded insertions where (>0) indicates decoupled writes. (default: 0) [$EIGENDA_PROXY_STORAGE_CONCURRENT_WRITE_THREADS]
    --storage.dispersal-backend value                                          Target EigenDA backend version for blob dispersal (e.g. V1 or V2). (default: "V1") [$EIGENDA_PROXY_STORAGE_DISPERSAL_BACKEND]
    --storage.fallback-targets value [ --storage.fallback-targets value ]      List of read fallback targets to rollover to if cert can't be read from EigenDA. [$EIGENDA_PROXY_STORAGE_FALLBACK_TARGETS]
+   --storage.write-on-cache-miss                                              While doing a GET, write to the cache if a blob is not found in the cache but is found in EigenDA. (default: false) [$EIGENDA_PROXY_STORAGE_WRITE_ON_CACHE_MISS]
```
samlaf, line 142: iiuc the semantics of our current flags, `--storage.cache-targets` will cause the PUT routes (only) to write certs to the cache after having dispersed right? (same for fallback but we prety much have internal consensus that fallback is redundant and we'll get rid of it eventually)

So turning on `--storage.write-on-cache-miss` will also populate the cache during retrieval (GET) routes.

Think this makes sense. Another possibility for the flag name is `--storage.use-cache-for-retrievals` or something.

cc @ethenotethan

samlaf:
```suggestion
   --storage.write-on-cache-miss                                              While doing a GET, write to the cache if a cert/blob pair is not found in the cache but is found in EigenDA. (default: false) [$EIGENDA_PROXY_STORAGE_WRITE_ON_CACHE_MISS]
```

samlaf: Also your msg says "in the cache", but what if there are multiple cache-targets turned on. Do you write to each? What about fallback targets?

samlaf (commented):
This LGTM but would like @ethenotethan to take a look before we merge.

For context @ethenotethan celo currently maintains a fork of proxy with this single PR in it, so I told them we'd happily upstream to not require them to have to maintain this if the implementation was clean. According to this PR it seems pretty clean to me, but I'll let you vet for yourself.

ethenotethan (commented):
this looks good, would it be possible to introduce a single E2E test that asserts behavior works as anticipated? example: https://github.com/Layr-Labs/eigenda-proxy/blob/0ccf931edfb35f41d6d4d4d9fd177f50b31701c3/test/e2e/server_test.go#L234-L274

docs/help_out.txt
gastonponti: It's actually using the same mechanism that the Put uses, so technically speaking will write to all the caches and fallbacks. I could change it to "secondary storages"
"write to the secondary storages if a cert/blob pair is not found in the cache but is found in EigenDA. (default: false)"
It could eventually rewrite a blob in a fallback source, but as it is the same one, it shouldn't be a problem

PR #424, opened 2025-06-30
fix: network id check bug (can't start on preprod)

samlaf:
## Fixes Issue

The behavior of proxy sould work fine for any external customers. But the way the chainid check is currently implemented makes it impossible to start proxy on our preprod environment. This PR fixes that.

> It had testnet hardcoded on chainid 17000, so wasn't possible to start proxy on preprod (which is also on 17000).
> Changed to alow multiple eigneda networks per chainid

Fixes #

### DIFF
--- a/common/client_config_v2.go
+++ b/common/client_config_v2.go
@@ -48,7 +48,7 @@ type ClientConfigV2 struct {
 	RBNRecencyWindowSize uint64
 
 	// The EigenDA network that is being used.
-	// It is optional, and when set will be used for validating that the eth-rpc chain ID matches network.
+	// It is optional, and when set will be used for validating that the eth-rpc chain ID matches the network.
 	EigenDANetwork EigenDANetwork
 }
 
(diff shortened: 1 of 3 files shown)


common/client_config_v2.go
```diff
@@ -47,7 +47,8 @@ type ClientConfigV2 struct {
 	// This check is optional and will be skipped when RBNRecencyWindowSize is set to 0.
 	RBNRecencyWindowSize uint64
 
-	// The EigenDA network that is being used. Used for validating chain ID matches network.
+	// The EigenDA network that is being used.
+	// It is optional, and when set will be used for validating that the eth-rpc chain ID matches network.
```
ethenotethan, line 51:
```suggestion
	// It is optional, and when set will be used for validating that the eth-rpc chain ID matches the network.
```

ethenotethan (approved):
lgtm

litt3 (approved):

